PR #185, opened 2023-05-08
[WIP] A parser for the titanic

Weyaaron:
This one has several open problems at the moment, but I wanted to get it out to enable cooperation and comments.
Any of these issues is up for grab, just overwrite the code and mark it as solved. 

- [ ] There is an issue with XPath, specifically with the summary. I have trouble extracting it. There is a comment in the code about it. 
- [ ] Should we include this parser at all? Since it is all about satire, it might mess with the intention to download news. 
- [ ] There are few articles, maybe #184 should be used on this one.


Weyaaron:
This PR is really low priority and may be redone once other more important tasks have been completed.

Issue #194, opened 2023-05-11
Add visual feedback to library


Weyaaron:
I had worries about the speed of the library. Turns out, the speed is fine, I was missing feedback. A few calls to tqdm should do the trick.

PR #211, opened 2023-05-13
[Proposal] Adds very basic feedback wrapper

Weyaaron:
This is an idea I just cooked up: I was missing some feedback from during the execution(#194). One solution might be a feature like this: A wrapper that provides optional feedback to the users. It could be easily expanded and maintained without clogging the main code. An obvious addition could be some sort of final summary after a run of the pipeline has finished. 
Edit: I forgot how to use it:

```python
de_de = PublisherCollection.at
crawler = Crawler(de_de)

with StatusDisplay(stepsize=50) as status:
     for article in status(crawler.crawl(max_articles=25, error_handling="suppress", only_complete=False)):
           print(article.source.url)
```
This will print a quick summary after each 500 Articles processed.

### DIFF
--- a/src/fundus/user_feedback.py
+++ b/src/fundus/user_feedback.py
@@ -1,32 +1,33 @@
-from typing import Iterator
+from typing import Iterator, Optional, Set
 
 from fundus.scraping.article import Article
 
 
 class StatusDisplay(Iterator[Article]):
-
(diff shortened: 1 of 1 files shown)


Weyaaron:
I received some feedback which indicated that this is at best low priority at the moment. Therefore, this pr is no longer relevant.

PR #212, opened 2023-05-15
[Proposal] Adds a new github action that measures doc string coverage

Weyaaron:
This new action uses https://pypi.org/project/docstr-coverage/ and https://github.com/KlexosNYC/docstr-coverage-action to measure the doc string coverage. I already excluded the publishers, since their code does not require docstrings in my opinion. I  suggest we do not accept prs that lower this percentage, the current failure percentage is 5%, our current status is something close to it.


MaxDall:
There is something really weird going on with this PR/branch. I will close this for now @Weyaaron.

PR #234, opened 2023-06-08
Id extraction from qse

Weyaaron:
This pr ports the id extraction from qse to fundus. The testing is not done yet, I would like to use the script once #231 has been done.


Weyaaron:
This feature will stay in QSE, this PR is no longer needed.

PR #249, opened 2023-06-26
Add LA times [Based on #251]

MaxDall:
This is based on #251  and should be merged afterward.

This adds support for the LA Times mentioned in the how_to_contribute.md


Weyaaron:
Seems fine to me, thanks for including it @MaxDall :)

Weyaaron (approved):

PR #276, opened 2023-07-07
GH258: Replaces single quotes with repr in f strings

### LINKED ISSUE #258: Use `repr` for strings within f-strings instead of single quotes.
Fundus utilizes the same construct 
```python
raise ValueError(f"Invalid value '{error_handling}' for parameter <error_handling>")
```
for displaying strings in f-strings.

As @dobbersc pointed out you could replace that pattern "...'{string}'..." with "...{string!r}..."

TODO: replace for entire code base.

Weyaaron:
This PR aims to update the fstrings present in this repository to use ```repr()``` instead of single quotes. I used a regex to find all of these instances.  

Closes #258


src/fundus/parser/utility.py
```diff
@@ -177,8 +177,8 @@ def generic_author_parsing(
         return []
 
     parameter_type_error: TypeError = TypeError(
-        f"<value> '{value}' has an unsupported type {type(value)}. "
-        f"Supported types are 'Optional[str], Dict[str, str], List[str], List[Dict[str, str]],'"
+        f"<value> {repr(value)} has an unsupported type {type(value)}."
```
MaxDall, line 180: That's not necessarily a string so  calling `__repr__` here leads to different results and thus was not intended to be changed with #258.

src/fundus/scraping/html.py
```diff
@@ -217,7 +216,7 @@ async def fetch(self) -> AsyncIterator[HTML]:
                     response.raise_for_status()
 
                 except (HTTPError, ClientError, HttpProcessingError, UnicodeError) as error:
-                    basic_logger.info(f"Skipped requested URL '{url}' because of '{error}'")
+                    basic_logger.info(f"Skipped requested URL {repr(url)} because of {repr(error)}")
```
MaxDall, line 219: `'{error}'` is not equal to `{repr(error)}` and should stay untouched.

MaxDall (changes requested):
Unfortunately it wasn't very obvious in the issue but i would prefer `!r` over `repr()`.

src/fundus/parser/utility.py
Weyaaron: Has been removed.

src/fundus/scraping/html.py
Weyaaron: Has been removed.

Weyaaron:
@MaxDall  I merged the latest master, the CI fails because I use a fork. All comments have been resolved.

src/fundus/parser/base_parser.py
```diff
@@ -70,9 +70,9 @@ def __lt__(self, other):
 
     def __repr__(self):
         if instance := self.__self__:
-            return f"bound {type(self).__name__} of {instance}: {self.__wrapped__} --> '{self.__name__}'"
+            return f"bound {type(self).__name__} of {instance}: {self.__wrapped__} --> !r{self.__name__}"
```
MaxDall, line 73: Unfortunately, it seems that you have misused the `!r` operator. It's supposed to be placed inside the curly braces like this:
```python
... {self.__name__!r}
```

MaxDall (changes requested):

src/fundus/parser/base_parser.py
Weyaaron: Has been done.

PR #278, opened 2023-07-10
GH262: Rework URL filter logic [Based on #277]

### LINKED ISSUE #256: Rename `HTMLSource`'s `url_filter` parameter to `url_filters` and type hint accordingly
Type-hinting `self.url_filter` would help since it differs from the past `url_filter`. Also, we could allow `Optional[list[URLFilter]]` in the first place. Then `url_filters` instead of `url_filter` would be more fitting. This is not really related to async, so maybe for another PR?

_Originally posted by @dobbersc in https://github.com/flairNLP/fundus/pull/239#discussion_r1248043886_

### LINKED ISSUE #262: Remove `add_url_filter` from `HTMLSource`
#261 moves the `HTMLSource` back to the pipeline so we can get of the horrible `add_url_filter` logic and directly init the `HTMLSource` with `URLFilter`'s using the `url_filters` parameter added with #256.

MaxDall:
Should be merged after #277 

Closes #262 
Closes #256

### DIFF
--- a/src/fundus/scraping/html.py
+++ b/src/fundus/scraping/html.py
@@ -29,7 +29,7 @@ from lxml.etree import XPath
 
 from fundus.logging import basic_logger
 from fundus.scraping.filter import URLFilter, inverse
-from fundus.utils.more_async import async_next, make_iterable_async
+from fundus.utils.more_async import make_iterable_async
 
 _default_header = {"user-agent": "Fundus"}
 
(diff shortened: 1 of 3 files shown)


src/fundus/scraping/html.py
```diff
@@ -240,33 +240,32 @@ def __init__(
...
-        self.url_filter = [] if not url_filter else [url_filter]
+        self.url_filter = url_filter
         self.request_header = request_header or _default_header
         if isinstance(url_source, URLSource):
             url_source.set_header(self.request_header)
 
-    def add_url_filter(self, url_filter: URLFilter) -> None:
-        self.url_filter.append(url_filter)
+    async def fetch(self, url_filter: Optional[URLFilter] = None) -> AsyncIterator[HTML]:
+        combined_filters = ([self.url_filter] if self.url_filter else []) + ([url_filter] if url_filter else [])
```
dobbersc, line 249:
```suggestion
        combined_filters: List[URLFilter] = ([self.url_filter] if self.url_filter else []) + ([url_filter] if url_filter else [])
```

src/fundus/scraping/pipeline.py
```diff
@@ -111,27 +111,20 @@ def constant_delay() -> float:
             else:
                 return delay
 
-        def build_unique_url_filter() -> URLFilter:
-            return lambda url: url in response_cache
+        def build_url_filter() -> URLFilter:
+            def _filter(url: str) -> bool:
+                return (url_filter and url_filter(url)) or (only_unique and url in response_cache)
```
dobbersc, line 116:
```suggestion
                return (url_filter is not None and url_filter(url)) or (only_unique and url in response_cache)
```

dobbersc (approved):

PR #281, opened 2023-07-11
Update APNews parser [Based on #280]

MaxDall:
This bumps up the AP Parser and version. Because only class variables are changed and the overall layout stays pretty much the same, I decided to try something else: Rather than using a completely new version I instead added a 'minor' version inheriting from the previous version.


Weyaaron (dismissed):
I really appreciate the ease of these changes.

Weyaaron (approved):

Issue #297, opened 2023-07-24
Update the documentation based on the comments from susanna


Weyaaron:
- [x] sometimes the next says we’ll crawl 5 articles but the code only crawls two

- [x]  In Tutorial 2 about the `Article` class, maybe state more clearly your definitions  add of `title`, `summary`, `section`, `subheadline`and so on. Makes it easier to follow and to try out the different parts. Also: What is DOM?

- [x]  Tutorial 3 for filtering the articles, e. g. by author: As much as I like Donald Duck, maybe one could chose a different, more realistic, example so users actually get good results instead of waiting, maybe wondering if the system is still running or what 😅 Maybe filtering for a specific publishing time range, language or topic?

- [x]  More info or examples on how to find the specific URLs to Sitemap, NewsMap, RSSFeed, as well as what their differences and purposes are, and how/why the need to differentiate them. At least for me, this was new and I found myself a bit puzzled at first. 

- [x]  Most importantly I would have liked more help for the basics of how to use the `CSSSelect` and `XPath` selectors, which is obviously the most work when adding a parser. I think that here, a small list of examples with html and some fitting example selectors, together with explanations what they do would be very helpful! 

_Originally posted by @susannaruecker in https://github.com/flairNLP/fundus/issues/296#issuecomment-1645568297_
            

Edit: I cut the original text down to bullet points.

Weyaaron:
I think points 1 and 3 are not controversial at all, I will start working on them. The other may require some more discussion.

Weyaaron:
All of these points have been addressed. And the contribution these are based upon is done(#305 ), therefore this can be closed.

PR #300, opened 2023-07-24
Removes the donald duck filter and replaces it with a better example

Weyaaron:
This is one of the bullet points in #297. This filter is a bit more complex, but should work most of the time in a timely manner.

### DIFF
--- a/docs/3_how_to_filter_articles.md
+++ b/docs/3_how_to_filter_articles.md
@@ -28,26 +28,28 @@ for article in crawler.crawl(max_articles=2, only_complete=Requires("title", "bo
 Writing custom extraction filters is supported and encouraged.
 To do so you need to define a `Callable` satisfying the `ExtractionFilter` protocol which you can find [here](../src/fundus/scraping/filter.py).
 
-Let's build a custom extraction filter that filters out all articles not written by an author called `Donald Duck`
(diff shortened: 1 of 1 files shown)


MaxDall:
Thanks for addressing this. The new example looks oddly complicated and noisy to me. I would prefer simplicity here to show a general concept. The focus here is to introduce extraction filtering in general.

I would prefer something more simply like this
```python
def topic_filter(extracted: Dict[str, Any]) -> bool:
    if topics := extracted.get("topics"):
        if "usa" in [topic.casefold() for topic in topics]:
            return False
    return True
```

or something more useful like

```python
def date_filter(extracted: Dict[str, Any]) -> bool:
    end_date = datetime.date.today() - datetime.timedelta(weeks=1)
    start_date = end_date - datetime.timedelta(weeks=1)
    if publishing_date := extracted.get("publishing_date"):
        return not (start_date <= publishing_date.date() <= end_date)
    return True

# or

def body_filter(extracted: Dict[str, Any]) -> bool:
    if body := extracted.get("body"):
        for word in ["pollution", "climate crisis"]:
            if word in str(body):
                return False
    return True
```
Also the filter proposed in this PR is missing a `not` before calling `any`

MaxDall (approved):

PR #302, opened 2023-08-08
Restructure contribution tutorials and add Table of Contents

MaxDall:


Weyaaron:
I noticed that a some changed files are numbered, some are not. I have no clear suggestion for an improvement, but I think it could be improved upon.

Weyaaron:
Maybe it could be split into developer documentation and user documentation? At the moment, the difference is not obvious.

Weyaaron:
Furthermore, the table of supported publishers and the attribute table are drowning in the other information. Maybe their visibility could be improved? I am open to accepting smaller changes and discussing this in our next meeting.

Weyaaron (approved):
It has been decided that we are fine with changing this structure later on.

PR #311, opened 2023-08-30
Landing page wording and style

alanakbik:
This PR suggest some minor tweaks to the landing README page: 
- centered tagline and starting links, inspired from Fabricator
- made the "Fundus is.." part less verbose 
- made the "Quick Start" part less verbose by removing the "requirements and installation" subheader
- slightly expanded on the explanation of example 1

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,9 @@ for article in crawler.crawl(max_articles=2):
     print(article)
 ```
 
-This should print something like this:
+That's already it!
+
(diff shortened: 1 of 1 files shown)


Weyaaron (approved):
I like these changes.

MaxDall:
Looks much cleaner now, thanks for the changes. I guess all that's left now would be a new logo 😅

PR #313, opened 2023-08-30
Add The Telegraph

MaxDall:


Weyaaron (approved):
I checked it out, looks solid to me :)

PR #317, opened 2023-09-01
Add Le Monde 

alanakbik:
Adds the first French news source: https://www.lemonde.fr/ 

Also corrects an example in the docs with the wrong return type.


src/fundus/publishers/fr/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+from fundus.publishers.fr.le_monde import LeMondeParser
+from fundus.scraping.filter import inverse, regex_filter
+from fundus.scraping.html import NewsMap, Sitemap
+
+
+class FR(PublisherEnum):
+    LeMonde = PublisherSpec(
+        name="Le Monde",
+        domain="https://www.lemonde.fr/",
+        sources=[NewsMap("https://www.lemonde.fr/sitemap_news.xml", reverse=True)],
```
MaxDall, line 11:
```suggestion
        sources=[NewsMap("https://www.lemonde.fr/sitemap_news.xml")],
```
It seems that this one is already in order from recent to older.

docs/how_to_add_a_publisher.md
```diff
@@ -499,7 +499,7 @@ class LATimesParser(ParserProxy):
             )
 
         @attribute
-        def publishing_date(self) -> datetime.datetime:
+        def publishing_date(self) -> Optional[datetime.datetime]:
```
MaxDall, line 502: Thanks for catching this!

MaxDall (changes requested):

MaxDall (approved):

PR #318, opened 2023-09-05
Replaces former logic in `crawl` with runner

MaxDall:
This:
- adds a new class `AsyncRunner` to `more_async.py` which handles an event loop
- replaces existing logic in `crawl()` with runner

### DIFF
--- a/src/fundus/scraping/pipeline.py
+++ b/src/fundus/scraping/pipeline.py
@@ -24,7 +24,7 @@ from fundus.scraping.article import Article
 from fundus.scraping.filter import ExtractionFilter, URLFilter
 from fundus.scraping.html import URLSource, session_handler
 from fundus.scraping.scraper import Scraper
-from fundus.utils.more_async import AsyncRunner, async_next
+from fundus.utils.more_async import ManagedEventLoop, async_next
 
 
 @runtime_checkable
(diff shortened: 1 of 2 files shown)


Weyaaron:
I am missing some documentation. With the context of #319 this makes obvious sense, but this context will be lost over time. In particular, I would like to see some comments if the class is intended for public use. I will approve nonetheless, since this documentation can be done later.

Weyaaron (approved):

Weyaaron (changes requested):
Actually, I changed my mind, this requires atleast documentation and I favour a slight tweak. In my opinion it is not obvious that this runner returns the event loop. And in my opinion it should not. Fixing this requires adding a new method which replaces the direct access to the loop with a method of the class Runner. Also, Runner does not inherit from abstractcontextmanager, which it probably should.

MaxDall:
@Weyaaron thanks for the input.

Wouldn't renaming `Runner` to something like `HandledEventLoop` fix all your concerns?

I called it `AsyncRunner` because it's descriptive of its purpose, and I still think it is. It is an async runner based on an event loop. The event loop itself is the runner. Adding more, like the method you mentioned, would just disguise the functionality.

> Also, Runner does not inherit from abstractcontextmanager, which it probably should.

Why?

Weyaaron:
We had a quick meeting and agreed that renaming it to "HandledEventLoop" loop does the job just fine. I will approve this once this has taken place.

PR #319, opened 2023-09-05
Add synced function to to get URLs from `URLSource`

MaxDall:
This adds functionality based on `AsyncRunner` to get URLs directly from `URLSource`'s


src/fundus/scraping/html.py
```diff
@@ -166,6 +167,19 @@ async def __aiter__(self) -> AsyncIterator[str]:
         async for url in self._get_pre_filtered_urls():
             yield url
 
+    def get_urls(self, max_urls: int = -1) -> Iterator[str]:
+        async_url_gen = self.__aiter__()
+        counter = 0
+        with AsyncRunner() as runner:
+            while True:
```
Weyaaron, line 174: I would prefer a version that uses range instead of this. The -1 default value confuses me. I think its readability is improved with the use of range.

MaxDall: `range` is of no good use to exhaust a generator, that's why no ranged is used here.

Weyaaron: We had a quick meeting and concluded that this code is better, but there should be a comment explaining the -1. I will approve once this has taken place.

MaxDall: It would be helpful if you could write the comment yourself to address what you didn't understand about the code in the first place.

Weyaaron: Good point, I will do so.

Weyaaron: Has been done, I approve this pr.

Weyaaron (approved):

PR #322, opened 2023-09-14
Change default value for `only_complete` parameter.

MaxDall:
Problem: Accepting only fully extracted articles as a default slows down Fundus and also presents it in a bad way. Many articles only lack some authors or topics and one could argue that the main focus is about the body.

Solution: This PR tries to solve this by accepting articles with at least a title, body, and publishing date set, and thus loosening the default.


Weyaaron (approved):
Looks solid to me.

PR #330, opened 2023-12-10
Makes article properties over writable by parser

### LINKED ISSUE #328: Allowing overwrite of lang attribute
In case certain publishers pass on the language of the article it should be possible to overwrite the auto-generated lang attribute of the article. This might make sense in cases where the language of the article cannot be accurately determined automatically.

MaxDall:
This PR adds functionality to write over properties like `lang` defined in the `article` class. We do so by replacing the former `property` logic with a custom one. The new implementation now caches the values as well so it now works like `cached_property` but instead of `cached_property` it can be overwritten.

I.e. to overwrite the `lang` attribute of `Article` use the following in your desired parser.

```python
from fundus.parser import ..., overwrite_attribute

    ...

    @overwrite_attribute
    def lang(self) -> str:
        return "This a new language value"
```

closes #328

### DIFF
--- a/src/fundus/parser/__init__.py
+++ b/src/fundus/parser/__init__.py
@@ -1,4 +1,10 @@
-from .base_parser import BaseParser, ParserProxy, attribute, function
+from .base_parser import (
+    BaseParser,
+    ParserProxy,
+    attribute,
+    function,
+    overwrite_attribute,
+)
 from .data import ArticleBody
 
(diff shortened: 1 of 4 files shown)


dobbersc (changes requested):

MaxDall:
No longer relevant after the latest rework #559

PR #332, opened 2024-01-11
Update `TheNation` parser

MaxDall:
This updates the selectors used for `TheNation` to suit the newest layout changes. I validated the date when the layout changes using CC-NEWS.


src/fundus/publishers/us/the_nation_parser.py
```diff
@@ -56,3 +66,13 @@ def title(self) -> Optional[str]:
         @attribute
         def topics(self) -> List[str]:
             return generic_topic_parsing(self.precomputed.meta.get("keywords"))
+
+    class V2(V1):
+        VALID_UNTIL = date.today()
+
+        _summary_selector = XPath("//article//div[contains(@class, 'article-title')]//p")
```
dobbersc, line 73: Just very minor: The V1 (and other parsers) have direct typehints for these variables for their first definition.

src/fundus/publishers/us/the_nation_parser.py
```diff
@@ -14,12 +22,14 @@
...
     class V1(BaseParser):
+        VALID_UNTIL = date(2023, 7, 22)
+
         # There is a known issue preventing lxml from extracting text content within
         # the specified summary node.
         # This is due to invalid XML provided by The Nation.
         # Currently(lxml 4.9.3), lxml does not accept p tags within any heading (h*) tag.
         # The "correct" selector would be ".article-header-content > h2 > p"
-        _summary_selector = CSSSelector(".article-header-content > h2")
+        _summary_selector: Selector = CSSSelector(".article-header-content > h2")
```
dobbersc, line 32: We could make this typehint more precise and annotate it directly with `CSSSelector`.

dobbersc (approved):
Thanks for fixing this. Just some very minor remarks for the typehints. I'll approve already.

src/fundus/publishers/us/the_nation_parser.py
MaxDall: Hmm, you sure? I only type hinted this variable in V1 because otherwise mypy would complain that we overwrite the type (CSSSelector with XPath). IMO type hinting these is a bit overkill unless necessary.

src/fundus/publishers/us/the_nation_parser.py
MaxDall: The problem here is that mypy will complain that we overwrite CSSSelector wit XPath.

src/fundus/publishers/us/the_nation_parser.py
dobbersc: Yes, you are right. In most parsers, we did not do this. It was properly only the parsers I've worked on to indicate where the variable is defined first. If it's overkill for you that's fine.

src/fundus/publishers/us/the_nation_parser.py
dobbersc: Yes, my bad.

dobbersc (dismissed):

src/fundus/publishers/us/the_nation_parser.py
MaxDall: I forgot that `CSSSelect` is derived from `XPath` so I removed `Selector` and had to touch the types again and then I accidentally added some to the new selectors as well ;) I guess my fingers slipped.

dobbersc (approved):

src/fundus/publishers/us/the_nation_parser.py
dobbersc: Haha, all good^^

PR #333, opened 2024-01-11
Fix `iNews` parser

MaxDall:
This loosens the paragraph selector used by `iNews`.

### DIFF
--- a/src/fundus/publishers/uk/i_news.py
+++ b/src/fundus/publishers/uk/i_news.py
@@ -15,7 +15,7 @@ from fundus.parser.utility import (
 class INewsParser(ParserProxy):
     class V1(BaseParser):
         _summary_selector = CSSSelector("article > h2")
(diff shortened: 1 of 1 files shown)


dobbersc:
What concrete example does require this change? Maybe it would be nice to introduce it as a test case.

MaxDall:
> What concrete example does require this change? Maybe it would be nice to introduce it as a test case.

I added an example to the description.

Good idea. Fundus currently doesn't test the article body so that would be a bigger thing to implement.

dobbersc:
> Update:
I opened a new PR https://github.com/flairNLP/fundus/pull/338 to deal with that.

Then, should this PR be merged after #338 including the test case?

MaxDall:
I would like to merge this before #338 which seems to take quite a while.

addie9800 (approved):

PR #336, opened 2024-01-12
Allow additional selectors for meta parsing

MaxDall:
This adds a new parameter `additional_selectors` to `get_meta_content`. The parameter expects a mapping in the form of a dictionary to add additional selectors to the meta parser.

Let's assume the following meta tag:
```html
<meta itemprop="headline" content="The Gospel According to Paul: A Comprehensive Study of Paul's Message in the 'New Testament'">
```

One can now pass the following mapping to the parser to extract this malformed `meta` tag.
```python
{
    "itemprop": CSSSelector("meta[itemprop]")
}
```

### DIFF
--- a/src/fundus/parser/utility.py
+++ b/src/fundus/parser/utility.py
@@ -121,7 +121,9 @@ def extract_article_body_with_selector(
 _meta_node_selector = CSSSelector("meta[name], meta[property]")
 
 
-def get_meta_content(tree: lxml.html.HtmlElement) -> Dict[str, str]:
+def get_meta_content(
+    tree: lxml.html.HtmlElement, additional_selectors: Optional[Dict[str, Union[XPath, CSSSelector]]] = None
+) -> Dict[str, str]:
     meta_nodes = _meta_node_selector(tree)
     meta: Dict[str, str] = {}
     for node in meta_nodes:
@@ -129,6 +131,13 @@ def get_meta_content(tree: lxml.html.HtmlElement) -> Dict[str, str]:
         value = node.attrib.get("content")
         if key and value:
             meta[key] = value
+
+    if additional_selectors:
+        for key, selector in additional_selectors.items():
+            for node in selector(tree):
+                if content := node.attrib.get("content"):
+                    meta[node.attrib[key]] = content
+
     return meta
 
 


src/fundus/parser/utility.py
```diff
@@ -121,14 +121,23 @@ def extract_nodes(selector: XPath, node_type: Type[Node]) -> List[Node]:
...
         value = node.attrib.get("content")
         if key and value:
             meta[key] = value
+
+    if additional_selectors:
+        for key, selector in additional_selectors.items():
+            for node in selector(tree):
+                if content := node.attrib.get("content"):
+                    meta[node.attrib[key]] = content
+
```
addie9800, line 140: As it seems to me the user needs to pass the same value for the key as the meta attribute they are filtering by. e.g. calling `get_meta_content(doc, dict({"test": lxml.cssselect.CSSSelector("meta[itemprop]")}))` raises a KeyError. I was wondering whether it is really necessary to pass a dict, which causes an exception if used incorrectly, if the value of the key is already predetermined.

src/fundus/parser/utility.py
```diff
@@ -121,14 +121,23 @@ def extract_nodes(selector: XPath, node_type: Type[Node]) -> List[Node]:
...
     for node in meta_nodes:
         key = node.attrib.get("name") or node.attrib.get("property")
         value = node.attrib.get("content")
         if key and value:
             meta[key] = value
+
+    if additional_selectors:
+        for key, selector in additional_selectors.items():
+            for node in selector(tree):
+                if content := node.attrib.get("content"):
```
addie9800, line 138: There is an edge case, if there are multiple meta tags using the attribute within the selector. For example using: a document with html: `'<meta itemprop="headline" content="Erster Inhalt"><meta itemprop="headline" content="Zweiter Inhalt"><meta name="robots" content="max-image-preview:large"/>'` and calling `get_meta_content(lxml.html.document_fromstring(html), dict({"itemprop": lxml.cssselect.CSSSelector("meta[itemprop]")}))` will get you the result: `{'headline': 'Zweiter Inhalt', 'robots': 'max-image-preview:large'}`. I'm not sure how meta tags are restricted, but in case there's the possibility of them having the same attribute value, this might lead to unwanted behavior.

addie9800 (changes requested):
Thanks for adding this feature 👍

src/fundus/parser/utility.py
MaxDall: Yeah, you're right. Using a dict with the current implementation doesn't make much sense. I'm gonna think of something else here after reading a bit through the documentation.

src/fundus/parser/utility.py
MaxDall: Within the meta `name`/`property`/`itemprop` building three different categories. `name` coming from HTML5, `property` from RDFa, and god knows where `itemprop` is coming from. Within those categories having duplicate names shouldn't be allowed and overriding seems to be an appropriate strategy here.

PR #337, opened 2024-01-12
Update `WorldTruth` parser

MaxDall:
This is based on #336, merge afterward.

Bumps up the parser version of `WorldTruth`


addie9800:
It seems to me as if WorldTruth is no longer available. I tried accessing it last week and today again and it seems to me that their server is down (possibly for good)

MaxDall:
Gonna close this for now

PR #338, opened 2024-01-12
Add `body` to unit tests

MaxDall:
The adds the `body` to parser unit tests as well as updating the test cases for that occasion.


dobbersc:
The CI fails because some publishers are missing the test case for the body attribute. 

When I ran the test case generation script, I noticed some publishers may have incorrectly extracted attributes. Did you roughly check the extracted bodies?

dobbersc:
Weirdly the order of the parser versions in the JSON for SZ and TheNation are also in a different order compared to the current commit. I think this may be related to the path globbing, which may not guarantee a deterministic list order.

MaxDall:
> Weirdly the order of the parser versions in the JSON for SZ and TheNation are also in a different order compared to the current commit. I think this may be related to the path globbing, which may not guarantee a deterministic list order.

I don't quite get this one. If I look into the JSONs the parser versions are in the same order as in the master. Could you provide me with a bit more context? I think I'm missing your point here

dobbersc:
> > Weirdly the order of the parser versions in the JSON for SZ and TheNation are also in a different order compared to the current commit. I think this may be related to the path globbing, which may not guarantee a deterministic list order.
> 
> I don't quite get this one. If I look into the JSONs the parser versions are in the same order as in the master branch. Could you provide me with a bit more context? I think I'm missing your point here.

It is essentially the same as with the order of attributes. I just have not pushed the changes as agreed. You can see the diff for TheNation.json here: https://www.diffchecker.com/6S7112p6/

src/fundus/parser/data.py
```diff
@@ -231,8 +272,18 @@ class ArticleSection(TextSequenceTree):
     headline: TextSequence
```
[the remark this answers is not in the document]
MaxDall: Headline was originally intended to be a TextSequence to catch any errors resulting from consequent headlines.

tests/utility.py
```diff
@@ -88,25 +90,32 @@ class ExtractionEncoder(json.JSONEncoder):
     def default(self, obj: object):
         if isinstance(obj, datetime.datetime):
             return str(obj)
+        if isinstance(obj, TextSequenceTree):
```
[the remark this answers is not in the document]
MaxDall: Sounds good to me!

MaxDall:
@dobbersc Is there anything left open here? I would like to merge this PR.

scripts/generate_parser_test_files.py
```diff
@@ -123,11 +123,15 @@ def main() -> None:
             for html in html_mapping.values():
                 versioned_parser = html.publisher.parser(html.crawl_date)
                 extraction = versioned_parser.parse(html.content)
-                new = {attr: value for attr, value in extraction.items() if attr in arguments.attributes}
+                missing_attributes = set(arguments.attributes) - set(
+                    test_data.get(type(versioned_parser).__name__) or {}
```
addie9800, line 127: If I'm not mistaken, this line is only needed to update old test cases. Shouldn't it be removed as soon as all old test data is updated?

addie9800 (commented):
This is a nice addition, there is one merge conflict left though

scripts/generate_parser_test_files.py
MaxDall: Depends if we want to add additional attributes in the future ;)

addie9800: True :)

Issue #344, opened 2024-02-01
[Bug]: Fundus not installing on Google Colab


alanakbik:
### Describe the bug

I tried running the tutorial code in a fresh [colab environment](https://colab.research.google.com ), but when running

```
pip install fundus
```

the installation fails with the output

```
google-colab 1.0.0 requires requests==2.31.0, but you have requests 2.28.2 which is incompatible.
yfinance 0.2.36 requires requests>=2.31, but you have requests 2.28.2 which is incompatible. 
```

### How to reproduce

```python
pip install fundus

from fundus import PublisherCollection, Crawler

# initialize the crawler for news publishers based in the US
crawler = Crawler(PublisherCollection.us)

# crawl 2 articles and print
for article in crawler.crawl(max_articles=2):
    print(article)
```

### Expected behavior.

It installs correctly and I can run all tutorials on Google Colab.

### Environment

```markdown
Python 3.10.12, CPU-only, Google Colab
```

MaxDall:
@alanakbik Thanks for reporting this.

Sadly I couldn't reproduce the exact error. I get the same message but Fundus seems to be installed anyway.

Just to be clear: Could you run this code snippet after a fresh `pip install fundus`

```python
from fundus import Crawler, PublisherCollection

async def crawl():
  crawler = Crawler(*PublisherCollection.de)
  async for article in crawler.crawl_async(max_articles=10, only_complete=False):
    print(article)

await crawl()
```

While investigating this I found several other problems regarding notebooks.
1. Most of Fundus won't run in the async environment of jupyter notebooks. (You can get the main crawler running by using the crawler's async api but not CCNewsCrawler`) -> I gonna hopefully get CCNews to run in a notebook with a PR today
2. crawl_asyncs default parameter for `only_complete` is bugged and won't crawl
3. The independent is blocking the complete us pipeline -> I will fix this in a PR today

alanakbik:
Yes, can confirm that your snippet runs for me!

MaxDall:
@alanakbik We changed the version restrictions for our dependencies so installing Fundus within google colab should no longer yield an error. The new release may take some while but the changes are already live on master so using `pip install git+https://github.com/flairNLP/fundus` should do the trick for now.

alanakbik:
Thanks, it now installs without problem! 

But both these snippets still don't work for me: 

```python
from fundus import CCNewsCrawler, PublisherCollection

crawler = CCNewsCrawler(*PublisherCollection.de)
[...]

PR #347, opened 2024-02-01
Replace `asyncio` with thread-based solution for WARC-path download

MaxDall:
Since I couldn't manage to run the `asyncio` solution within an already running event loop (see #344, especially inside jupyter notebooks) I reverted this part to the former thread-based solution.


dobbersc (approved):
Back to threading :).

PR #352, opened 2024-02-05
Update Bild Sources, Update SZ Parser

addie9800:
I found a SiteMap and NewsMap for Bild, so I decided to add that.

I have also fixed the paragraph and subheadline selectors for SZ. I have not created a new parser class for this, since SZ has seemed to have updated their UI entirely and the selectors work (at least as well as the old ones) on all available articles (e.g. https://www.sueddeutsche.de/politik/bundesrat-union-sperrt-sich-gegen-neue-handwerksordnung-1.309844 from 2003)


src/fundus/publishers/de/sz.py
```diff
@@ -14,9 +15,14 @@
...
-        _paragraph_selector = CSSSelector('main [itemprop="articleBody"] > p, ' "main .css-korpch > div > ul > li")
+        _paragraph_selector = XPath(
+            "//div[@itemprop='articleBody']//p[@data-manual='paragraph'" " and not(contains(text(), '© dpa-infocom'))]"
+        )
         _summary_selector = CSSSelector("main [data-manual='teaserText']")
-        _subheadline_selector = CSSSelector("main [itemprop='articleBody'] > h3")
+        _subheadline_selector = XPath(
+            "//div[@itemprop='articleBody']//h3[@data-manual='subheadline']|"
+            "//div[@itemprop='articleBody']//h2[@data-manual='subheadline']"
+        )
```
MaxDall, line 25: Introducing a new layout is necessary here. If a publisher changes their site layout, it always effects all articles regardless of their publishing time since they are not static but rendered dynamically. The reason we do the extra proxy step is to be able to parse articles which were crawled before the layout changed. The proxy determines the matching parser based on the crawl date and not the publishing date. This is in order to be able to support large HTML archives like CC-NEWS. 

PS: I would stick to a minor version `V1_1` inheriting from `V1` since only the selectors changed.

MaxDall, line 25: Normally `h2` is reserved for titles so that seems a bit odd to me. Could you explain the reasoning why using both?

src/fundus/publishers/de/__init__.py
```diff
@@ -180,7 +180,11 @@ class DE(PublisherEnum):
     Bild = PublisherSpec(
         name="Bild",
         domain="https://www.bild.de/",
-        sources=[RSSFeed("https://www.bild.de/rssfeeds/vw-neu/vw-neu-32001674,view=rss2.bild.xml")],
+        sources=[
+            RSSFeed("https://www.bild.de/rssfeeds/vw-neu/vw-neu-32001674,view=rss2.bild.xml"),
+            NewsMap("https://www.bild.de/sitemap-news.xml"),
+            Sitemap("https://www.bild.de/sitemap-index.xml"),
+        ],
```
MaxDall, line 187: Very nice you found those!

src/fundus/publishers/de/sz.py
```diff
@@ -42,3 +48,9 @@ def title(self) -> Optional[str]:
         @attribute
         def topics(self) -> List[str]:
             return generic_topic_parsing(self.precomputed.ld.bf_search("keywords"))
+
+        @attribute(validate=False)
+        def free_access(self) -> bool:
+            if self.precomputed.ld.bf_search("isAccessibleForFree") is None:
+                return True
+            return not self.precomputed.ld.bf_search("isAccessibleForFree") == "False"
```
MaxDall, line 56: Since there is an ongoing debate about weather we add this as a core attribute or not i would rather do this all at once. Maybe an issue about this topic would help.

MaxDall (changes requested):
Thanks for updating both Bild and SZ 👍. For the future it might me good to separate PRs like this with independent changes so that the debate on one addition does not block the other. In this case `Bild` would be an easy merge.

src/fundus/publishers/de/sz.py
addie9800: I added it for backwards compatibility. But that was based on the thought that older articles might also be parsed by accessing the url. When I looked though older articles I found at least a case or two where the subheadings used h2 and I thought it couldn't hurt to add that to the selectors as well

MaxDall (approved):

PR #353, opened 2024-02-12
Add Heise

addie9800:


addie9800:
Missing ?seite=all at end of all URLs, otherwise multipage articles will not be crawled

PR #355, opened 2024-02-14
Add utility to retrieve test articles

MaxDall:
This adds a utility function to easily get the parsed test articles for a specific parser.


addie9800 (approved):
Thanks for adding this

PR #357, opened 2024-02-15
Unbatch Fundus

### LINKED ISSUE #344: [Bug]: Fundus not installing on Google Colab

MaxDall:
This PR removes the remaining async logic from the main crawler and in the process merges CCNewsCrawler and MainCrawler logic. While this is a major redesign from the inside there is not much changing on the outside, but for the few things it does change, I adjusted the documentation accordingly.

The main advantage of this redesign is that Fundus main crawler no longer operates in batches but runs every publisher independently using threads and queues. Now, bad connections, like timeouts and denied access no longer halt the pipeline because the crawler has to wait for a new batch.

Closes #344


src/fundus/scraping/crawler.py
```diff
@@ -178,123 +430,24 @@ def filter_warc_path_by_date(path: str) -> bool:
...
         else:
-            article_iter = self._parallel_crawl(warc_paths, article_task)
-
-        for article_idx, article in enumerate(article_iter, start=1):
-            if not only_unique or article.html.responded_url not in response_cache:
-                response_cache.add(article.html.responded_url)
-                yield article
-            if article_idx == max_articles:
-                break
+            yield from self._parallel_crawl(warc_paths, article_task)
```
addie9800, line 453: This crashes when running on gruenau using the following code. The problem does not occur on a single thread.
```
from fundus import PublisherCollection, CCNewsCrawler
from datetime import datetime

publisher = PublisherCollection.de
crawler = CCNewsCrawler(publisher, start=datetime(2024, 3, 1))
for article in crawler.crawl(max_articles=30, only_complete=False):
    print(article)
```

With the following Tracestack:
```
multiprocessing.pool.RemoteTraceback:
"""
Traceback (most recent call last):
  File "/usr/local/anaconda3-2023.03/lib/python3.10/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/usr/local/anaconda3-2023.03/lib/python3.10/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
  File "/vol/fob-vol3/nebenf20/breidina/Arbeit/fundus/src/fundus/scraping/crawler.py", line 69, in __call__
    return self._deserialize()(*args, **kwargs)
  File "/vol/fob-vol3/nebenf20/breidina/Arbeit/fundus/src/fundus/scraping/crawler.py", line 66, in _deserialize
    return cast(Callable[_P, _T], dill.loads(self._serialized_target))
  File "/usr/local/anaconda3-2023.03/lib/python3.10/site-packages/dill/_dill.py", line 286, in loads
    return load(file, ignore, **kwds)
  File "/usr/local/anaconda3-2023.03/lib/python3.10/site-packages/dill/_dill.py", line 272, in load
    return Unpickler(file, ignore=ignore, **kwds).load()
  File "/usr/local/anaconda3-2023.03/lib/python3.10/site-packages/dill/_dill.py", line 419, in load
    obj = StockUnpickler.load(self)
  File "/usr/local/anaconda3-2023.03/lib/python3.10/site-packages/dill/_dill.py", line 568, in _create_type
    return typeobj(*args)
  File "/usr/local/anaconda3-2023.03/lib/python3.10/typing.py", line 348, in __init_subclass__
    raise TypeError("Cannot subclass special typing classes")
TypeError: Cannot subclass special typing classes
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/vol/fob-vol3/nebenf20/breidina/Arbeit/fundus/test.py", line 7, in <module>
    for article in crawler.crawl(max_articles=30, only_complete=False):
  File "/vol/fob-vol3/nebenf20/breidina/Arbeit/fundus/src/fundus/scraping/crawler.py", line 203, in crawl
    for article in self._build_article_iterator(
  File "/vol/fob-vol3/nebenf20/breidina/Arbeit/fundus/src/fundus/scraping/crawler.py", line 453, in _build_article_iterator
    yield from self._parallel_crawl(warc_paths, article_task)
  File "/vol/fob-vol3/nebenf20/breidina/Arbeit/fundus/src/fundus/scraping/crawler.py", line 380, in _parallel_crawl
    yield from pool_queue_iter(pool.map_async(serialized_article_task, warc_paths), article_queue)
  File "/vol/fob-vol3/nebenf20/breidina/Arbeit/fundus/src/fundus/scraping/crawler.py", line 110, in pool_queue_iter
    handle.get(timeout=0.1)
  File "/usr/local/anaconda3-2023.03/lib/python3.10/multiprocessing/pool.py", line 774, in get
    raise self._value
TypeError: Cannot subclass special typing classes
```

src/fundus/scraping/scraper.py
```diff
@@ -64,22 +42,57 @@ async def scrape(
...
+        html_sources = [
+            WebSource(
+                url_source=url_source,
+                publisher=publisher.publisher_name,
+                request_header=publisher.request_header,
+                delay=delay,
+            )
+            for url_source in url_sources
+        ]
+        parser_mapping = {publisher.publisher_name: publisher.parser}
```
addie9800, line 89:
```suggestion
        parser_mapping: Dict[str, ParserProxy] = {publisher.publisher_name: publisher.parser}
```

src/fundus/scraping/session.py
```diff
@@ -0,0 +1,92 @@
...
+
+import requests.adapters
+
+from fundus.logging import basic_logger
+
+_default_header = {"user-agent": "Fundus"}
+
+
+class SessionHandler:
+    """Object for handling  project global aiohttp.ClientSessions
```
addie9800, line 11: The docstring is outdated

src/fundus/scraping/session.py
```diff
@@ -0,0 +1,92 @@
...
+    def _session_factory(self) -> requests.Session:
+        """Builds a new Session
+
+        This returns a new client session build from pre-defined configurations:
+        - pool_connections: 50
+        - pool_maxsize: 50
+        - hooks = {'request': lambda request:}
+
+        Returns:
+            An new ClientSession
```
addie9800, line 36:
```suggestion
            A new requests.Session
```

src/fundus/scraping/session.py
```diff
@@ -0,0 +1,92 @@
...
+        self.pool_connections = pool_connections
+        self.pool_maxsize = pool_maxsize
+
+    def _session_factory(self) -> requests.Session:
+        """Builds a new Session
+
+        This returns a new client session build from pre-defined configurations:
+        - pool_connections: 50
+        - pool_maxsize: 50
+        - hooks = {'request': lambda request:}
```
addie9800, line 33:
```suggestion
```

src/fundus/scraping/url.py
```diff
@@ -0,0 +1,130 @@
...
+                sitemap_locs = [node.text_content() for node in self._sitemap_selector(tree)]
+                filtered_locs = list(filter(inverse(self.sitemap_filter), sitemap_locs))
+                for loc in reversed(filtered_locs) if self.reverse else filtered_locs:
+                    yield from yield_recursive(loc)
+
+        yield from yield_recursive(self.url)
+
+
+@dataclass
+class NewsMap(Sitemap):
```
addie9800, line 129: Is there a need to have this class? To me it seems that it is treated as any other `SiteMap`

addie9800 (changes requested):
Thanks for updating this, this really improves the performance a lot.

scripts/generate_parser_test_files.py
```diff
@@ -5,25 +5,25 @@
...
         crawler = Crawler(enum)
+        return next(crawler.crawl(max_articles=1, error_handling="suppress", only_complete=True), None)
     else:
-        source = FundusSource([url], publisher=enum.publisher_name)
-        scraper = Scraper(source, parser=enum.parser)
-        crawler = BaseCrawler(scraper)
-    return next(crawler.crawl(max_articles=1, error_handling="suppress", only_complete=True), None)
+        source = WebSource([url], publisher=enum.publisher_name)
+        scraper = BaseScraper(source, parser_mapping={enum.publisher_name: enum.parser})
+        return next(scraper.scrape(error_handling="suppress", extraction_filter=RequiresAll()))
```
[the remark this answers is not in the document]
MaxDall: I guess giving more feedback makes sense here. Im gonna log/print some message here.

src/fundus/scraping/url.py
MaxDall: Yes, this is necessary in order to be able to differentiate between `Sitemap` and `NewsMap`

src/fundus/scraping/crawler.py
```diff
@@ -106,10 +113,213 @@ def pool_queue_iter(handle: MapResult[Any], queue: Queue[_T]) -> Iterator[_T]:
...
+                    fitting_publisher.append(publisher)
+
+            if not fitting_publisher:
+                basic_logger.error(
+                    f"Could not find any fitting publisher for required attributes  "
+                    f"`{', '.join(extraction_filter.required_attributes)}`"
+                )
+                return
+
+        article_idx = 0
```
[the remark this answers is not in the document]
MaxDall: You're right! Something else: Suggesting code changes here within GitHub seems somewhat dangerous because it's tempting to just commit then but they won't propagate to the rest of the code.

src/fundus/scraping/crawler.py
```diff
@@ -106,10 +113,213 @@ def pool_queue_iter(handle: MapResult[Any], queue: Queue[_T]) -> Iterator[_T]:
...
+        publisher: PublisherEnum,
+        error_handling: Literal["suppress", "catch", "raise"],
+        extraction_filter: Optional[ExtractionFilter] = None,
+        url_filter: Optional[URLFilter] = None,
+    ) -> Iterator[Article]:
+        def build_delay() -> Optional[Delay]:
+            if isinstance(self.delay, float):
+
+                def constant_delay() -> float:
+                    return self.delay  # type: ignore[return-value]
```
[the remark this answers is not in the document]
MaxDall: Nice!

docs/5_how_to_search_for_publishers.md
```diff
@@ -15,7 +15,8 @@ You can search through the collection to get only publishers fitting your use ca
 Let's get some publishers based in the US, supporting an attribute called `topics` and `NewsMap` as a source, and use them to initialize a crawler afterward.
 
 ````python
-from fundus import Crawler, PublisherCollection, NewsMap
+from fundus import Crawler, PublisherCollection
```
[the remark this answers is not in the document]
MaxDall: No, I guess that happened while using PyCharm's refactoring tool.

docs/4_how_to_filter_articles.md
```diff
@@ -184,7 +184,8 @@ You can preselect the source for your articles when initializing a new `Crawler`
 Let's initiate a crawler who only crawls from `NewsMaps`'s.
 
 ````python
-from fundus import Crawler, PublisherCollection, NewsMap
+from fundus import Crawler, PublisherCollection
```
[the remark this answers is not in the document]
MaxDall: No, I guess that happened while using PyCharm's refactoring tool.

src/fundus/scraping/session.py
MaxDall: I adjusted the doc string accordingly.

src/fundus/scraping/session.py
MaxDall: I added the fitting hooks to the doc string.

src/fundus/scraping/crawler.py
```diff
@@ -106,10 +113,213 @@ def pool_queue_iter(handle: MapResult[Any], queue: Queue[_T]) -> Iterator[_T]:
...
+        Args:
+            *publishers (Union[PublisherEnum, Type[PublisherEnum]]): The publishers to crawl.
+            restrict_sources_to (Optional[List[Type[URLSource]]]): Lets you restrict
+                sources defined in the publisher specs. If set, only articles from given source types
+                will be yielded.
+            delay (Optional[Union[float, Delay]]): Set a delay time in seconds to be used between article
+                downloads. You can set a delay directly using float or any callable satisfying the Delay
+                protocol. If set to None, no delay will be used between batches. See Delay for more
+                information. Defaults to None.
+        """
```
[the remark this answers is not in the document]
MaxDall: I added the missing parameter.

MaxDall:
@addie9800 Thanks for the review. This one is ready for another round. Maybe @dobbersc also want to give it a look.

src/fundus/scraping/crawler.py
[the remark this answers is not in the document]
addie9800: Good point, I'll keep that in mind

src/fundus/scraping/crawler.py
addie9800: Ah sorry, to try both variants I forced the value of self.processes to be 0 and then it ran and worked fine using the portion of code:
```
article_iter = self._single_crawl(warc_paths, article_task)
 yield from self._single_crawl(warc_paths, article_task)
```
But without forcing it to run with self.processes = 0, it crashed with the above mentioned error message

addie9800:
> @addie9800 Thanks for the review. This one is ready for another round. Maybe @dobbersc also want to give it a look.

I'm happy for now, except for the two still open comments I just replied to.

src/fundus/scraping/crawler.py
MaxDall: As @addie9800 found out this is caused when using `typing-extensions` versions `4.4` and `4.5`. To prevent this I'm gonna set `typing-extension` >= `4.6`.

src/fundus/scraping/html.py
```diff
@@ -274,99 +37,188 @@ class HTML:
...
+class WebSourceInfo(SourceInfo):
     type: str
     url: str
 
 
-class FundusSource:
+class HTMLSource(Protocol):
+    @abstractmethod
+    def fetch(self, url_filter: Optional[URLFilter] = None) -> Iterator[HTML]:
+        ...
```
dobbersc, line 64: This looks a bit weird to me: Is this class indented to be a protocol or an abstract class with an abstract method?

src/fundus/scraping/html.py
```diff
@@ -274,99 +37,188 @@ class HTML:
...
+                    else SourceInfo(self.publisher)
+                )
+
+                yield HTML(
+                    requested_url=url,
+                    responded_url=str(response.url),
+                    content=html,
+                    crawl_date=datetime.now(),
+                    source=source,
+                )
```
dobbersc, line 139: Since each source now has a corresponding SourceInfo class, I think it would make sense to change the associated variables and parameters source to `source_info` where a SourceInfo is used. Otherwise these two may get mixed up.

src/fundus/scraping/html.py
```diff
@@ -274,99 +37,188 @@ class HTML:
...
+                        )
+                else:
+                    basic_logger.warning(
+                        f"Couldn't detect charset for record {record.record_id!r} from {target_url!r} "
+                        f"with invalid original charset {record.http_charset!r}."
+                    )
+
+            return None
+
+        with requests.Session() as session:
```
dobbersc, line 186: Would it be appropriate to use the handled global session here, or is this a different use case?

scripts/generate_parser_test_files.py
```diff
@@ -5,25 +5,25 @@
...
         crawler = Crawler(enum)
+        return next(crawler.crawl(max_articles=1, error_handling="suppress", only_complete=True), None)
     else:
-        source = FundusSource([url], publisher=enum.publisher_name)
-        scraper = Scraper(source, parser=enum.parser)
-        crawler = BaseCrawler(scraper)
-    return next(crawler.crawl(max_articles=1, error_handling="suppress", only_complete=True), None)
+        source = WebSource([url], publisher=enum.publisher_name)
+        scraper = BaseScraper(source, parser_mapping={enum.publisher_name: enum.parser})
+        return next(scraper.scrape(error_handling="suppress", extraction_filter=RequiresAll()), None)
```
dobbersc, line 26: The `else` is not necessary here, and can be removed to save indentation.

dobbersc (changes requested):
Thanks for the great rework! The interface definitely looks cleaner now.

src/fundus/scraping/session.py
```diff
@@ -0,0 +1,93 @@
...
+        self.pool_connections = pool_connections
+        self.pool_maxsize = pool_maxsize
+
+    def _session_factory(self) -> requests.Session:
+        """Builds a new Session
+
+        This returns a new client session build from pre-defined configurations:
+        - pool_connections: 50
+        - pool_maxsize: 50
+        - hooks = {'response': raise_for_status(), _response_log():}
```
dobbersc, line 33: The docstring for the hooks look malformed, but I'm not really sure what should go here.

src/fundus/scraping/crawler.py
```diff
@@ -106,10 +114,220 @@ def pool_queue_iter(handle: MapResult[Any], queue: Queue[_T]) -> Iterator[_T]:
...
+
+    @staticmethod
+    def _threaded_crawl(
+        publishers: Tuple[PublisherEnum, ...], article_task: Callable[[PublisherEnum], Iterator[Article]]
+    ) -> Iterator[Article]:
+        article_queue: Queue[Article] = Queue(len(publishers))
+        wrapped_article_task = queue_wrapper(article_queue, article_task)
+
+        with ThreadPool(processes=len(publishers) or None) as pool:
+            yield from pool_queue_iter(pool.map_async(wrapped_article_task, publishers), article_queue)
```
dobbersc, line 303: Just a minor question: We do not need the dill wrapper for the standard Crawler since it is thread based. Is that correct?

src/fundus/scraping/html.py
dobbersc: Maybe this would be better for a separate PR.

src/fundus/scraping/crawler.py
dobbersc: I did a fresh install from this PR's branch and had no crashes when following the tutorials.

src/fundus/scraping/crawler.py
```diff
@@ -106,10 +113,213 @@ def pool_queue_iter(handle: MapResult[Any], queue: Queue[_T]) -> Iterator[_T]:
...
+    def _single_crawl(
+        publishers: Tuple[PublisherEnum, ...], article_task: Callable[[PublisherEnum], Iterator[Article]]
+    ) -> Iterator[Article]:
+        article_iterators = [article_task(publisher) for publisher in publishers]
+        while article_iterators:
+            for iterator in article_iterators:
+                try:
+                    yield next(iterator)
+                except StopIteration:
+                    article_iterators.remove(iterator)
```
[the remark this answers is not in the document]
MaxDall: > Isn't the same issue still there, since tmp is just a direct reference to article_iterators?

Yes, that's true. I overlooked that i would have needed any form of `copy`, which makes this somehow inefficient.

> I think using [interleave_longest](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.interleave_longest) or [roundrobin](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.roundrobin) from more itertools would help here.

Yes. I still had in my mind that both share a significant flaw that requires one element from each iterator to be retrieved before yielding the first one, making it unusable for our case, but that's only the case for `interleave_longest`. I'm gonna replace the existing code with `roundrobin.`

README.md
```diff
@@ -100,7 +100,7 @@ Maybe you want to crawl a specific news source instead. Let's crawl news article
 from fundus import PublisherCollection, Crawler
 
 # initialize the crawler for Washington Times
-crawler = Crawler(PublisherCollection.us.WashingtonTimes)
+crawler = Crawler(PublisherCollection.us.TheNewYorker)
```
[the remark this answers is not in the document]
MaxDall: Thanks, gonna change it to ``New Yorker`

src/fundus/scraping/session.py
```diff
@@ -0,0 +1,93 @@
...
+
+        # adapters
+        session.mount(
+            "http://",
+            requests.adapters.HTTPAdapter(pool_connections=self.pool_connections, pool_maxsize=self.pool_maxsize),
+        )
+        session.mount(
+            "https://",
+            requests.adapters.HTTPAdapter(pool_connections=self.pool_connections, pool_maxsize=self.pool_maxsize),
+        )
```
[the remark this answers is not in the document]
MaxDall: I made some changes so that `pool_connections` == `len(publishers)` and `pool_maxsize` == 0.

src/fundus/scraping/crawler.py
```diff
@@ -30,17 +32,23 @@
...
+from fundus.scraping.html import CCNewsSource
+from fundus.scraping.scraper import CCNewsScraper, WebScraper
+from fundus.scraping.session import session_handler
+from fundus.scraping.url import URLSource
 
 _T = TypeVar("_T")
 _P = ParamSpec("_P")
 
+Publisher: TypeAlias = Union[PublisherEnum, Type[PublisherEnum], PublisherCollectionMeta]
+
```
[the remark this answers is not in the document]
MaxDall: Yes, all of those are different types and also valid. We use `more_itertools.collapse` so that's why it doesn't make a difference for the crawler in the end.

src/fundus/scraping/crawler.py
```diff
@@ -106,10 +114,220 @@ def pool_queue_iter(handle: MapResult[Any], queue: Queue[_T]) -> Iterator[_T]:
...
+        def build_extraction_filter() -> Optional[ExtractionFilter]:
+            if isinstance(only_complete, bool):
+                return None if only_complete is False else RequiresAll()
+            else:
+                return only_complete
+
+        response_cache: Set[str] = set()
+
+        extraction_filter = build_extraction_filter()
+        fitting_publisher: List[PublisherEnum] = []
```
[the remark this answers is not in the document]
MaxDall: Agree

src/fundus/scraping/crawler.py
```diff
@@ -106,10 +114,220 @@ def pool_queue_iter(handle: MapResult[Any], queue: Queue[_T]) -> Iterator[_T]:
...
+                basic_logger.error(
+                    f"Could not find any fitting publisher for required attributes  "
+                    f"`{', '.join(extraction_filter.required_attributes)}`"
+                )
+                return
+
+        article_count = 0
+        for article in self._build_article_iterator(
+            tuple(fitting_publisher or self.publishers), error_handling, build_extraction_filter(), url_filter
+        ):
```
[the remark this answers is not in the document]
MaxDall: Agree, changed it accordingly.

src/fundus/scraping/crawler.py
```diff
@@ -106,10 +114,220 @@ def pool_queue_iter(handle: MapResult[Any], queue: Queue[_T]) -> Iterator[_T]:
...
+                downloads. You can set a delay directly using float or any callable satisfying the Delay
+                protocol. If set to None, no delay will be used between batches. See Delay for more
+                information. Defaults to None.
+            threading (bool): If True, the crawler will use a dedicated thread per publisher, if set to False,
+                the crawler will use a single thread for a publishers and load articles succesively. This will greatly
+                influence performance, and it is highly recommended to use a threaded crawler. Deafults to True.
+        """
+
+        if not publishers:
+            raise ValueError("param <publishers> of <Crawler.__init__> has to be non empty")
```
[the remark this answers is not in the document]
MaxDall: Good catch ;)

src/fundus/scraping/html.py
MaxDall: The `abstractmethod` is a leftover.

src/fundus/scraping/html.py
MaxDall: IMO since we are utilizing `multiprocessing` here, using the handled global session could be more confusing, suggesting that all processes would use the same object, when instead, they would all copy a process local version of the handled global session and operate on different objects.

scripts/generate_parser_test_files.py
MaxDall: That is in fact true 😄 hmm, somehow I think the else makes the code more readable. What do you think?

src/fundus/scraping/crawler.py
MaxDall: Yes :)

src/fundus/scraping/html.py
MaxDall: Good point, I implemented the changes within this PR.

src/fundus/scraping/crawler.py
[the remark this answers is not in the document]
dobbersc: > I still had in my mind that both share a significant flaw that requires one element from each iterator to be retrieved before yielding the first one

Interesting, I did not know that. Sounds like a weird implementation detail. Great, that roundrobin works.

src/fundus/scraping/html.py
dobbersc: Looks good!

src/fundus/scraping/html.py
dobbersc: Yes, that makes sense. Thanks for the clarification.

scripts/generate_parser_test_files.py
dobbersc: I guess it's preference regarding what is more readable. IMO less indentation usually results in more readable code. Ruff also does this transformation automatically, if we decide to use it someday: https://docs.astral.sh/ruff/rules/superfluous-else-return/

But in the end, this is not critical, leave it in if you like the else more.

src/fundus/scraping/session.py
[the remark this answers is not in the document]
MaxDall: - Yes, the connection pool caches established connections per host, and with every run, we utilize len(publishers) hosts. In this context, "Host" covers subdomains as well.
- Yes, the 0 is a type. It's supposed to be a 1 because there is only one thread per host active and we never encounter a situation where two threads want to connect to the same host.

scripts/generate_parser_test_files.py
MaxDall: I agree. You had me with `ruff`.

src/fundus/scraping/session.py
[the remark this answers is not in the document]
dobbersc: Very nice :)

dobbersc (approved):

addie9800 (approved):
Looks great, thanks 👍

PR #358, opened 2024-02-15
Add release workflow to automatically publish on PyPi

MaxDall:
This workflow adds the following jobs that trigger on any release with [type](https://docs.github.com/en/webhooks/webhook-events-and-payloads#release) `released`.

Jobs:

1. Test/Lint using existing workflows
2. Build distribution
3. Upload to TestPyPi
4. Test installation from TestPyPi
5. Upload to PyPi

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -57,6 +57,7 @@ jobs:
     environment:
       name: testpypi
       url: https://test.pypi.org/p/fundus
+      
     permissions:
       id-token: write
 
@@ -66,6 +67,7 @@ jobs:
         with:
           name: python-package-distributions
           path: dist/
+          
       - name: Publish distribution 📦 to TestPyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
@@ -113,6 +115,7 @@ jobs:
         with:
           name: python-package-distributions
           path: dist/
+          
       - name: Publish distribution 📦 to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
 
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -1,0 +1,114 @@
+name: Publish Python 🐍 distribution 📦 to PyPI and TestPyPI
+
+on:
+  release:
+    types:
+      - released
+
+jobs:
+
+  test:
+    name: Test the latest release commit
+    uses: ./.github/workflows/tests.yml
+
+  lint:
+    name: Lint the latest release commit
+    uses: ./.github/workflows/lint.yml
+
+  build:
+    name: Build distribution 📦
+    needs:
+      - test
+      - lint
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: 'pyproject.toml'
+
+      - name: Install pypa/build
+        run: >-
+          python3 -m
+          pip install
+          build
+          --user
+
+      - name: Build a binary wheel and a source tarball
+        run: python3 -m build
+
+      - name: Store the distribution packages
+        uses: actions/upload-artifact@v3
+        with:
+          name: python-package-distribution
+          path: dist/
+
+  publish-to-testpypi:
+    name: Publish Python 🐍 distribution 📦 to TestPyPI
+    needs:
+      - build
+    runs-on: ubuntu-latest
+
+    environment:
+      name: testpypi
+      url: https://test.pypi.org/p/<package-name>fundus
+    permissions:
+      id-token: write
+
+    steps:
+      - name: Download all the dists
+        uses: actions/download-artifact@v3
+        with:
+          name: python-package-distributions
+          path: dist/
+      - name: Publish distribution 📦 to TestPyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          repository-url: https://test.pypi.org/legacy/
+
+  test-distribution:
+    name: Install and test TestPyPi distribution
+    needs:
+      - publish-to-testpypi
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: 'pyproject.toml'
+
+      - name: Install package
+        run: >- 
+          python3 -m 
+          pip install 
+          --index-url https://test.pypi.org/simple/ 
+          --extra-index-url https://pypi.org/simple/ 
+          fundus==${{ github.event.release.tag_name }}
+
+  publish-to-pypi:
+    name: Publish Python 🐍 distribution 📦 to PyPI
+    if: startsWith(github.ref, 'refs/tags/')
+
+    needs:
+    - test-distribution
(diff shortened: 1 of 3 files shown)


MaxDall:
I tested the action on a separate public test repository I created. You can find the corresponding run [here](https://github.com/MaxDall/test-commitizen/actions/runs/8436800068)

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,118 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: 'pyproject.toml'
+
+      - name: Install package
+        run: >- 
+          python3 -m 
+          pip install 
+          --index-url https://test.pypi.org/simple/ 
+          --extra-index-url https://pypi.org/simple/
```
dobbersc, line 92: Is it intended to provide the extra index URL as the regular pypi distribution? I would have thought that the `test-distribution` job should only install fundus from testpypi and not from the regular pypi repository.

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,118 @@
...
+        run: >- 
+          python3 -m 
+          pip install 
+          --index-url https://test.pypi.org/simple/ 
+          --extra-index-url https://pypi.org/simple/ 
+          fundus==${{ github.event.release.tag_name }}
+
+  publish-to-pypi:
+    name: Publish Python 🐍 distribution 📦 to PyPI
+    if: startsWith(github.ref, 'refs/tags/')
```
dobbersc, line 97: Why is this check in the standard pypi publishing job but not in the one to testpypi? I thought it would not be necessary but maybe it is just as check to be sure?

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,118 @@
...
+      id-token: write
+
+    steps:
+      - name: Download all the dists
+        uses: actions/download-artifact@v3
+        with:
+          name: python-package-distributions
+          path: dist/
+      - name: Publish distribution 📦 to TestPyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
```
dobbersc, line 70: Just to be consistent with the other steps.

```suggestion
          path: dist/
          
      - name: Publish distribution 📦 to TestPyPI
        uses: pypa/gh-action-pypi-publish@release/v1
```

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,118 @@
...
+      id-token: write
+
+    steps:
+      - name: Download all the dists
+        uses: actions/download-artifact@v3
+        with:
+          name: python-package-distributions
+          path: dist/
+      - name: Publish distribution 📦 to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
```
dobbersc, line 117:
```suggestion
          path: dist/
          
      - name: Publish distribution 📦 to PyPI
        uses: pypa/gh-action-pypi-publish@release/v1
```

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,118 @@
...
+    name: Publish Python 🐍 distribution 📦 to TestPyPI
+    needs:
+      - build
+    runs-on: ubuntu-latest
+
+    environment:
+      name: testpypi
+      url: https://test.pypi.org/p/fundus
+    permissions:
+      id-token: write
```
dobbersc, line 61:
```suggestion
      url: https://test.pypi.org/p/fundus
      
    permissions:
      id-token: write
```

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,118 @@
...
+
+    steps:
+      - name: Download all the dists
+        uses: actions/download-artifact@v3
+        with:
+          name: python-package-distributions
+          path: dist/
+      - name: Publish distribution 📦 to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+
```
dobbersc, line 118: There are also a lot of deprecation notices on the [test run](https://github.com/MaxDall/test-commitizen/actions/runs/8436800068). But fixing them would be for another PR.

dobbersc (changes requested):
This looks really useful! I have just some minor comments / questions.

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,118 @@
...
+        run: python3 -m build
+
+      - name: Store the distribution packages
+        uses: actions/upload-artifact@v3
+        with:
+          name: python-package-distributions
+          path: dist/
+
+  publish-to-testpypi:
+    name: Publish Python 🐍 distribution 📦 to TestPyPI
```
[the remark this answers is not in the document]
MaxDall: Yes, that's basically it.

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,118 @@
...
+          --index-url https://test.pypi.org/simple/ 
+          --extra-index-url https://pypi.org/simple/ 
+          fundus==${{ github.event.release.tag_name }}
+
+  publish-to-pypi:
+    name: Publish Python 🐍 distribution 📦 to PyPI
+    if: startsWith(github.ref, 'refs/tags/')
+
+    needs:
+    - test-distribution
```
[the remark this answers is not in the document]
MaxDall: You're right!

.github/workflows/publish-package.yml
MaxDall: If we specify an `index-url` also site-packages (dependencies) are resolved using the same index, so we also need the `extra-index-url` to tell pip to look on `PyPi` for everything extra then the package itself.

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,118 @@
...
+      - test
+      - lint
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: 'pyproject.toml'
```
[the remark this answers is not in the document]
dobbersc: Yea, I'm not sure there as well. I would just I the minimal version we support, e.g. 3.8.

.github/workflows/publish-package.yml
MaxDall: It isn't. Seems like I forgot to remove it.

.github/workflows/publish-package.yml
dobbersc: Oh, thanks makes sense. Thanks for the explanation.

.github/workflows/publish-package.yml
MaxDall: Yeah, I also noticed them but wasn't motivated enough to check if a simple version bump for all actions would solve the problem or if it's on the runners side.

.github/workflows/publish-package.yml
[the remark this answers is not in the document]
MaxDall: I pinned the version to `3.8`

dobbersc (approved):

.github/workflows/publish-package.yml
dobbersc: Lets just see how it is in future runs in the main repository and try a version bump later. Maybe open an issue for that?

dobbersc (approved):

PR #360, opened 2024-02-17
Refactor `ExtractionFilter` and `Requires`

MaxDall:
This is mainly to be able to log which attributes exactly are missing in the extraction. I also added some missing doc strings to the `filter` module.


addie9800 (approved):
Thanks for adding this

PR #361, opened 2024-02-17
Implement module scoped logging

MaxDall:
Since #357 rewrites a huge part of the core pipeline, this PR is based on the redesign and should be merged afterwards.

This removes the central logging solution of `basic_logger` and implements module-scoped logging. Every module utilizing logging must now have its own logger called `__module_logger__`.
```python
from fundus.logging import create_logger

__module_logger__ = create_logger(__name__)
```
One can set a global log level
```python
import logging
from fundus.logging import set_log_level

set_log_level(logging.DEBUG)
```
effecting only loggers created with `create_logger`, or setting the log level of individual module-scoped loggers by importing them.
```python
import logging
from fundus.scraping.html import __module_logger__

__module_logger__.setLevel(logging.DEBUG)
```


src/fundus/logging/__init__.py
dobbersc, line 1: Since there is no other file in the `logging` folder, I think we can move the contents to a `logging.py` file.

src/fundus/parser/data.py
```diff
@@ -16,10 +16,13 @@
 
 from typing_extensions import TypeAlias
 
-from fundus.logging import basic_logger
+from fundus.logging import create_logger
+
+__module_logger__ = create_logger(__name__)
```
dobbersc, line 21: This is preference, but I would keep the variable name for the logger simple. Just `logger` should be enough. IMO, seeing the `__` everywhere clutters the code, as the logger is not related to any magic/dunder methods.

src/fundus/logging/__init__.py
```diff
@@ -1,3 +1,22 @@
...
+    logger = logging.getLogger(name)
+    logger.setLevel(logging.ERROR)
+    logger.addHandler(_stream_handler)
+    _loggers.append(logger)
+    return logger
+
+
+def set_log_level(level: int):
+    for logger in _loggers:
+        logger.setLevel(level)
```
dobbersc, line 22: This definitely works. But I wondered if such a setup is possible where each module can just call `logger.getLogger(__name__)` to get their module logger, just like in the python [logging guide](https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial). I failed to get this to work but wanted to give the idea here. Maybe you tried this as well?

dobbersc (changes requested):

src/fundus/logging/__init__.py
MaxDall: Good idea. The only problem is see is how we set the log level for all Fundus logger at once? Setting them individually wouldn't be a problem, but I have no solution for how we would set the whole library to `DEBUG` without keeping track of the loggers.

dobbersc (approved):

PR #362, opened 2024-02-18
Add free access atrribute

addie9800:
As discussed, I have added the free_access attribute natively to all publishers and overwriting it in the, to my knowledge, only special case (Bild).

### DIFF
--- a/docs/attribute_guidelines.md
+++ b/docs/attribute_guidelines.md
@@ -62,7 +62,7 @@ Those attributes will be validated with unit tests when used.
         <td>free_access</td>
         <td>A boolean which is set to be False, if the article is restricted to users with a subscription. This usually indicates
         that the article cannot be crawled completely.
-        <i>This attribute is implemented by default</i></td>
+        <i><b>This attribute is implemented by default</b></i></td>
         <td><code>bool</code></td>
         <td><code></code></td>
     </tr>
--- a/docs/attribute_guidelines.md
+++ b/docs/attribute_guidelines.md
@@ -58,4 +58,12 @@ Those attributes will be validated with unit tests when used.
         <td><code>List[str]</code></td>
         <td><code>generic_topic_parsing</code></td>
     </tr>
+    <tr>
+        <td>free_access</td>
+        <td>A boolean which is set to be True, if the article is restricted to users with a subscription. This usually indicates
+        that the article cannot be crawled completely.
+        <i>This attribute is implemented by default</i></td>
+        <td><code>bool</code></td>
+        <td><code></code></td>
+    </tr>
 </table>
--- a/docs/attribute_guidelines.md
+++ b/docs/attribute_guidelines.md
@@ -60,7 +60,7 @@ Those attributes will be validated with unit tests when used.
     </tr>
     <tr>
         <td>free_access</td>
-        <td>A boolean which is set to be True, if the article is restricted to users with a subscription. This usually indicates
+        <td>A boolean which is set to be False, if the article is restricted to users with a subscription. This usually indicates
         that the article cannot be crawled completely.
         <i>This attribute is implemented by default</i></td>
         <td><code>bool</code></td>
--- a/docs/how_to_add_a_publisher.md
+++ b/docs/how_to_add_a_publisher.md
@@ -483,7 +483,7 @@ function by adding the following snippet to your parser:
 @attribute
 def free_access(self) -> bool:
     # Your personalized logic goes here
-    pass
+    ...
 ```
 
 Usually you can identify a premium article by an indicator within the URL or by using XPath or CSSSelector and selecting
--- a/docs/how_to_add_a_publisher.md
+++ b/docs/how_to_add_a_publisher.md
@@ -16,6 +16,7 @@
     * [Working with `lxml`](#working-with-lxml)
     * [CSS-Select](#css-select)
     * [XPath](#xpath)
+  * [Checking the free_access attribute](#checking-the-free_access-attribute)
   * [Finishing the Parser](#finishing-the-parser)
 * [6. Generate unit tests](#6-generate-unit-tests)
 * [7. Opening a Pull Request](#7-opening-a-pull-request)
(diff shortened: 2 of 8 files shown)


docs/how_to_add_a_publisher.md
```diff
@@ -469,6 +469,24 @@ Instead, we recommend referring to [this](https://devhints.io/xpath) documentati
...
+please verify that there is a tag `isAccessibleForFree` within the `<script type="application/ld+json">` blocks in the
+source code of premium articles that is set to either `false` or `False`. It doesn't matter if the tag is missing in the
+freely accessible articles. If this is the case, you can continue with the next step. If not, please overwrite the
+existing function by adding the following snippet to your parser:
+
+```python
+@attribute
+def free_access(self) -> bool:
+    # Your personalized logic goes here
+    pass
```
MaxDall, line 484:
```suggestion
    ...
```

docs/how_to_add_a_publisher.md
```diff
@@ -469,6 +469,24 @@ Instead, we recommend referring to [this](https://devhints.io/xpath) documentati
 Make sure to examine other parsers and consult the [attribute guidelines](attribute_guidelines.md) for specifics on attribute implementation. 
 We strongly encourage utilizing these utility functions, especially when parsing the `ArticleBody`.
 
+### Checking the free_access attribute
+
+In case your new publisher does not have a subscription model, you can go ahead and skip this step. If it does,
+please verify that there is a tag `isAccessibleForFree` within the `<script type="application/ld+json">` blocks in the
+source code of premium articles that is set to either `false` or `False`. It doesn't matter if the tag is missing in the
+freely accessible articles. If this is the case, you can continue with the next step. If not, please overwrite the
+existing function by adding the following snippet to your parser:
```
MaxDall, line 478: I think in general this is a good introduction to the attribute. Just two things:
1. Rather than referring to `<script ... ld+json">` I would lead the user to the `ld` attribute of the article.
2. This section is missing in the Table of contents

src/fundus/parser/base_parser.py
```diff
@@ -172,7 +172,7 @@ def _search_members(cls, obj_type: type) -> List[Tuple[str, Any]]:
     @classmethod
     def attributes(cls) -> AttributeCollection:
         attrs: List[Attribute] = [
-            func for _, func in cls._search_members(Attribute) if func.__name__ not in ["__ld", "__meta"]
+            func for _, func in cls._search_members(Attribute) if func.__name__ not in ["__ld", "__meta", "free_access"]
```
MaxDall, line 175: `free_access` should not be excluded here because it is a required attribute from the `attribute_guidelines`. If you check your branch's [`supported_publishers.md`](https://github.com/flairNLP/fundus/blob/add-free-access-atrribute/docs/supported_publishers.md) you will see that is flagged as missing for every publisher.

src/fundus/scraping/scraper.py
```diff
@@ -34,6 +34,7 @@ async def scrape(
             supported_attributes = set(
                 more_itertools.flatten(collection.names for collection in self.parser.attribute_mapping.values())
             )
+            supported_attributes.add("free_access")
```
MaxDall, line 37: This should be removed and was only necessary because the attribute was excluded in the first place.

MaxDall (changes requested):
Thanks @addie9800  this is a really nice addition to the parser in general. Also good job with keeping the documentation clean and easy.

src/fundus/scraping/scraper.py
addie9800: It's weird I thought I needed to add it to keep the BaseParser attribute tests from failing, but now, after I removed it, everything still works fine

src/fundus/parser/base_parser.py
addie9800: I had to add this here, because otherwise the tests for the BaseParser would fail due to it expecting only 2 attributes. Should I instead modify the tests?

MaxDall: Hmm, I'm undecided about what I want this to be like. On the one hand `BaseParser` was never supposed to support attributes also present in the guidelines natively and adding them somehow dilutes the design. On the other hand, manually adding free access to every parser is redundant and would serve no purpose.

The original design Idea was to have standardized attributes described in the `attribute_guidelines` which every parser had to add individually so that one could differentiate between the parser not supporting an attribute like `topics` and the parser no longer being able to parse said attribute.

`free_access` fits none of the above cases:
1. Being boolean, there is no information gained to know if it is supported by a specific parser or not. If it's not, simply yielding `True` already provides every piece of information.
2. It's not as standardized as the `meta` tag or `ld+json` section as we have seen with `Bild` so there is a need to support individual implementations.

I guess adding it to the guideline. treating it as every other attribute there while having a default in the `BaseParser`, and altering the tests would be a solution here.

addie9800: Ok, sounds good to me. I have updated the code.

docs/how_to_add_a_publisher.md
```diff
@@ -469,6 +470,25 @@ Instead, we recommend referring to [this](https://devhints.io/xpath) documentati
...
+
+```python
+@attribute
+def free_access(self) -> bool:
+    # Your personalized logic goes here
+    ...
+```
+
+Usually you can identify a premium article by an indicator within the URL or by using XPath or CSSSelector and selecting
+the element asking to to purchase a subscription to view the article.
```
MaxDall, line 490: Sorry, I forgot to mention this. We agreed on some standardization within the markdown documents having every sentence in a different line. Could you change this part accordingly?

MaxDall (changes requested):

docs/how_to_add_a_publisher.md
addie9800: Oh ok, makes sense. I have updated the relevant part.

docs/attribute_guidelines.md
```diff
@@ -58,4 +58,12 @@ Those attributes will be validated with unit tests when used.
         <td><code>List[str]</code></td>
         <td><code>generic_topic_parsing</code></td>
     </tr>
+    <tr>
+        <td>free_access</td>
+        <td>A boolean which is set to be False, if the article is restricted to users with a subscription. This usually indicates
+        that the article cannot be crawled completely.
+        <i>This attribute is implemented by default</i></td>
```
MaxDall, line 65:
```suggestion
        <i><b>This attribute is implemented by default</b></i></td>
```

docs/how_to_add_a_publisher.md
```diff
@@ -469,6 +470,23 @@ Instead, we recommend referring to [this](https://devhints.io/xpath) documentati
 Make sure to examine other parsers and consult the [attribute guidelines](attribute_guidelines.md) for specifics on attribute implementation. 
 We strongly encourage utilizing these utility functions, especially when parsing the `ArticleBody`.
 
+### Checking the free_access attribute
+
+In case your new publisher does not have a subscription model, you can go ahead and skip this step.
+If it does, please verify that there is a tag `isAccessibleForFree` within the HTMLs `ld+json` elements (refer to the section [Extracting attributes from Precomputed](#extracting-attributes-from-precomputed) for details) in the source code of premium articles that is set to either `false` or `False`.
```
MaxDall, line 476:
```suggestion
If it does, please verify that there is a tag `isAccessibleForFree` within the HTMLs `ld+json` elements (refer to the section [Extracting attributes from Precomputed](#extracting-attributes-from-precomputed) for details) in the source code of premium articles that is set to either `false` or `False`,  `true`/`True` respectively.
```

MaxDall (changes requested):
Just two minor things in the documentation and then this is good to go 🎉

MaxDall (approved):

PR #364, opened 2024-02-21
Add URL parameter to test generation script

MaxDall:
This adds a new parameter `-u` to `scripts.generate_parser_test_files`
This parameter enables one to specify a URL to be used as a test case, see #333 
The `-p` parameter is necessary if URL(s) are given and the order of given publishers corresponds to the order of given URL(s).

### DIFF
--- a/scripts/generate_parser_test_files.py
+++ b/scripts/generate_parser_test_files.py
@@ -26,7 +26,7 @@ def get_test_article(enum: PublisherEnum, url: Optional[str] = None) -> Optional
     return next(crawler.crawl(max_articles=1, error_handling="suppress", only_complete=True), None)
 
 
-if __name__ == "__main__":
+def main() -> None:
     parser = ArgumentParser(
         prog="generate_parser_test_files",
         description=(
@@ -124,3 +124,7 @@ if __name__ == "__main__":
             test_data_file.write(test_data)
             bar.update()
             subprocess.call(["git", "add", test_data_file.path], stdout=subprocess.PIPE)
+
+
+if __name__ == "__main__":
+    main()
--- a/scripts/generate_parser_test_files.py
+++ b/scripts/generate_parser_test_files.py
@@ -1,5 +1,5 @@
 import subprocess
-from argparse import ArgumentParser
+from argparse import ArgumentParser, Namespace
 from logging import WARN
 from typing import List, Optional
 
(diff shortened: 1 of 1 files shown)


scripts/generate_parser_test_files.py
```diff
@@ -31,9 +39,17 @@ def get_test_article(enum: PublisherEnum) -> Optional[Article]:
         "attributes",
         metavar="A",
         nargs="*",
-        help=f"the attributes which should be used to create test cases. default: {', '.join(attributes_required_to_cover)}",
+        help=f"the attributes which should be used to create test cases. "
+        f"default: {', '.join(attributes_required_to_cover)}",
```
dobbersc, line 43:
```suggestion
        help=(
            "the attributes which should be used to create test cases. "
            f"default: {', '.join(attributes_required_to_cover)}"
        ),
```

scripts/generate_parser_test_files.py
```diff
@@ -31,9 +39,17 @@ def get_test_article(enum: PublisherEnum) -> Optional[Article]:
...
     )
     parser.add_argument("-p", dest="publishers", metavar="P", nargs="+", help="only consider given publishers")
+    parser.add_argument(
+        "-u",
+        dest="urls",
```
dobbersc, line 48: To keep it consistent with the other command line arguments and offer a long version.

```suggestion
        "--urls",
```

scripts/generate_parser_test_files.py
```diff
@@ -5,16 +5,24 @@
...
 from fundus.scraping.article import Article
+from fundus.scraping.html import FundusSource
+from fundus.scraping.scraper import Scraper
 from tests.test_parser import attributes_required_to_cover
 from tests.utility import HTMLTestFile, get_test_case_json, load_html_test_file_mapping
 
 
-def get_test_article(enum: PublisherEnum) -> Optional[Article]:
-    crawler = Crawler(enum)
+def get_test_article(enum: PublisherEnum, url: Optional[str] = None) -> Optional[Article]:
```
dobbersc, line 18: This is not critical for now but `url` shadows the name from the outer scope. It would be nice anyway to wrap the main script into a main function to avoid global exposure.

dobbersc (changes requested):
Thanks for adding this. Just some very minor comments:

scripts/generate_parser_test_files.py
MaxDall: Good point, i restructured the script a bit.

dobbersc (approved):

PR #366, opened 2024-02-22
Adjust paragraph selector for CNBC

MaxDall:
Adjusts the CNBC paragraph selector slightly for better quality.

Example articles:
https://www.cnbc.com/2023/04/28/many-new-bitcoin-crypto-buyers-influenced-by-friends-why-to-be-cautious.html
https://www.cnbc.com/2024/02/22/americans-with-debt-putting-less-money-toward-credit-card-payments.html

### DIFF
--- a/src/fundus/publishers/us/cnbc.py
+++ b/src/fundus/publishers/us/cnbc.py
@@ -2,6 +2,7 @@ import datetime
 from typing import List, Optional
 
 from lxml.cssselect import CSSSelector
+from lxml.etree import XPath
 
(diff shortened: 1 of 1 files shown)


addie9800 (approved):
Thanks for updating CNBC 👍 I did find an article, which had something along the lines of "More Articles from CNBC Pro:" in their plaintext, but since it was in the same block as the last paragraph, I doubt there's a way to remove it without too much hassle. I did want to meintion it though: https://www.cnbc.com/2024/02/23/credit-card-interest-rates-are-at-record-highs.html

PR #367, opened 2024-02-22
Adjust paragraph selector for gateway pundit

MaxDall:


addie9800:
Currently the titles are parsed as `Hunter Biden's Lawyers Lash Out at Special Counsel, Claim Photos on iPhone Are Sawdust, Not Crack Cocaine | The Gateway Pundit | by Cristina Laila`. I was wondering whether it would make sense to cut off the extra bits after the `|`

src/fundus/publishers/us/the_gateway_pundit.py
```diff
@@ -13,7 +13,12 @@
 
 class TheGatewayPunditParser(ParserProxy):
     class V1(BaseParser):
-        _paragraph_selector = CSSSelector("div.entry-content > p")
+        _related = r"^Click\s$"
```
addie9800, line 16: Copyright at the end is parsed into the plaintext e.g. https://www.thegatewaypundit.com/2024/02/report-biden-tapping-brakes-push-evs/ or https://www.thegatewaypundit.com/2024/02/state-supreme-court-rules-embryos-are-children/

The plaintext for this article: https://www.thegatewaypundit.com/2024/02/kid-rocks-father-passes-away-president-trump-responds/ ends in `WATCH:` and refers to a video

It's similar here with `Watch the video:` https://www.thegatewaypundit.com/2024/02/dr-phil-blasts-medical-associations-pushing-gender-affirming/ but since it's not in any special block, I'm not sure there's an easy fix.

In this article the parser collected some JavaScript Script I think, pretty weird. https://www.thegatewaypundit.com/2024/02/elon-musk-nominated-nobel-peace-prize-defending-free/ 
`Via George Behizy. !function(r,u,m,b,l,e){r._Rumble=b,r[b]||(r[b]=function(){(r[b]._=r[b]._||[]).push(arguments);if(r[b]._.length==1){l=u.createElement(m),e=u.getElementsByTagName(m)[0],l.async=1,l.src="https://rumble.com/embedJS/u5o49d"+(arguments[1].video?'.'+arguments[1].video:'')+"/?url="+encodeURIComponent(location.href)+"&args="+encodeURIComponent(JSON.stringify([].slice.apply(arguments))),e.parentNode.insertBefore(l,e)}})}(window, document, "script", "Rumble")` 
Similar here: https://www.thegatewaypundit.com/2024/02/fix-is-new-york-judge-merchan-upcoming-ny/

addie9800 (changes requested):
Thanks for updating this 👍

src/fundus/publishers/us/the_gateway_pundit.py
MaxDall: Thanks for catching these edge cases. 

About the copyright and `WATCH:` part: 
IMO it's not easy to tell if that's part of the article body or not, or if we should parse it. There are arguments for both sides. At this point, it's solely dependent on our definition of what's part of the body and what's not. The problem is that we never defined this, leaving us with coherence problems, etc. We need to discuss this, but not in this PR and I would leave it as it is.

I will alter the selector to exclude nodes containing `<script>` tags.

addie9800 (dismissed):

src/fundus/publishers/us/the_gateway_pundit.py
addie9800: Sounds good, I agree with that, I also wasn't sure, but wanted to mention it. Btw,_ I forgot to address this in the new review, but what do think in regard to the titles, from the comment above?

PR #368, opened 2024-02-22
Adjust paragraph selector for Fox News parser

MaxDall:
This adds a new `summary` selector as well as including paragraphs with optional `div.paywall` parent node.

### DIFF
--- a/src/fundus/publishers/us/fox_news.py
+++ b/src/fundus/publishers/us/fox_news.py
@@ -1,7 +1,7 @@
 import datetime
 from typing import List, Optional
 
-from lxml.cssselect import CSSSelector
+from lxml.etree import XPath
 
 from fundus.parser import ArticleBody, BaseParser, ParserProxy, attribute
 from fundus.parser.utility import (
@@ -14,7 +14,7 @@ from fundus.parser.utility import (
 
 class FoxNewsParser(ParserProxy):
     class V1(BaseParser):
-        _paragraph_selector = CSSSelector(".article-body > p")
+        _paragraph_selector = XPath("//div[@class='article-body'] / p[text()]")
 
         @attribute
         def body(self) -> ArticleBody:
--- a/src/fundus/publishers/us/fox_news.py
+++ b/src/fundus/publishers/us/fox_news.py
@@ -17,7 +17,7 @@ class FoxNewsParser(ParserProxy):
     class V1(BaseParser):
         _summary_selector = CSSSelector(".article-meta > h2")
         _paragraph_selector = XPath(
-            "(//div[@class='article-body'] | //div[@class='article-body']/div[@class='paywall']) "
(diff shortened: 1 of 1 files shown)


src/fundus/publishers/us/fox_news.py
```diff
@@ -14,7 +14,7 @@
 
 class FoxNewsParser(ParserProxy):
     class V1(BaseParser):
-        _paragraph_selector = CSSSelector(".article-body > p")
+        _paragraph_selector = XPath("//div[@class='article-body'] / p[text()]")
```
addie9800, line 17: The new selector does not work on some articles: e.g. https://www.foxnews.com/politics/speaker-johnson-slams-desperate-biden-calling-gop-worse-segregationists-fundraiser and https://www.foxnews.com/world/putin-puts-west-notice-flight-nuclear-capable-bomber
It's a bit weird though, because they seem to be premium articles, but I am still able to access them, even from Edge where I don't have any blockers installed. The use a slightly different class in the div block 'article-content'

src/fundus/publishers/us/fox_news.py
```diff
@@ -14,7 +14,7 @@
...
 
         @attribute
         def body(self) -> ArticleBody:
```
addie9800, line 20: I think we could add a summary/subtitle selector for FoxNews as well

addie9800 (changes requested):
Thanks for updating Fox :)

src/fundus/publishers/us/fox_news.py
MaxDall: Good point, i will add this as wel.

src/fundus/publishers/us/fox_news.py
```diff
@@ -14,7 +15,11 @@
 
 class FoxNewsParser(ParserProxy):
     class V1(BaseParser):
-        _paragraph_selector = CSSSelector(".article-body > p")
+        _summary_selector = CSSSelector(".article-meta > h2")
+        _paragraph_selector = XPath(
+            "(//div[@class='article-body'] | //div[@class='article-body']/div[@class='paywall']) "
+            "/p[not(child::script) and text()]"
```
addie9800, line 21:
```suggestion
            "//p[not(child::script) and text()]"
```
For some reason within fundus it needs the extra slash

addie9800 (changes requested):

src/fundus/publishers/us/fox_news.py
MaxDall: That's exactly what we don't want to have. Having the extra slash will select all `p` not only those following but also those  just being after `(//div[@class='article-body'] | //div[@class='article-body']/div[@class='paywall'])` in the DOM tree, collecting a lot of junk in the process.

addie9800: Ok, I didn't see that when testing. But without the extra slash, the "premium" articles plaintext comes back as None

PR #369, opened 2024-02-23
Fix SZ selector style and type hint

MaxDall:


addie9800 (approved):
I feel you, I had the same issue some time ago :)

PR #370, opened 2024-02-23
Updated MDR Parser

addie9800:


src/fundus/publishers/de/mdr.py
```diff
@@ -17,7 +18,15 @@
...
     class V1(BaseParser):
         _author_substitution_pattern: Pattern[str] = re.compile(r"MDR \w*$|MDR \w*-\w*$|MDRfragt-Redaktionsteam|^von")
-        _paragraph_selector = CSSSelector("div.paragraph")
+        _source_detection: str = (
+            r"((MDR AKTUELL \(ans\))|(Quell(e|en): [A-z\.]{3,4})|(\([A-z]{1,4}\))|([A-z]{3}/[A-z]{3}))"
+        )
+        _paragraph_selector = XPath(
+            f"//div[@class='paragraph '"
+            f" and not(div[@class='mediaCon mediaLeft mediaSizeA cssImage hasNoRessort item-delegated-lightbox'])"
+            f" ]//p[not(re:test(em, '{_source_detection}') or re:test(text(), '{_source_detection}'))]",
```
MaxDall, line 27:
```suggestion
            f"//div[@class='paragraph '] "
            f"/p[not(re:test(em, '{_source_detection}') or re:test(text(), '{_source_detection}'))]",
```

MaxDall: Wouldn't this be much easier?

src/fundus/publishers/de/mdr.py
```diff
@@ -17,7 +18,15 @@
 class MDRParser(ParserProxy):
     class V1(BaseParser):
         _author_substitution_pattern: Pattern[str] = re.compile(r"MDR \w*$|MDR \w*-\w*$|MDRfragt-Redaktionsteam|^von")
-        _paragraph_selector = CSSSelector("div.paragraph")
+        _source_detection: str = (
+            r"((MDR AKTUELL \(ans\))|(Quell(e|en): [A-z\.]{3,4})|(\([A-z]{1,4}\))|([A-z]{3}/[A-z]{3}))"
+        )
```
MaxDall, line 23: Could you provide an example for this? Best would be actually some regex101 example to show how the regex is supposed to work. IMO reading regexes it's always really hard to understand what's going on without having an example.

MaxDall (changes requested):

src/fundus/publishers/de/mdr.py
addie9800: True, I don't remember exactly why I added that

src/fundus/publishers/de/mdr.py
addie9800: I have now updated the regex and added a regex101 link

MaxDall: Thanks 👍 The regex does not match within the next to last example. Is this on purpose?

addie9800: Yes, that is intended. That case was accidentally filtered out in a previous version of the regex.

PR #371, opened 2024-02-23
Fiz zeit paragraph selector

addie9800:
I updated the paragraph selector, but I didn't add a new parser version due to the new XPath variant being more or less equivalent to the previous CSSSelector().


MaxDall:
> I updated the paragraph selector, but I didn't add a new parser version due to the new XPath variant being more or less equivalent to the previous CSSSelector().

Thanks! You should only bump a parser version if the Layout changes on publisher side, so right decission.

MaxDall (approved):

PR #376, opened 2024-02-24
BusinessInsider: multiple Updates

addie9800:
I have updated the title parser due to the former version using the <title> tags value which doesn't coincide with the actual headline of the article. 
I have also addressed the issue of external contents being parsed with errors


MaxDall:
Hey @addie9800 thanks for fixing this 👍 Could you explain to me why this PR also adds a new test case for the publisher?

addie9800:
> Hey @addie9800 thanks for fixing this 👍 Could you explain to me why this PR also adds a new test case for the publisher?

I just realized the mistake in my train of thought: My thought was that with the updated selector, we would also need to update the testcase, because I was thinking of the version in the PR with the body added to the unit tests.

MaxDall:
@addie9800 Could you open your changes in a new PR that does not alter the test case?

MaxDall:
@addie9800 Thanks a lot. I will close this now.

PR #379, opened 2024-02-26
Fix `_summary_selector` for `TheNewYorker`

MaxDall:


addie9800 (changes requested):
Thanks for updating TheNewYorker

addie9800 (approved):
Thanks for updating TheNewYorker

PR #382, opened 2024-02-27
Add functionality to exclude tags from extraction and normalize space

MaxDall:
As #338 showed (see Focus/Nation) sometimes JS is extracted as text into the article. This is due to Fundus using a broad XPath expression `string()` to extract text from nodes. While this makes it fast and easy to extract text from subsequent nodes, it does not allow Fundus to skip tag types (like `<script>`) during text extraction.

To address this issue this PR implements:
- a new method `text_content(excluded_tags: List[str])` defined on `Node` to extract text only from non-excluded nodes down the path
- furthermore this PR adds a normalization function `normalize-space` to normalize spacing within all extracted text. Previously, extracted text was in a weird state between `raw` and `normalized`. The normalization is done with `' '.join(.split())`

### DIFF
--- a/src/fundus/parser/utility.py
+++ b/src/fundus/parser/utility.py
@@ -41,10 +41,10 @@ class Node:
     # //*[not(self::script) and text()]/text(), but for whatever reason, that's actually 50-150% slower
     # than simply using the implemented mixture below
     def text_content(self, excluded_tags: Optional[List[str]] = None) -> str:
-        excluded_tags = excluded_tags or []
+        guarded_excluded_tags: List[str] = excluded_tags or []
 
         def _text_content(element: lxml.html.HtmlElement) -> str:
-            if element.tag in excluded_tags:
+            if element.tag in guarded_excluded_tags:
                 return ""
             text = element.text or "" if not isinstance(element, lxml.html.HtmlComment) else ""
             children = "".join([_text_content(child) for child in element.iterchildren()]) or ""
--- a/src/fundus/parser/utility.py
+++ b/src/fundus/parser/utility.py
@@ -122,7 +122,9 @@ def extract_article_body_with_selector(
         first = next(instructions)
         instructions = itertools.chain([first, []], instructions)
 
-    summary = TextSequence(map(lambda x: normalize_whitespace(x.text_content(excluded_tags=["script"])), next(instructions)))
+    summary = TextSequence(
+        map(lambda x: normalize_whitespace(x.text_content(excluded_tags=["script"])), next(instructions))
+    )
     sections: List[ArticleSection] = []
 
     for chunk in more_itertools.chunked(instructions, 2):
--- a/src/fundus/parser/utility.py
+++ b/src/fundus/parser/utility.py
@@ -17,7 +17,6 @@ from typing import (
     cast,
 )
 
-import dateutil.tz
 import lxml.html
 import more_itertools
 from dateutil import parser
@@ -27,6 +26,10 @@ from lxml.etree import XPath
 from fundus.parser.data import ArticleBody, ArticleSection, TextSequence
 
 
+def normalize_whitespace(text: str) -> str:
+    return " ".join(text.split())
+
+
 @total_ordering
 @dataclass(eq=False)
 class Node:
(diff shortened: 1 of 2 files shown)


src/fundus/parser/utility.py
```diff
@@ -27,15 +26,32 @@
...
     position: int
     node: lxml.html.HtmlElement = field(compare=False)
     _break_selector: ClassVar[XPath] = XPath("*//br")
 
-    def striped(self, chars: Optional[str] = None) -> str:
-        return str(self).strip(chars)
+    # one could replace this recursion with XPath using an expression like this:
+    # //*[not(self::script) and text()]/text(), but for whatever reason, that's actually 50-150% slower
+    # than simply using the implemented mixture below
+    def text_content(self, excluded_tags: Optional[List[str]] = None) -> str:
```
dobbersc, line 43: Just in general to understand the problem at hand: 

> This is due to Fundus using a broad XPath expression string() to extract text from nodes.

I don't see an XPATH that uses the `string()` expression anywhere. From the previous code, there is only the `self._get_break_preserved_node().text_content()`. So does lxml use the `string()` XPATH to extract the text content internally? At least in the text content attribute in the HTML inspector for the FOCUS test article, I could see the javascript there: 
```
textContent: 'Über unsere Gutschein-Seite erhalten Sie exklusiv noch ein zweites Geschenk, wenn Sie den dort abrufbaren Code ab 150 Euro Bestellwert im Warenkorb hinterlegen.  !function(){var t=window.addEventListener?"addEventListener":"attachEvent";(0,window[t])("attachEvent"==t?"onmessage":"message",function(t){if("string"==typeof t.data&&t.data.indexOf("documentHeight:")>-1){var e=t.data.split("documentHeight:")[1],n=(e=t.data.split("documentHeight:")[1],e=parseInt(e)+30,function(t){for(var e=document.getElementsByTagName("iframe"),n=0,a=e.length;n<a;n++)if(e[n].contentWindow===t.source&&-1!=e[n].id.indexOf("chipwidget_"))return e[n];return null}(t));n&&(n.style.height=e+"px")}},!1)}();'
```

src/fundus/parser/utility.py
```diff
@@ -27,15 +26,32 @@
...
+        guarded_excluded_tags: List[str] = excluded_tags or []
+
+        def _text_content(element: lxml.html.HtmlElement) -> str:
+            if element.tag in guarded_excluded_tags:
+                return ""
+            text = element.text or "" if not isinstance(element, lxml.html.HtmlComment) else ""
+            children = "".join([_text_content(child) for child in element.iterchildren()]) or ""
```
dobbersc, line 50: Since `"".join([...])` will always return a string, I think that the `or ""` is not needed.

```suggestion
            children = "".join([_text_content(child) for child in element.iterchildren()])
```

src/fundus/parser/utility.py
```diff
@@ -27,15 +26,32 @@
...
+            tail = element.tail or ""
+            return text + children + tail
```
dobbersc, line 52:
```suggestion
            return f"{text}{children}{tail}"
```

src/fundus/parser/utility.py
```diff
@@ -106,13 +122,13 @@ def extract_nodes(selector: XPath, node_type: Type[Node]) -> List[Node]:
         first = next(instructions)
         instructions = itertools.chain([first, []], instructions)
 
-    summary = TextSequence(map(lambda x: x.striped("\n"), next(instructions)))
+    summary = TextSequence(map(lambda x: normalize_whitespace(x.text_content(["script"])), next(instructions)))
```
dobbersc, line 125: Maybe to increase the readability we could use a keyword argument here to help to immediately know that the script tag is the one to be ignored and not be extracted the text content from.

```suggestion
    summary = TextSequence(map(lambda x: normalize_whitespace(x.text_content(excluded_tags=["script"])), next(instructions)))
```

src/fundus/parser/utility.py
```diff
@@ -106,13 +122,13 @@ def extract_nodes(selector: XPath, node_type: Type[Node]) -> List[Node]:
...
     sections: List[ArticleSection] = []
 
     for chunk in more_itertools.chunked(instructions, 2):
         if len(chunk) == 1:
             chunk.append([])
-        texts = [list(map(lambda x: x.striped("\n"), c)) for c in chunk]
+        texts = [list(map(lambda x: normalize_whitespace(x.text_content(["script"])), c)) for c in chunk]
```
dobbersc, line 131:
```suggestion
        texts = [list(map(lambda x: normalize_whitespace(x.text_content(excluded_tags=["script"])), c)) for c in chunk]
```

dobbersc (changes requested):

src/fundus/parser/utility.py
```diff
@@ -27,15 +26,32 @@
...
+    # one could replace this recursion with XPath using an expression like this:
+    # //*[not(self::script) and text()]/text(), but for whatever reason, that's actually 50-150% slower
+    # than simply using the implemented mixture below
+    def text_content(self, excluded_tags: Optional[List[str]] = None) -> str:
+        guarded_excluded_tags: List[str] = excluded_tags or []
+
+        def _text_content(element: lxml.html.HtmlElement) -> str:
+            if element.tag in guarded_excluded_tags:
+                return ""
+            text = element.text or "" if not isinstance(element, lxml.html.HtmlComment) else ""
```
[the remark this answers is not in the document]
MaxDall: Me neither, but i think recognizing missing content is harder than additional so im in favor of excluding here rather then restricting.

src/fundus/parser/utility.py
MaxDall: Yes, `lxml.html.HTMLElement.text_content()` is just a wrapper for `HTMLElement.xpath("string()")`.

src/fundus/parser/utility.py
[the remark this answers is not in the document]
dobbersc: Yes, that is reasonable.

src/fundus/parser/utility.py
MaxDall: I would prefer the plus concatenation over f-strings, it's just personal preference here.

dobbersc: Ok, it's just to avoid the creation of an immediate string `text + children` with the f-string.

dobbersc (approved):

PR #384, opened 2024-02-28
Bump `TheIntercept` version to `V1_1`

MaxDall:
This updates `TheIntercept` to adapt to the latest Layout changes. Unfortunately, the Publisher Coverage Action couldn't catch when the layout change occurred. This is because the summary was parsed from the metadata and not from the body itself. This is why there was always something in the body although the layout could not be parsed anymore. I removed this part from the legacy parser and parsed the summary directly from the article body.


src/fundus/publishers/us/the_intercept.py
```diff
@@ -55,3 +58,11 @@ def topics(self) -> List[str]:
                 return []
 
             return [keyword[9:] for keyword in keywords if keyword.startswith("Subject: ")]
+
+    class V1_1(V1):
+        VALID_UNTIL = date.today()
+        _summary_selector = XPath(
+            "//p[@class='post__excerpt'] | //h2[preceding-sibling::h1[contains(@class, 'post__title')]]"
+        )
+        _paragraph_selector = CSSSelector("div.entry-content > div.entry-content__content > p")
```
addie9800, line 67: In the paragraph selector, I think we should also add p blocks that are children of a blockquote block, as it is used in this article for example: https://theintercept.com/2024/02/28/congress-pakistan-election-recognition/

addie9800 (changes requested):
Thanks for updateing this 👍

src/fundus/publishers/us/the_intercept.py
MaxDall: I added block quotes as well

addie9800 (approved):

PR #385, opened 2024-03-07
Fix malformed HTML for `TheNation`

MaxDall:
This fixes parsing for the nation once again 🎉 
- Fixes malformed HTML with missing </div> closing tag
- Fixes new version couldn't parse older articles

### DIFF
--- a/src/fundus/publishers/us/the_nation_parser.py
+++ b/src/fundus/publishers/us/the_nation_parser.py
@@ -94,3 +94,10 @@ class TheNationParser(ParserProxy):
         # remove aside function from V1
         def _remove_aside(self):
             pass
+
+        @attribute
+        def topics(self) -> List[str]:
+            if topics := generic_topic_parsing(self.precomputed.meta.get("keywords")):
+                return topics
+            else:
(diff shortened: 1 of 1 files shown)


src/fundus/publishers/us/the_nation_parser.py
```diff
@@ -63,8 +83,13 @@ def topics(self) -> List[str]:
     class V2(V1):
```
addie9800, line 83: It seems they (mostly) updated the way they assign topics. It seems to me that they stopped using the `keywords` tag, and switched to `sailthru.tags`. For this article for example, no topics were parsed: https://www.thenation.com/article/archive/gop-threatens-pepfar-hiv-aids-africa/

addie9800 (changes requested):
Thanks for updating this 👍

src/fundus/publishers/us/the_nation_parser.py
MaxDall: Thanks for catching this 👍 i altered topic extraction for `V2` accordingly

addie9800 (approved):

PR #388, opened 2024-03-08
Bump `WAZ` to version `V1_1`

MaxDall:


src/fundus/publishers/de/waz.py
```diff
@@ -44,3 +45,9 @@ def topics(self) -> List[str]:
             authors = generic_author_parsing(self.precomputed.meta.get("author"))
             topics = generic_topic_parsing(self.precomputed.meta.get("keywords"))
             return [topic for topic in topics if topic not in authors]
+
+    class V1_1(V1):
+        VALID_UNTIL = datetime.date.today()
+        _paragraph_selector = CSSSelector(".article-body > p")
```
addie9800, line 51: This falsely also extracts the summary a second time

addie9800, line 51: There is an author byline, that should be easy to remove since it has an attribure rel='author'. An example is here: https://www.waz.de/sport/fussball/article241952512/Rettig-weist-Kritik-am-DFB-aus-Politik-zurueck-Kenntnisfrei.html

addie9800, line 51: There are some Lesen Sie auch sections, but they might only be detectable using regex. E.g. https://www.waz.de/staedte/duisburg/article241953128/Patzer-bei-Let-s-Dance-Tony-Bauer-nach-Ausrutscher-ermahnt.html or https://www.waz.de/region/rhein-und-ruhr/article241950806/50-Jahre-VW-Golf-Wenn-aus-Abneigung-Liebe-wird.html
Also this article has an extra paragraph that doesn't really belong to the article: https://www.waz.de/staedte/muelheim/article241946498/Ideen-fuer-Muelheims-Innenstadt-Bis-Ende-2025-gibt-es-Geld.html Perhaps we can filter using `>>`?

addie9800 (changes requested):
Thanks for keeping WAZ updated 👍

src/fundus/publishers/de/waz.py
MaxDall: Hmm, hard to say with `>>`. I don't want to drop actual article content.

MaxDall: I added this.

MaxDall: Done.

addie9800: I don't really see another option except for perhaps the `+++`, but we could also leave it in for now.

addie9800 (dismissed):

src/fundus/publishers/de/waz.py
MaxDall: I added something for `>> ... +++`. Regarding `Lesen sie auch` is think that's too risky for a headline. There could a valid one getting dropped.

addie9800 (approved):

PR #392, opened 2024-03-24
Add Query Parameter to URL & Adding Heise

addie9800:
It works on this link for example: https://www.heise.de/hintergrund/30-Jahre-Beneath-A-Steel-Sky-Big-LINC-is-watching-you-9658837.html?wt_mc=rss.red.ho.ho.rdf.beitrag.beitrag
In Fundus, the actual requested URL will be https://www.heise.de/hintergrund/30-Jahre-Beneath-A-Steel-Sky-Big-LINC-is-watching-you-9658837.html?wt_mc=rss.red.ho.ho.rdf.beitrag.beitrag&seite=all

To have a proper testing case I also had to add Heise as a publisher, I used the code from a former PR #353 , that was marked as draft due to other issues.


src/fundus/publishers/us/__init__.py
```diff
@@ -42,7 +42,14 @@ class US(PublisherEnum):
...
         domain="https://www.theintercept.com/",
-        sources=[RSSFeed("https://theintercept.com/feed/?rss")],
+        sources=[
+            RSSFeed("https://theintercept.com/feed/?lang=en"),
+            Sitemap(
+                "https://theintercept.com/sitemap_index.xml",
+                reverse=True,
+                sitemap_filter=inverse(regex_filter("post-sitemap")),
+            ),
+        ],
```
MaxDall, line 52: That this is displayed as a change here confuses me since it is already part of the master branch. There is no merge commit in this branch's commit history so I'm wondering how these changes got applied.

MaxDall (changes requested):
Thanks for expanding Fundus capabilities. Looks like an awesome feature. Besides some minor things there is one major concern I have to address here. This PR is built on top of the current master branch. There is gonna be a huge change to the core logic soon with #357 necessitating a complete rewrite of this PR. Could you open a new PR based on #357? Afterward, I can review the actual logic. Furthermore, it would be nice to split this PR. One addressing query parameters and a second one depending on the first one introducing `Heise` as a publisher.

src/fundus/publishers/de/__init__.py
```diff
@@ -202,6 +203,18 @@ class DE(PublisherEnum):
         parser=WAZParser,
     )
 
+    Heise = PublisherSpec(
+        name="Heise",
+        domain="https://www.heise.de",
+        sources=[
+            RSSFeed("https://www.heise.de/rss/heise.rdf"),
+            Sitemap("https://www.heise.de/sitemapindex.xml"),
+            NewsMap("https://www.heise.de/news/sitemap.xml"),
```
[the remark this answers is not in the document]
addie9800: Filtering out the bloat from the sitemap actually results in a really, really bad runtime. I cancelled it every time, because it took more than a minute to load a single article. What would you suggest is the better option: just removing the sitemap and relying on the content of the RSS feed or maybe, instead of the index, using just this: https://www.heise.de/sitemap.xml it actually doen't look like bloat to me, but like an unsorted list of the recent articles

src/fundus/publishers/us/__init__.py
addie9800: I think I vaguely remember what happened there. Since I created a local branch some time ago working on this, I might have had an old version of the master branch I started off with. For some reason having it the way it was resulted in a crash (maybe it was the validators issue, I'm not too sure atm) but I cross referenced my current branch with the master branch and that seemed to be causing the problem, so I updated my local branch accordingly

addie9800:
This PR has been split into #403 and #409

PR #398, opened 2024-03-27
Fix domains for several publishers

MaxDall:
Including:
- `TAZ`
- `NTV`
- `APNews`
- `TheIntercept`
- `FreeBeacon`

### DIFF
--- a/src/fundus/publishers/de/__init__.py
+++ b/src/fundus/publishers/de/__init__.py
@@ -158,7 +158,7 @@ class DE(PublisherEnum):
 
     NTV = PublisherSpec(
         name="N-Tv",
-        domain="https://www.ntv.de/",
+        domain="https://www.n-tv.de/",
         sources=[NewsMap("https://www.n-tv.de/news.xml"), Sitemap("https://www.n-tv.de/sitemap.xml")],
         parser=NTVParser,
     )
--- a/src/fundus/publishers/de/__init__.py
+++ b/src/fundus/publishers/de/__init__.py
@@ -176,7 +176,7 @@ class DE(PublisherEnum):
 
     Taz = PublisherSpec(
         name="Die Tageszeitung (taz)",
-        domain="https://www.taz.de/",
+        domain="https://taz.de/",
         sources=[
             NewsMap("https://taz.de/sitemap-google-news.xml"),
             Sitemap("https://taz.de/sitemap-index.xml"),
--- a/docs/supported_publishers.md
+++ b/docs/supported_publishers.md
@@ -238,8 +238,8 @@
         <div>N-Tv</div>
       </td>
       <td>
-        <a href="https://www.ntv.de/">
-          <span>www.ntv.de</span>
+        <a href="https://www.n-tv.de/">
+          <span>www.n-tv.de</span>
         </a>
       </td>
       <td>&#160;</td>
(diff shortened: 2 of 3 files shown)


src/fundus/publishers/de/__init__.py
```diff
@@ -158,7 +158,7 @@ class DE(PublisherEnum):
 
     NTV = PublisherSpec(
         name="N-Tv",
-        domain="https://www.ntv.de/",
+        domain="https://www.n-tv.de/",
```
addie9800, line 161:
```suggestion
        domain="https://n-tv.de/",
```

addie9800 (changes requested):
Was there a specific reason to remove the `www` from just those publishers?

src/fundus/publishers/de/__init__.py
MaxDall: `https://www.n-tv.de/` is the correct domain.

MaxDall:
The severe problem this PR addresses is that we use `urlparse().netloc` to map URLs to publishers within CCNewsCrawling. `.netloc` returns the entire domain with a subdomain, domain, and top-level domain. Let's make an example using `APNews`.

Currently, the "domain" is set as `https://www.apnews.com/`, resulting in `www.apnews.com` as domain (subdomain `www` + second-level domain `apnews` + top-level domain `com`). The correct domain and therefore the one CC-NEWS will crawl and use as `target-uri` is `https://apnews.com/`, with `apnews.com` as the domain. This will result in a mismatch and prevent Fundus from parsing `APNews` within using the CC-NEWS dataset.

There would be a solution using `tldextract` and parsing only the second-level domain + top-level domain, but I think having the subdomain as well prevents Fundus from parsing unwanted stuff.

addie9800:
> The severe problem this PR addresses is that we use `urlparse().netloc` to map URLs to publishers within CCNewsCrawling. `.netloc` returns the entire domain with a subdomain, domain, and top-level domain. Let's make an example using `APNews`.
> 
> Currently, the "domain" is set as `https://www.apnews.com/`, resulting in `www.apnews.com` as domain (subdomain `www` + second-level domain `apnews` + top-level domain `com`). The correct domain and therefore the one CC-NEWS will crawl and use as `target-uri` is `https://apnews.com/`, with `apnews.com` as the domain. This will result in a mismatch and prevent Fundus from parsing `APNews` within using the CC-NEWS dataset.
> 
> There would be a solution using `tldextract` and parsing only the second-level domain + top-level domain, but I think having the subdomain as well prevents Fundus from parsing unwanted stuff.

Oh ok, so the main reason for this change is to have the proper identifier when crawling CC-News. I wasn't aware of that, and was a bit confused about the way you chose to make those changes. I am wondering though, if there is another solution, because the way I understand it, this would require a check for every added publisher to see if the domain used in the PublisherSpec corresponds to the one used within CCNews. And if so, we should also update the documentation accordingly

MaxDall:
It's not about the one used in CC-NEWS, it's the other way around. CC-NEWS uses the correct one, the domains set within our specification aren't correct. Even if `https://www.apnews.com/` gets resolved to the correct site, the actual site's domain is `https://apnews.com/`. You can also look it up in your browser. Just click on the URL.

> this would require a check for every added publisher
That's what I did, but not if it's the one used in CC-NEWS, but if it's the correct one.

addie9800:
I think I understood the issue now. I think a test might be a good idea. Because even though you get rerouted to https://apnews.com, if you navigated there using www.apnews.com, and you copy the link, at least in my browser will still give you https://www.apnews.com. But I think that might be something for another PR

MaxDall:
Hmm. I'm not getting rerouted as well. That makes everything a bit more complicated. Or it doesn't and we should match without subdomain.

PR #403, opened 2024-04-06
Add query parameter [Based on #357]

addie9800:
This PR is a rebrand of and closes #392 where the code is changed to work with #357 
The major change compared to the former version ist that the query_parameter attribute accepts a dictionary of key value pairs. Adding Heise will be done in a separate PR


scripts/generate_parser_test_files.py
```diff
@@ -18,7 +18,7 @@
 
 def get_test_article(enum: PublisherEnum, url: Optional[str] = None) -> Optional[Article]:
     if url is not None:
-        source = WebSource([url], publisher=enum.publisher_name)
+        source = WebSource([url], publisher=enum.publisher_name, query_parameters=enum.query_parameter)
```
MaxDall, line 21: Since this is about crawling a specified URL we should not pass the query parameter here.

src/fundus/scraping/html.py
```diff
@@ -69,12 +69,14 @@ def __init__(
...
         url_filter: Optional[URLFilter] = None,
         request_header: Optional[Dict[str, str]] = None,
+        query_parameters: Optional[Dict[str, str]] = None,
         delay: Optional[Delay] = None,
     ):
         self.url_source = url_source
         self.publisher = publisher
         self.url_filter = url_filter
         self.request_header = request_header or _default_header
+        self.query_parameters = query_parameters
```
MaxDall, line 79: i would suggest

```python
self.query_parameters = query_parameters or {}
```

and remove

```python
if self.query_parameters is not None:
```

MaxDall (changes requested):
@addie9800 Thanks for adding. Looks like a great feature :)

src/fundus/scraping/html.py
addie9800: Sounds good

src/fundus/scraping/html.py
```diff
@@ -107,12 +107,11 @@ def filter_url(u: str) -> bool:
...
-                    for key, value in self.query_parameters.items():
-                        if "?" in url:
-                            url += "&" + key + "=" + value
-                        else:
-                            url += "?" + key + "=" + value
+                for key, value in self.query_parameters.items():
+                    if "?" in url:
+                        url += "&" + key + "=" + value
+                    else:
+                        url += "?" + key + "=" + value
```
MaxDall, line 114: Sorry, one last thing i missed: This should be outside the `try ... catch` block.

MaxDall (changes requested):

MaxDall (approved):

PR #406, opened 2024-04-06
bi authentication bug workaround

addie9800:
After discovering a bug that BI US sometimes prints incomplete articles, the closest @MaxDall and I got to tracing the issue is that there seems to be some kind of error within the authentication of premium users of the US bi website. The incomplete version is a slight modification of the complete version, where the main content is within another div container, which is blocked from view, when using a browser. What triggers whether or not they ask for a premium login could not be determined. So, the workaround is to expand the selectors to cover both cases.

### DIFF
--- a/src/fundus/publishers/us/business_insider.py
+++ b/src/fundus/publishers/us/business_insider.py
@@ -15,12 +15,16 @@ from fundus.parser.utility import (
 
 class BusinessInsiderParser(ParserProxy):
     class V1(BaseParser):
-        _summary_selector = CSSSelector("article ul.summary-list > li")
+        _summary_selector = CSSSelector("article ul[class^='summary-list'] > li")
         _subheadline_selector = CSSSelector("article h2")
         _paragraph_selector = XPath(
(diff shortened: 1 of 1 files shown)


MaxDall (changes requested):
Thanks for giving this a shot 👍 I think the paragraph selector could be simplified by utilizing optional path nodes like `//(.|div[...])`

addie9800:
> Thanks for giving this a shot 👍 I think the paragraph selector could be simplified by utilizing optional path nodes like `//(.|div[...])`

Yeah, I also tried that, but as it turns out that's a XPath 2.0 feature and lxml only supports XPath 1.0. Trying this results in an Exception raised by lxml

MaxDall:
> Yeah, I also tried that, but as it turns out that's a XPath 2.0 feature and lxml only supports XPath 1.0. Trying this results in an Exception raised by lxml

Good to know! I thought I added it once, but i guess I forgot about the part where an exception was thrown.

PR #410, opened 2024-04-15
Adding a swiss publisher (SRF)

addie9800:
I added the Swiss Radio and Televsion ( https://srf.ch ) as a publisher. Note that live tickers cannot be parsed properly due to the content of the articles only being loaded at a later point, so the HTML loaded by Fundus does not contain any content (refer to comment within code for a detailed explanation)


MaxDall (approved):
Looks really great 👍

PR #411, opened 2024-04-15
Attributes Parameter in Test Generation Script

addie9800:
I noticed that the attributes paramater of the `scripts.generate_parser_test_files` script, does not work properly or at least does not work as the others do. I updated this to have the default syntax which requires the user to input the parameters:
`--attributes attr1 att2 attr3` or `-a attr1 attr2 attr3`


MaxDall:
Hey, @addie9800. That's intended, but one may argue that it wasn't a good solution in the first place ;) I support these changes.

MaxDall (approved):

PR #416, opened 2024-04-15
Add `Rheinische Post` as publisher

MaxDall:


addie9800 (approved):
Thanks for adding the publisher 👍

PR #422, opened 2024-04-18
Fix a bug in bf_search regarding boolean values

MaxDall:
There was a nasty bug in `LD.bf_search` basically rendering the parsing of boolean values unusable. This PR fixes this issue by introducing a new sentinel value and fixing some logic bugs.

One problem this could cause is that now the `free_access` attribute finally works as intended and this may cause some trouble since the publishers won't set it honestly.

Update: I opened another PR #423 to address this

### DIFF
--- a/src/fundus/parser/data.py
+++ b/src/fundus/parser/data.py
@@ -18,6 +18,8 @@ from typing_extensions import TypeAlias
 
 LDMappingValue: TypeAlias = Union[List[Dict[str, Any]], Dict[str, Any]]
 
+_sentinel = object()
+
 
 class LinkedDataMapping:
     """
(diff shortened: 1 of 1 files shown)


addie9800 (commented):
Is there a reason to create the `_sentinel` object instead of using `None`?

MaxDall:
> Is there a reason to create the `_sentinel` object instead of using `None`?

The sentinel is unique whereas `None` isn't. One could not differentiate if the value of a key was `None` or not present at all and thus `default` would be returned for all `None` values, besides it only should be returned when the key isn't present.

addie9800 (approved):
Thanks for the explanation, that makes sense. Looks good and thanks for catching this 👍

PR #434, opened 2024-04-20
Reference paper in README.md and add `cite` section

MaxDall:


dobbersc (dismissed):
I think it would be nice to add the table of results as well. Then one can immediately infer the quality of Fundus and compare it to existing libraries.

PR #451, opened 2024-04-23
Add meta-info file automatically to git

MaxDall:
This PR automatically adds meta info files generated for test cases to git.


addie9800 (approved):
Good addition. I did also encounter that every once in a while, but forgot about it.

PR #453, opened 2024-04-24
Added US publisher Rolling Stone

brandjakHU:


src/fundus/publishers/us/rolling_stone.py
```diff
@@ -0,0 +1,44 @@
...
+    extract_article_body_with_selector,
+    generic_author_parsing,
+    generic_date_parsing,
+    generic_topic_parsing,
+)
+
+
+class RollingStoneParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = CSSSelector("div.a-content > * > p.paragraph")
```
MaxDall, line 17: The selector fails to parse pages like this:
https://www.rollingstone.com/music/music-lists/most-anticipated-coachella-sets-2024-1234999086/le-sserafim-1235000035/
https://www.rollingstone.com/music/music-lists/best-canadian-musicians-1234779020/

src/fundus/publishers/us/rolling_stone.py
```diff
@@ -0,0 +1,44 @@
...
+        _summary_selector = CSSSelector("article > header > div.article-excerpt")
+        _subheadline_selector = CSSSelector("article > header > div.article-kicker")
```
MaxDall, line 19: I could not find an example article for the subheadline selector, could you provide one in this comment?

Also, the subheadlines included in this article are not parsed properly:
https://www.rollingstone.com/politics/politics-news/donald-trump-indictments-lawsuits-legal-cases-explained-1234767770/

src/fundus/publishers/us/__init__.py
```diff
@@ -164,3 +165,13 @@ class US(PublisherEnum):
         ],
         parser=BusinessInsiderParser,
     )
+
+    RollingStone = PublisherSpec(
+        name="Rolling Stone",
+        domain="https://www.rollingstone.com/",
+        sources=[
+            NewsMap("https://www.rollingstone.com/news-sitemap.xml"),
+            Sitemap("https://www.rollingstone.com/sitemap_index.xml"),
```
MaxDall, line 174: There is some bloat in the sitemap Fundus won't be able to parse. Could you update the sitemap like this:

```python
Sitemap(
    "https://www.rollingstone.com/sitemap_index.xml",
    sitemap_filter=inverse(lor(regex_filter("post-sitemap"), regex_filter("pmc_list-sitemap"))),
)
```

You can import everything from `fundus.scraping.filter`:

```python
from fundus.scraping.filter import regex_filter, inverse, lor
```

MaxDall (changes requested):
Thanks for adding 👍 Looks really good!

src/fundus/publishers/us/rolling_stone.py
brandjakHU: Thanks, the lists have a slightly different structure than normal articles. The CSSSelector is fixed and works for both now
https://github.com/brandjakHU/fundus/blob/e8918df4fca60325ae08b4cfb56b09559be83efc/src/fundus/publishers/us/rolling_stone.py#L18

MaxDall: Thanks for addressing this. Unfortunately, they still don't work for the articles I posted, further, the new selector selects a lot of empty paragraphs. I would suggest to use `div.a-content p.paragraph` which parses both correctly.

src/fundus/publishers/us/rolling_stone.py
MaxDall: For Fundus, `subheadlines` are headlines that separate paragraphs. In the example article i posted the subheadlines would be `Paying Off a Porn Star`, `Election Interference and Jan. 6`, ...
I would suggest `div.a-content h2.heading, div.a-content h2.c-gallery-vertical-featured-image__title`

brandjakHU: Thanks for clarifying. I added `div.a-content h2.heading`. For the other subheadlines, the class `h2.c-gallery-vertical-featured-image__title` could not be parsed because it is not in the original sourcecode (there, the content is only `h2` without any classes, maybe the class is added with a script before being shown in the browser). Instead, i selected via the identifier that is used by all of these lists with `div.a-content div#pmc-gallery-vertical h2`

src/fundus/publishers/us/rolling_stone.py
brandjakHU: Thanks, i changed it to `div.a-content p.paragraph`. I originally used only `p` because i found one list article which uses `<p>` nodes without the `paragraph`-class (https://www.rollingstone.com/music/music-lists/best-beyonce-songs-1378620/find-your-way-back-2019-1380991/). However, all other articles i looked at use `class=paragraph`, so maybe it is better to select like this and avoid empty paragraphs?

MaxDall (approved):

PR #454, opened 2024-04-24
Use `!r` instead of single quotes for strings in f-strings

MaxDall:
Additionally this re formats some logging messages.

closes #276

### DIFF
--- a/src/fundus/publishers/base_objects.py
+++ b/src/fundus/publishers/base_objects.py
@@ -59,8 +59,8 @@ class PublisherEnum(Enum, metaclass=PublisherEnumMeta):
         for url_source in spec.sources:
             if not isinstance(url_source, URLSource):
                 raise TypeError(
-                    f"Unexpected type '{type(url_source).__name__}' as source for {self.name}. "
-                    f"Allowed are '{', '.join(cls.__name__ for cls in iterate_all_subclasses(URLSource))}'"
+                    f"Unexpected type {type(url_source).__name__!r} as source for {self!r}. "
+                    f"Allowed are {', '.join(repr(cls.__name__) for cls in iterate_all_subclasses(URLSource))}"
                 )
             source_mapping[type(url_source)].append(url_source)
 
@@ -75,8 +75,8 @@ class PublisherEnum(Enum, metaclass=PublisherEnumMeta):
         for source_type in source_types:
             if not inspect.isclass(source_type) or not issubclass(source_type, URLSource):
                 raise TypeError(
-                    f"Got unexpected type '{source_type}'. "
-                    f"Allowed are '{', '.join(cls.__name__ for cls in iterate_all_subclasses(URLSource))}'"
+                    f"Got unexpected type {source_type!r}. "
+                    f"Allowed are {', '.join(repr(cls.__name__) for cls in iterate_all_subclasses(URLSource))}"
                 )
         return all(bool(self.source_mapping.get(source_type)) for source_type in source_types)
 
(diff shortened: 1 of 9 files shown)


src/fundus/publishers/base_objects.py
```diff
@@ -59,8 +59,8 @@ def __init__(self, spec: PublisherSpec):
         for url_source in spec.sources:
             if not isinstance(url_source, URLSource):
                 raise TypeError(
-                    f"Unexpected type '{type(url_source).__name__}' as source for {self.name}. "
-                    f"Allowed are '{', '.join(cls.__name__ for cls in iterate_all_subclasses(URLSource))}'"
+                    f"Unexpected type {type(url_source).__name__!r} as source for {self!r}. "
+                    f"Allowed are {', '.join(repr(cls.__name__) for cls in iterate_all_subclasses(URLSource))}"
```
addie9800, line 63: If there is no special reason to do it this way, I would suggest, for consistencies sake, to do it this way: `', '.join(cls.__name__ for ... )!r`

src/fundus/publishers/base_objects.py
```diff
@@ -75,8 +75,8 @@ def supports(self, source_types: List[Type[URLSource]]) -> bool:
         for source_type in source_types:
             if not inspect.isclass(source_type) or not issubclass(source_type, URLSource):
                 raise TypeError(
-                    f"Got unexpected type '{source_type}'. "
-                    f"Allowed are '{', '.join(cls.__name__ for cls in iterate_all_subclasses(URLSource))}'"
+                    f"Got unexpected type {source_type!r}. "
+                    f"Allowed are {', '.join(repr(cls.__name__) for cls in iterate_all_subclasses(URLSource))}"
```
addie9800, line 79: If there is no special reason to do it this way, I would suggest, for consistencies sake, to do it this way: `', '.join(cls.__name__ for ... )!r`

src/fundus/scraping/scraper.py
```diff
@@ -36,28 +36,28 @@ def scrape(
...
-                        err.args = (str(err) + "\n\n" + error_message,)
-                        raise err
+                        error.args = (str(error) + "\n\n" + error_message,)
+                        raise error
                     elif error_handling == "catch":
-                        yield Article(html=html, exception=err)
+                        yield Article(html=html, exception=error)
                     elif error_handling == "suppress":
-                        logger.info(f"Skipped article at '{html.requested_url}' because of: {err!r}")
+                        logger.info(f"Skipped article at {html.requested_url!r} because of: {error}")
```
addie9800, line 48:
```suggestion
                        logger.info(f"Skipped article at {html.requested_url!r} because of: {error!r}")
```

addie9800 (changes requested):
I think there are still some leftover cases, where the string representation has not been changed yet, namely: data.py:55, (not sure if needed but) article.py:86-90, utility.py:199, mdr.py:25, dw.py:26, the_gateway_pundit.py:19, base_object.py:74

src/fundus/publishers/base_objects.py
MaxDall: Those two produce different outcomes and the current one is the desired.

MaxDall:
> I think there are still some leftover cases, where the string representation has not been changed yet, namely: data.py:55, (not sure if needed but) article.py:86-90, utility.py:199, mdr.py:25, dw.py:26, the_gateway_pundit.py:19, base_object.py:74

Hey, @addie9800. Thanks for the detailed review 👍 I checked every occurrence of `'{` in the codebase before I commited and all remaining occurrences are either not a string or on purpose.

addie9800 (approved):
👍

PR #457, opened 2024-04-25
Sportschau added

JuSteif:
Sportschau is a german sports magazine


src/fundus/publishers/de/__init__.py
```diff
@@ -30,6 +31,17 @@
...
 class DE(PublisherEnum):
+    SportSchau = PublisherSpec(
+        name="Sportschau",
+        domain="https://www.sportschau.de/",
+        parser=SportSchauParser,
+        sources=[
+            RSSFeed("https://www.sportschau.de/index~rss2.xml"),
+            Sitemap("https://www.sportschau.de/index~sitemap_p-0.xml"),
+            NewsMap("https://www.sportschau.de/kompakt-sp-100~news.xml"),
+        ],
```
addie9800, line 42: For some reason sportschau seems to also link articles from other websites such as mdr, which then naturally can't be parsed by your Parser. I would suggest adding `url_filter=inverse(regex_filter("sportschau.de"))` to your PublisherSpec. Don't worry if it doesn't seem to make a difference yet, this is due to a bug that is fixed with #459

src/fundus/publishers/de/sportschau.py
```diff
@@ -0,0 +1,43 @@
...
+from fundus.parser.utility import (
+    extract_article_body_with_selector,
+    generic_author_parsing,
+    generic_date_parsing,
+)
+
+
+class SportSchauParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = CSSSelector("article >p.textabsatz")
```
addie9800, line 16: As of now, this selector also include the summary. I would suggest extending this selector by `article >p.textabsatz:not(:nth-of-type(1))`

addie9800 (changes requested):
Thanks a lot for adding this 👍

src/fundus/publishers/de/__init__.py
JuSteif: sorry, what does "inverse" do, because it seems like it does not work

addie9800: Inverse works like a logical not for filters. I did forget to mention you still have to import it though. Sorry about that:

`from fundus.scraping.filter import inverse, regex_filter`

addie9800: The PR looks good, I just forgot to mention that you have to regenerate the test files if you update the extraction logic: `scripts.generate_parser_test_files -p SportSchau -oj`

MaxDall (dismissed):

addie9800 (approved):

PR #458, opened 2024-04-25
Add PublisherSpec for "Der Postillon"

buschd-nlp:
As described in Task 3 of the current exercise sheet.


src/fundus/publishers/de/postillon.py
```diff
@@ -0,0 +1,41 @@
...
+    extract_article_body_with_selector,
+    generic_author_parsing,
+    generic_date_parsing,
+    generic_topic_parsing,
+)
+
+
+class PostillonParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector  = CSSSelector("div[id=post-body] > p")
```
addie9800, line 17: The Postillon includes blockqutos sometimes, which aren't parsed atm. e.g. https://www.der-postillon.com/2024/04/alltagsfrage-baden-verboten.html
There are two options to fix this as far as I see: `div[id=post-body] p` or `div[id=post-body] > p, div[id=post-body] > blockquote > p`

addie9800 (changes requested):
Great idea adding the Postillon to the repo :)

src/fundus/publishers/de/postillon.py
buschd-nlp: `div[id=post-body] p` seems morally correct here, pushed.

addie9800 (approved):

addie9800 (dismissed):

PR #460, opened 2024-04-25
RBB24 added

henrikkirchmann:
Added https://www.rbb24.de/ news from Berlin and Brandenburg


src/fundus/publishers/de/rbb24.py
```diff
@@ -0,0 +1,95 @@
...
+from fundus.parser.utility import (
+    extract_article_body_with_selector,
+    generic_author_parsing,
+    generic_date_parsing,
+)
+
+
+class RBB24Parser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = XPath("//div[contains(concat(' ', @class , ' '), ' textblock ')]")
```
addie9800, line 17: This way, the text gets extracted as a full block. If you change it to `//div[contains(concat(' ', @class , ' '), ' textblock ')]/p` the amoung of extracted text stays the same, but it's separated into paragraphs. Btw, this might be a simpler solution, which has the same result: `//div[@class='textblock']/p`

src/fundus/publishers/de/rbb24.py
```diff
@@ -0,0 +1,95 @@
...
+        _subheadline_selector = XPath("//h4[contains(concat(' ', @class , ' '), ' texttitle ')]")
+        _author_selector = CSSSelector("span.authorname")
+        _date_selector = CSSSelector("div.lineinfo")
+
+        @attribute
+        def body(self) -> ArticleBody:
+            article_body = extract_article_body_with_selector(
+                self.precomputed.doc,
+                summary_selector=XPath("//div[contains(concat(' ', @class , ' '), ' shorttext ')]/p"),
```
addie9800, line 26: For more clarity, could you move this selector up to where you define the other variables as well?

addie9800 (changes requested):
Thanks for adding this. Also really nice solution dealing with the author problem. 👍

addie9800 (approved):
This looks good 👍

PR #464, opened 2024-04-26
Add support for newspaper "Kicker"

Benjamin2107:
This PR adds support for the german newspaper "Kicker".


src/fundus/publishers/de/__init__.py
```diff
@@ -240,3 +242,14 @@ class DE(PublisherEnum):
         ],
         parser=RheinischePostParser,
     )
+
+    Kicker = PublisherSpec(
+        name="Kicker",
+        domain="https://www.kicker.de/",
+        sources=[
+            RSSFeed("https://newsfeed.kicker.de/news/aktuell"),
+            NewsMap("https://leserservice.kicker.de/sitemap_0.xml"),
```
addie9800, line 251: From what I can see this would would be considered a `Sitemap`. The NewsMap for kicker can be found here: https://newsfeed.kicker.de/googlesitemapnews.xml
I would also consider adding a sitemap_filter (as a parameter when creating the `Sitemap` obejct) for `leserservice.kicker.de`: `sitemap_filter=regex_filter('leserservice.kicker.de')`

addie9800 (changes requested):
This looks really good, thanks for adding kicker 👍

src/fundus/publishers/de/__init__.py
Benjamin2107: Oh, at first I also had it as a sitemap. I'm going to change it.

addie9800 (dismissed):
Very nice, thanks 👍

PR #468, opened 2024-04-26
Add new publisher "Frankfurter Rundschau"

SunitaSi:
I have added the newspaper "Frankfurter Rundschau" (https://www.fr.de)


MaxDall (dismissed):
Thanks for adding :) This looks really good 👍

PR #475, opened 2024-04-28
Addition of parser for the german newspaper "Gamestar"

Opticus10:


src/fundus/publishers/de/gamestar.py
```diff
@@ -0,0 +1,43 @@
...
+        def publishing_date(self) -> Optional[datetime]:
+            return generic_date_parsing(self.precomputed.ld.bf_search("datePublished"))
+
+        @attribute
+        def topics(self) -> List[str]:
+            keywords = self.precomputed.meta.get("keywords")
+            if not keywords:
+                return [""]
+            else:
+                return keywords.strip(" ").split(",")
```
MaxDall, line 43: There is a utility function called `generic_topic_parsing` to parse topics which you could import from `fundus.parser.utility`.

src/fundus/publishers/de/gamestar.py
```diff
@@ -0,0 +1,43 @@
...
+from fundus.parser.utility import (
+    extract_article_body_with_selector,
+    generic_author_parsing,
+    generic_date_parsing,
+)
+
+
+class GamestarParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = CSSSelector("div.article-content > p")
```
MaxDall, line 16: Judging by [this](https://www.gamestar.de/artikel/neu-bei-steam,3412554.html) article, there is a lot of bloat, `captions` etc., extracted alongside the actual content. Further, this newspaper seems to include `subheadlines` (headlines between paragraphs) and a `summary` (little intro text). You can example for both [here](https://www.gamestar.de/artikel/spielerin-baut-skyrim-setup-und-spuert-schmerzen,3412551.html). Could you add those?

MaxDall (changes requested):
Thanks for adding 👍

MaxDall (dismissed):

MaxDall (dismissed):

PR #478, opened 2024-04-28
Added Euronews to fundus

dkm1006:


src/fundus/publishers/fr/__init__.py
```diff
@@ -13,3 +15,17 @@ class FR(PublisherEnum):
...
+        domain="https://www.euronews.com/",
+        sources=[
+            Sitemap("https://www.euronews.com/sitemaps/en/articles.xml"),
+            Sitemap("https://de.euronews.com/sitemaps/de/articles.xml"),
+            Sitemap("https://fr.euronews.com/sitemaps/fr/articles.xml"),
+            NewsMap("https://www.euronews.com/sitemaps/en/latest-news.xml"),
+            NewsMap("https://de.euronews.com/sitemaps/de/latest-news.xml"),
+            NewsMap("https://fr.euronews.com/sitemaps/fr/latest-news.xml"),
+        ],
+        parser=EuronewsParser,
```
MaxDall, line 30: Since this was added to `fr` could you rename the publisher to `EuronewsFR` and add only sources with `fr` subdomain?

If the page layout stays the same, it would be nice if you could add the other languages to their corresponding regions and reuse the parser. You wouldn't have to create three separate files, just declare the parser in one region and import it in the others. If not, don't worry, adding just this one is enough ;)

src/fundus/publishers/fr/euronews.py
```diff
@@ -0,0 +1,42 @@
...
+    class V1(BaseParser):
+        @attribute
+        def title(self) -> Optional[str]:
+            return self.precomputed.meta.get("og:title")
+
+        @attribute
+        def body(self) -> ArticleBody:
+            summary_selector = CSSSelector("p.c-article-summary")
+            subheadline_selector = CSSSelector("div.c-article-content > h2")
+            paragraph_selector = CSSSelector("div.c-article-content > p")
```
MaxDall, line 19: Those should be added as class variables of the parser so they won't be compiled every time `body` is called.

Please refer to another parser to see how to add them as class variables.

MaxDall (changes requested):
Thanks for adding 👍 looks really good!

src/fundus/publishers/fr/euronews.py
dkm1006: Sure, I’ll do it tonight.

src/fundus/publishers/fr/__init__.py
dkm1006: For fr, en and de the layout is the same, the other languages I haven’t tried, yet. (I expect the same though). 
I’m not sure I’ll have time to add all the other languages that they offer but I’ll split fr, en and de tonight.

MaxDall: @dkm1006 That looks really great! Thanks a lot :) 

Good idea. I added a `shared` directory to the publisher dir.

MaxDall (approved):

addie9800 (approved):

PR #479, opened 2024-04-28
Add new publisher Haberturk

ozelalisen:
First publisher from Turkey :)


src/fundus/publishers/tr/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+
+from .haberturk import HaberturkParser
+
+
+class TR(PublisherEnum):
+    Haberturk = PublisherSpec(
+        name="Haberturk",
+        domain="https://www.haberturk.com/",
+        sources=[
+            Sitemap("https://www.haberturk.com/sitemap.xml"),
```
MaxDall, line 12: There seems to be some stuff indexed in the sitemaps not referring to articles, like `sitemap_headlines`. Could make use of the `sitemap_filter` parameter of `Sitemap` to access only sitemaps referring articles. You can import everything you need from `fundus.scraping.filter`. You can find examples of how to use those in `fundus.publishers.us.__init__.py`. For further information read [this tutorial](https://github.com/flairNLP/fundus/blob/master/docs/4_how_to_filter_articles.md#url-filter).

src/fundus/publishers/tr/haberturk.py
```diff
@@ -0,0 +1,41 @@
...
+    extract_article_body_with_selector,
+    generic_author_parsing,
+    generic_date_parsing,
+    generic_topic_parsing,
+)
+
+
+class HaberturkParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = CSSSelector("article[property='articleBody']")
```
MaxDall, line 17: Fundus separates articles into paragraphs and subheadlines introduced with a short summary. I don't speak Turkish, but from the layout it looks like this publisher supports all of them. Could you add the missing selectors and extract the content paragraph wise and not as s single block.

MaxDall (changes requested):
Thanks for adding our first Turkish publisher 🚀

src/fundus/publishers/tr/__init__.py
ozelalisen: Regex filter added

src/fundus/publishers/tr/haberturk.py
ozelalisen: I have chosen it as a subheading, I could also adopt it as a summary. Do you think it would be more suitable for a summary?

MaxDall: @ozelalisen Yes, that part pretty much summarizes the rest of the article ;)

ozelalisen: Alright, that is fixed

MaxDall (dismissed):

PR #486, opened 2024-04-28
Add netzpolitik.org publisher

SebChmie:


src/fundus/publishers/de/netzpolitik_org.py
```diff
@@ -0,0 +1,50 @@
...
+    generic_date_parsing,
+    generic_topic_parsing,
+    parse_title_from_root,
+    strip_nodes_to_text,
+)
+
+
+class NetzpolitikOrgParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = CSSSelector("div.entry-content > p")
```
MaxDall, line 18: Judging by [this article](https://netzpolitik.org/2024/mastodon-und-peertube-eu-macht-fediverse-dicht-weil-niemand-die-server-betreiben-will/) the selector seems to have problems extracting text wrapped in `blockquote` tags.

src/fundus/publishers/de/__init__.py
```diff
@@ -30,6 +31,13 @@
 
 # noinspection PyPep8Naming
 class DE(PublisherEnum):
+    NetzpolitikOrg = PublisherSpec(
+        name="netzpolitik.org",
+        domain="https://netzpolitik.org/",
+        sources=[Sitemap("https://netzpolitik.org/sitemap.xml"), RSSFeed("https://netzpolitik.org/feed/")],
```
MaxDall, line 37: Could you add a filter to the sitemap by utilizing the `sitemap_filter` parameter of `Sitemap` to request only pages of type `sitemap_posttype_post`. You can do this by utilizing `regex_filter` and `inverse` from the `fundus.scraping.filter` module.

MaxDall (changes requested):
Thanks for adding 👍 This one looks really good and thank you that you made the effort to understand how Fundus works. I can tell from your PR :)

src/fundus/publishers/de/__init__.py
SebChmie: Did you mean `sitemap-posttype-post` instead of `sitemap_posttype_post`?

MaxDall (approved):

addie9800 (approved):

PR #487, opened 2024-04-29
Added winfuture publisher

lindhork:


src/fundus/publishers/de/winfuture.py
```diff
@@ -0,0 +1,46 @@
...
+
+        @attribute
+        def body(self) -> ArticleBody:
+            return extract_article_body_with_selector(
+                doc=self.precomputed.doc,
+                paragraph_selector=self._paragraph_selector,
+                summary_selector=self._summary_selector,
+                subheadline_selector=self._subheadline_selector,
+            )
+            # return ArticleBody("TEST",[])
```
MaxDall, line 34:
```suggestion
```

src/fundus/publishers/de/winfuture.py
```diff
@@ -0,0 +1,46 @@
...
+            return self.precomputed.meta.get("og:title")
+
+        @attribute
+        def body(self) -> ArticleBody:
+            return extract_article_body_with_selector(
+                doc=self.precomputed.doc,
+                paragraph_selector=self._paragraph_selector,
+                summary_selector=self._summary_selector,
+                subheadline_selector=self._subheadline_selector,
+            )
```
MaxDall, line 33: Unfortunately, the layout used by this newspaper is so uncommon, that you can't rely on this utility function. The article isn't extracted as paragraphs and the layout isn't in order.

MaxDall (changes requested):
Unfortunately the newspaper you choose has such an unconventional layout that i would say it is not worth the effort try making it compatible with Fundus.

MaxDall:
@addie9800 That's impressive 👍 Thanks a lot :)

MaxDall (approved):

PR #492, opened 2024-04-29
added french publisher le figaro

mornningstar:


src/fundus/publishers/fr/__init__.py
```diff
@@ -13,3 +14,14 @@ class FR(PublisherEnum):
         ],
         parser=LeMondeParser,
     )
+
+    LeFigaro = PublisherSpec(
+        name="Le Figaro",
+        domain="https://www.lefigaro.fr/",
+        sources=[
+            Sitemap("https://sitemaps.lefigaro.fr/lefigaro.fr/articles.xml"),
+            Sitemap("https://www.lefigaro.fr/elections/resultats/sitemap.xml"),
```
MaxDall, line 23: You can remove this `https://www.lefigaro.fr/elections/resultats/sitemap.xml` sitemap.

src/fundus/publishers/fr/le_figaro.py
```diff
@@ -0,0 +1,37 @@
...
+    extract_article_body_with_selector,
+    generic_author_parsing,
+    generic_date_parsing,
+)
+
+
+class LeFigaroParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector: CSSSelector = CSSSelector("p[class='fig-paragraph']")
+        _author_selector: CSSSelector = CSSSelector("a[rel='author']")
```
MaxDall, line 18: `_author_selector` looks like a leftover.

Further, the publisher uses a little summary to introduce their articles as well as subheadlines between sections of paragraphs. Could you add those? Here is an [example article](https://etudiant.lefigaro.fr/article/etudes/a-strasbourg-rennes-menton-saint-etienne-des-etudiants-bloquent-les-iep-20240430/) where you can find both.

After adding, you have to run `python -m scripts.generate_parser_test_files -p LeFigaro -oj`

MaxDall (changes requested):
Thanks for adding 👍

src/fundus/publishers/fr/le_figaro.py
mornningstar: I will add that!
What do you mean by "leftover"? This is how I was able to extract the authors name..

MaxDall (approved):

PR #495, opened 2024-04-29
add new publisher the standards

Merchants-11:


src/fundus/publishers/uk/the_standard.py
```diff
@@ -0,0 +1,40 @@
...
+    generic_date_parsing,
+    generic_topic_parsing,
+)
+
+
+class TheStandardParser(fundus.parser.ParserProxy):
+    class V1(fundus.parser.BaseParser):
+        _paragraph_selector = CSSSelector("div[data-element*=story-body] > p")
+
+        def body(self) -> ArticleBody:
```
MaxDall, line 20: You forgot to register this as `attribute` using the attribute decorator

src/fundus/publishers/uk/__init__.py
```diff
@@ -52,3 +53,13 @@ class UK(PublisherEnum):
         ],
         parser=INewsParser,
     )
+
+    TheStandard = PublisherSpec(
```
MaxDall, line 57: A more suitable class name would be `Standard`

MaxDall (changes requested):
Thanks for adding 👍 but since you disabled the `Body` attribute this one is pretty much not usable.

MaxDall (approved):

PR #497, opened 2024-04-30
Add MalayMail publisher

syafrahman:


src/fundus/publishers/my/malay_mail.py
```diff
@@ -0,0 +1,35 @@
...
+from fundus.parser import ArticleBody, BaseParser, ParserProxy, attribute
+from fundus.parser.utility import (
+    extract_article_body_with_selector,
+    generic_author_parsing,
+    generic_date_parsing,
+)
+
+
+class MalayMailParser(ParserProxy):
+    class V1(BaseParser):
```
addie9800, line 15: It would be great, if you could also add a function to extract keywords. This can be done very easily by using the keywords tag. MalayMail seems to support it within it's meta data and ld JSON

src/fundus/publishers/my/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+from fundus.scraping.url import RSSFeed, Sitemap
+
+
+class MY(PublisherEnum):
+    MalayMail = PublisherSpec(
+        name="Malay Mail",
+        domain="https://www.malaymail.com/",
+        sources=[
+            Sitemap("https://www.malaymail.com/sitemap.xml"),
+            RSSFeed("https://www.malaymail.com/feed/rss/malaysia"),
```
addie9800, line 12:
```suggestion
            RSSFeed("https://www.malaymail.com/feed/rss/"),
```
This seems to be a broader RSSFeed. While it still seems to have the same articles as the /malaysia/ feed, it also has a couple extra

src/fundus/publishers/my/malay_mail.py
```diff
@@ -0,0 +1,35 @@
...
+        _paragraph_selector = CSSSelector("div.article-body > p")
```
addie9800, line 16: Longer articles also contain subheadlines. For example this one: https://www.malaymail.com/news/world/2024/05/02/bogota-cuts-ties-with-israel-over-genocidal-gaza-campaign/131974 
Could you also add a subheadline_selector that covers this case?

addie9800 (changes requested):
This looks really good. Thanks for adding our first Malaysian publisher 👍

addie9800 (approved):

PR #500, opened 2024-04-30
Require at least one paragraph to evaluate `ArticleBody` to true

MaxDall:
This PR requires `ArticleSection` and `ArticleBody` respectively to consist of at least one `paragraph` to be evaluated as `True`. This is to increase the data quality of retrieved articles.

I had to replace an old `WAZ` test case which consisted of only a summary.


dobbersc (approved):
Thanks for the fix!

PR #504, opened 2024-05-02
Clean URLs before inserting into response_cache

addie9800:
Depending on the Source of the URL (RSSFeed or Sitemap) Publishers tend to sometimes add a parameter to the url indicating the origin. E.g. Kicker appends #rssom. If max_articles is set to a large enough value, this may lead to the same article being crawled twice, since the response_cache distinguishes the two URLs

### DIFF
--- a/src/fundus/scraping/crawler.py
+++ b/src/fundus/scraping/crawler.py
@@ -26,6 +26,7 @@ from typing import (
     Union,
     cast,
 )
+from urllib.parse import urljoin, urlparse
 
 import dill
 import more_itertools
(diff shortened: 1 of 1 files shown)


src/fundus/scraping/crawler.py
```diff
@@ -212,8 +213,11 @@ def build_extraction_filter() -> Optional[ExtractionFilter]:
         for article in self._build_article_iterator(
             tuple(fitting_publishers), error_handling, build_extraction_filter(), url_filter
         ):
-            if not only_unique or article.html.responded_url not in response_cache:
-                response_cache.add(article.html.responded_url)
+            cached_url = article.html.responded_url
+            if any(parameter_indicator in cached_url for parameter_indicator in ("?", "#")):
+                cached_url = urljoin(article.html.responded_url, urlparse(article.html.responded_url).path)
```
MaxDall, line 218: For readability, I would suggest making this a function `remove_query_parameter` and then rename `cached_url` in something alongside `url_without_query_parameter`

MaxDall (changes requested):
Really good idea!

MaxDall (approved):

PR #505, opened 2024-05-02
Change deserialisation function for dates

addie9800:
This PR changes the deserialisation function from the datetime library to the dateutils library, because the datetime library cannot currently correctly parse dates ending with 'Z', which comminly used to denote a timezone difference in JS dates


MaxDall:
Hmm, I'm sure I'm missing something here, but how should this ever be a problem? We deserialize a `datetime` serialization. Maybe there is something else happening here?

PR #506, opened 2024-05-02
Rearrange If Statement

addie9800:
This is just a minor change, but it since the overwrite flag is verified before the JSON data is loaded, it allows us to overwrite the JSON, even if the JSON is malformed for whatever reason. With the former version, the JSON would first be loaded and cause an Exception if it was malformed. First seen in #466


MaxDall (approved):
Thanks, that's a really good idea 👍

PR #508, opened 2024-05-06
Add timeout to publisher_coverage.py

MaxDall:
This adds a new timeout wrapper to publisher coverage enabling Fundus to use `Sitemap`s for coverage as well.

### DIFF
--- a/scripts/utility.py
+++ b/scripts/utility.py
@@ -1,13 +1,9 @@
+import _thread as thread
 import threading
 from functools import wraps
 from typing import Callable, Optional, TypeVar, overload
 
 from typing_extensions import ParamSpec
-
-try:
-    import thread
-except ImportError:
-    import _thread as thread
 
 P = ParamSpec("P")
 T = TypeVar("T")
--- a/scripts/utility.py
+++ b/scripts/utility.py
@@ -39,7 +39,6 @@ def timeout(func: Callable[P, T], time: float, silent: bool = False) -> Callable
                 raise TimeoutError from err
         finally:
             timer.cancel()
-            timer.join()
         return result
 
     return wrapper
--- a/scripts/utility.py
+++ b/scripts/utility.py
@@ -14,23 +14,23 @@ def _interrupt_handler() -> None:
 
 
 @overload
-def timeout(func: Callable[P, T], time: int, silent: bool = False) -> Callable[P, T]:
+def timeout(func: Callable[P, T], time: float, silent: bool = False) -> Callable[P, T]:
     ...
 
 
 @overload
-def timeout(func: Callable[P, T], time: int, silent: bool = True) -> Callable[P, Optional[T]]:
+def timeout(func: Callable[P, T], time: float, silent: bool = True) -> Callable[P, Optional[T]]:
     ...
 
 
-def timeout(func: Callable[P, T], time: int, silent: bool = False) -> Callable[P, Optional[T]]:
+def timeout(func: Callable[P, T], time: float, silent: bool = False) -> Callable[P, Optional[T]]:
     @wraps(func)
     def wrapper(*args: P.args, **kwargs: P.kwargs) -> Optional[T]:
         # register interrupt handler
         timer = threading.Timer(time, _interrupt_handler)
-        timer.start()
 
         try:
+            timer.start()
             result = func(*args, **kwargs)
         except KeyboardInterrupt as err:
             if silent:
--- a/scripts/utility.py
+++ b/scripts/utility.py
@@ -14,16 +14,16 @@ def _interrupt_handler() -> None:
 
 
 @overload
-def timeout(func: Callable[P, T], time: int, silent: Literal[False] = ...) -> Callable[P, T]:
+def timeout(func: Callable[P, T], time: float, silent: Literal[False] = ...) -> Callable[P, T]:
     ...
 
 
 @overload
-def timeout(func: Callable[P, T], time: int, silent: Literal[True]) -> Callable[P, Optional[T]]:
+def timeout(func: Callable[P, T], time: float, silent: Literal[True]) -> Callable[P, Optional[T]]:
     ...
 
 
-def timeout(func: Callable[P, T], time: int, silent: bool = False) -> Callable[P, Optional[T]]:
+def timeout(func: Callable[P, T], time: float, silent: bool = False) -> Callable[P, Optional[T]]:
     @wraps(func)
     def wrapper(*args: P.args, **kwargs: P.kwargs) -> Optional[T]:
         # register interrupt handler
--- a/scripts/utility.py
+++ b/scripts/utility.py
@@ -1,7 +1,7 @@
 import _thread as thread
 import threading
 from functools import wraps
-from typing import Callable, Optional, TypeVar, overload
+from typing import Callable, Literal, Optional, TypeVar, overload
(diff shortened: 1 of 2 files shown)


scripts/utility.py
```diff
@@ -0,0 +1,49 @@
+import threading
+from functools import wraps
+from typing import Callable, Optional, TypeVar, overload
+
+from typing_extensions import ParamSpec
+
+try:
+    import thread
+except ImportError:
+    import _thread as thread
```
dobbersc, line 10: When does the import of `thread` fail? I could only find the documentation for [`_thread`](https://docs.python.org/3/library/_thread.html#module-_thread) in the Python3 docs. Is `thread` just a Python2 thing, since it is available [here](https://python.readthedocs.io/en/v2.7.2/library/thread.html)? So just using `import _thread` would be enough.

Also weird that the module is protected, properly because it is a low-level API. But it should be fine since it's just for the publisher coverage and provides an easy way to interrupt the main thread.

Alternatively, we could avoid using `_thead` entirely by passing a mutable argument to the interpreter function. But I'm not sure if this is really more elegant. (Example [here](https://stackoverflow.com/a/25189669))

scripts/utility.py
```diff
@@ -0,0 +1,49 @@
...
+        try:
+            result = func(*args, **kwargs)
+        except KeyboardInterrupt as err:
+            if silent:
+                return None
+            else:
+                raise TimeoutError from err
+        finally:
+            timer.cancel()
+            timer.join()
```
dobbersc, line 46: Just to clarify: The timer needs to get cancelled if the `KeyboardInterrupt` is actually thrown by user action and not by the `_interrupt_handler`. Thus the timer may not have been finished.

Also, what is the purpose of `.join`? Shouldn't `.cancel` do everything that's needed as described in the documentation?

dobbersc (changes requested):

scripts/utility.py
MaxDall: The timer is canceled when a result comes in or a `KeybordInterupt` is raised, either by user action or by `thread.interrupt_main()`.

The join is just for safety reasons. I couldn't find a reference in the documentation that the thread is joined when using calling `cancel`. Could you give me a link?

scripts/utility.py
MaxDall: Good point, I'm gonna change that import according to your comment.

I would avoid a solution using a mutable argument, but interesting idea anyways! Thanks for the insight.

scripts/utility.py
```diff
@@ -0,0 +1,49 @@
...
+def timeout(func: Callable[P, T], time: int, silent: bool = False) -> Callable[P, T]:
+    ...
+
+
+@overload
+def timeout(func: Callable[P, T], time: int, silent: bool = True) -> Callable[P, Optional[T]]:
+    ...
+
+
+def timeout(func: Callable[P, T], time: int, silent: bool = False) -> Callable[P, Optional[T]]:
```
dobbersc, line 26: I don't think the overload type hints work as intended. Example:

```py
def test(a: int) -> int:
    return a + 1
    
reveal_type(timeout(test, time=20))  # Revealed type is "def (a: builtins.int) -> builtins.int"
reveal_type(timeout(test, time=20, silent=True))  # Revealed type is "def (a: builtins.int) -> builtins.int"
reveal_type(timeout(test, time=20, silent=False))  # Revealed type is "def (a: builtins.int) -> builtins.int"
```

I would expect the second variant to yield `Optional[int]` as return type.

```suggestion
@overload
def timeout(func: Callable[P, T], time: int, silent: Literal[False] = ...) -> Callable[P, T]:
    ...


@overload
def timeout(func: Callable[P, T], time: int, silent: Literal[True]) -> Callable[P, Optional[T]]:
    ...


def timeout(func: Callable[P, T], time: int, silent: bool = False) -> Callable[P, Optional[T]]:
```

The overloaded types of `silent` directly match the literal type of the parameter while the ellipsis signifies the optional presence of the parameter. (This is also explained [here](https://mypy.readthedocs.io/en/latest/more_types.html#function-overloading) in the mypy documentation.)

Now we get:

```
main.py:52: note: Revealed type is "def (a: builtins.int) -> builtins.int"
main.py:53: note: Revealed type is "def (a: builtins.int) -> Union[builtins.int, None]"
main.py:54: note: Revealed type is "def (a: builtins.int) -> builtins.int"
```

scripts/utility.py
dobbersc: Yes, I agree with you.

scripts/utility.py
dobbersc: The only documentation I could find was [this](https://docs.python.org/3/library/threading.html#threading.Timer.cancel).

> Stop the timer, and cancel the execution of the timer’s action. This will only work if the timer is still in its waiting stage.

I don't know if it will be joined silently or not. I just assumed that this method takes care of everything, since stopping the timer is all we require "directly". [This](https://github.com/python/cpython/blob/b6839942a8906fccdd64e749abeefe8a61ce7e03/Lib/threading.py#L1333) is the line in the cpython source code but with my knowledge it does not help.

scripts/utility.py
MaxDall: Really nice!!

scripts/utility.py
MaxDall: I will remove the `join`

scripts/utility.py
MaxDall: But why the `Literal`? Shouldn't `silent: bool = True/False` be correct? According to the documentation? What am I missing here because it obviously isn't 😅

dobbersc: In my opinion, the mypy documentation does not touch on this (or I missed it). As I understand it, mypy only checks the type itself and the absence or presence:

> The default values of a function’s arguments don’t affect its signature – only the absence or presence of a default value does.

So to tell mypy that we return `Callable[P, T]` for `slient=False` we need to restrict the type itself to only `False` via `Literal[False]`, as `bool` would still allow for both values `True` and `False`. When using `silent: bool = True/False` mypy would not accept the variant where the caller does not provide a `silent` parameter, and the default is used. This case is covered by the ellipsis, which is then annotated with `Literal[True]` since the default value is `True`. 

Example with ellipsis:
```
# Applies when "flag" is omitted or set to True.
@overload
def func(a: int, flag: Literal[True] = ...) -> int: ...


# Applies when "flag" is set to False.
@overload
def func(a: int, flag: Literal[False]) -> str: ...


def func(a: int, flag: bool = True) -> Union[int, str]:
    return a if flag else str(a)
```

Example without the ellipsis that is equivalent to the previous example:
```py
# Applies when "flag" is omitted.
@overload
def func(a: int) -> int: ...

# Applies when "flag" is set to True.
@overload
def func(a: int, flag: Literal[True]) -> int: ...

# Applies when "flag" set to False.
@overload
def func(a: int, flag: Literal[False]) -> str: ...

def func(a: int, flag: bool = True) -> Union[int, str]:
    return a if flag else str(a)
```

But is is also just speculation. There is also an issue about this overloads behavior [here](https://github.com/python/mypy/issues/6580).

dobbersc (approved):

scripts/utility.py
MaxDall: Thank you for the detailed explanation :+1:

PR #514, opened 2024-05-12
Update LD Selector

addie9800:
When working with Fundus I found that way too much JSON is extracted when computing the self.precomputed elements. This causes multiple issues: 
- unnecessary but malformed JSON blocks cause avoidable crashes
- Additional JSON Blocks may cause irrelevant data to be extracted (e.g. the website (not article) author of the freiepresse test case: https://www.freiepresse.de/chemnitz/chemnitzer-hutfestival-wann-das-fest-steigt-was-neu-ist-und-wer-fuer-eine-gute-show-sorgen-wird-artikel13351201)
All current parsers mention the type within their JSON, only the old version of TheNamibian did not follow that convention, hence the override.

### DIFF
--- a/src/fundus/parser/base_parser.py
+++ b/src/fundus/parser/base_parser.py
@@ -26,10 +26,13 @@ import lxml.html
 import more_itertools
 from lxml.etree import XPath
 
+from fundus.logging import create_logger
 from fundus.parser.data import LinkedDataMapping
 from fundus.parser.utility import get_meta_content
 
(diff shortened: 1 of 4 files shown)


MaxDall:
To be honest, I would rather keep the generous LD selector and catch `JSONDecodeErrors` during parsing. You can fix the author for FreiePresse using `get_value_by_key_path` instead of `bf_search` .

addie9800 (commented):
I can't approve my own PR... But it looks good 👍

MaxDall (approved):

PR #517, opened 2024-05-16
Remove previous file when using `-o` option in test case script

MaxDall:
Previously, when using `-o` option to overwrite existing test cases, the HTML file and meta.info entry had to be removed by hand. This PR automates that step.


addie9800 (approved):
This is a really useful addition :)

PR #518, opened 2024-05-16
Fix author parsing for `BSZ`

MaxDall:
This PR replaces the current generic implementation of author parsing for `BSZ` as well as replacing the current test case with a more challenging one.


src/fundus/parser/data.py
```diff
@@ -85,7 +88,7 @@ def get_value_by_key_path(self, key_path: List[str], default: Any = None) -> Opt
             tmp = nxt
         return tmp
 
-    def bf_search(self, key: str, depth: Optional[int] = None, default: Any = None) -> Optional[Any]:
+    def bf_search(self, key: str, depth: Optional[int] = None, default: Optional[_T] = None) -> Union[Any, _T]:
```
addie9800, line 91: Why does it make sense to update the return value to `Union[Any, _T]`?

addie9800 (approved):
Thanks for the update 👍

src/fundus/parser/data.py
MaxDall: So mypy does know that if a default is given, at least that type is returned.

addie9800: Ah ok. I didn't know mypy was able to make that connection. Thanks

PR #520, opened 2024-05-17
Add sitemap filter to `BusinessInsiderDE`

MaxDall:


src/fundus/publishers/de/__init__.py
```diff
@@ -299,7 +299,11 @@ class DE(PublisherEnum):
         domain="https://www.businessinsider.de/",
         sources=[
```
addie9800, line 300: While we're at it, I also found a RSS Feed: https://www.businessinsider.de/feed/businessinsider-alle-artikel

addie9800 (approved):

PR #521, opened 2024-05-17
Remove unreachable source for `FreeBeacon`

MaxDall:


src/fundus/publishers/us/__init__.py
```diff
@@ -121,7 +121,7 @@ class US(PublisherEnum):
     FreeBeacon = PublisherSpec(
         name="The Washington Free Beacon",
         domain="https://freebeacon.com/",
-        sources=[NewsMap("https://freebeacon.com/post_google_news.xml")],
+        sources=[],
```
addie9800, line 124: There is a RSS Feed we could use as an altenative. The time I checked, they were having server issues (status 500), but it still seems to be valid and perhaps we can use it as soon as they're back online: https://freebeacon.com/feed/

addie9800: I also found the working sitemap: https://freebeacon.com/wp-sitemap.xml
The most recent post in there is from yesterday, so I think it's still up-to-date. It needs to be crawled in reverse though with an inverse filter of `posts-post`

addie9800 (changes requested):
Thanks for updating this 👍

addie9800 (approved):

PR #522, opened 2024-05-17
Update `TechCrunch`

MaxDall:
This PR:
- upgrades `TechCrunch` to adapt the newest layout
- fix a bug in the extraction of the previous version


src/fundus/publishers/us/techcrunch.py
```diff
@@ -14,10 +14,42 @@
...
+        def authors(self) -> List[str]:
+            return generic_author_parsing(self.precomputed.meta.get("author"))
+
+        @attribute
+        def publishing_date(self) -> Optional[datetime.datetime]:
+            return generic_date_parsing(self.precomputed.ld.bf_search("datePublished"))
+
+        @attribute
+        def title(self) -> Optional[str]:
+            return self.precomputed.meta.get("og:title")
```
addie9800, line 42: I would parse this from the JSON, since the og:title tag value has an extra `| TechCrunch` appended to the headline

src/fundus/publishers/us/techcrunch.py
```diff
@@ -14,10 +14,42 @@
 
 
 class TechCrunchParser(ParserProxy):
+    class V2(BaseParser):
+        _summary_selector: XPath = CSSSelector("div.entry-content > p#speakable-summary")
+        _paragraph_selector: XPath = CSSSelector("div.entry-content > p:not(#speakable-summary)")
```
addie9800, line 19: Do we want to also parse bulletpoint lists? If so, I have found one usage: https://techcrunch.com/2024/05/15/senate-study-proposes-at-least-32b-yearly-for-ai-programs/

src/fundus/publishers/us/techcrunch.py
addie9800: Actually the same problem also occurs with V1. Also, since (almost) only the selectors are changed, I think also a minor parser update (V1_1) would be enough to spare all the duplicate code.

addie9800 (changes requested):
Thanks for updating this

src/fundus/publishers/us/techcrunch.py
MaxDall: The headline referenced in the LD+JSON suffers from encoding problems. I left it as is and removed `| TechCrunch`

addie9800 (approved):

PR #531, opened 2024-06-07
Show details about incomplete articles in Publisher Coverage

addie9800:
When going through the failed publishers, I that it is a bit tedious to find the problem of why a certain article is not being parsed and the publisher coverage fails for a certain article. Hence I added some lines to the output to give an idea of how serious the fail is: e.g. there are no authors -> probably none given in the article and can just be ignored. No title or no plaintext -> something weird is going on and should be checked out. And also, we would know exactly where to look.

### DIFF
--- a/scripts/publisher_coverage.py
+++ b/scripts/publisher_coverage.py
@@ -68,7 +68,12 @@ def main() -> None:
                 else:
                     print(
                         f"❌ FAILED: {publisher_name!r} - No complete articles received "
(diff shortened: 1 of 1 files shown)


MaxDall (approved):
@addie9800 thanks for adding. I can sees this being somewhat misleading, because `incomplete_article` here doesn't necessarily reflect the parser state. It's just a random article extracted without filters in place, that could also be a format missing the attributes in the first place, like video articles. I will approve this anyways.

Issue #309, opened 2023-08-24
Collection of already tried publishers


MaxDall:
A collection of publishers that have been tried but could not be added to Fundus due to various problems. To add a new publisher please note the date you tried to add it and the reason your progress was stopped.

| Publisher | Date | Encountered Problem |
|-----------|------|---------------------|
| Wall Street Journal | 24.08.2023 | Paywall |

SugarTzu:
Date: 27.04.2024 
https://www.express.de/
urllib.error.HTTPError: HTTP Error 403: Forbidden
publisher uses bot protection ?

Feyrbrand:
Date: 28.04.2024
[https://medium.com/](https://medium.com/)
urllib.error.HTTPError: HTTP Error 403: Forbidden
publisher uses bot protection ?

PR #532, opened 2024-06-07
Remove Washington Times due to Block

addie9800:
Washington Times is blocking crawling and requests return with a `403 - Forbidden` Error. Since I don't see a solution, I would suggest removing Washington Times and updating #309 accordingly.


addie9800:
We will not remove the publisher, but add a deprecated flag

PR #533, opened 2024-06-10
Remove Reuters

addie9800:
Removal of Reuters due to Bot Protection. No articles are crawlable, we receive a `401 - Forbidden` Error


addie9800:
We will not remove the publisher, but add a deprecated flag

PR #534, opened 2024-06-12
Deprecated Flag for Uncrawlable Publishers

addie9800:
This PR adds a flag for uncrawlable publishers. The Publisher Coverage Script is updated to skip publishers marked as deprecated. Also, deprecated publishers are marked as such by strike-through in the supported_publishers.md

### DIFF
--- a/src/fundus/scraping/crawler.py
+++ b/src/fundus/scraping/crawler.py
@@ -128,10 +128,9 @@ def remove_query_parameters_from_url(url: str) -> str:
 
 class CrawlerBase(ABC):
     def __init__(self, *publishers: Publisher):
-        if not publishers:
-            raise ValueError("param <publishers> of <Crawler.__init__> has to be non empty")
-
         self.publishers: List[PublisherEnum] = list(set(more_itertools.collapse(publishers)))
+        if not self.publishers:
+            raise ValueError("param <publishers> of <Crawler.__init__> must include at least one publisher.")
 
     @abstractmethod
     def _build_article_iterator(
@@ -150,7 +149,6 @@ class CrawlerBase(ABC):
         only_complete: Union[bool, ExtractionFilter] = Requires("title", "body", "publishing_date"),
         url_filter: Optional[URLFilter] = None,
         only_unique: bool = True,
-        ignore_deprecated: bool = True,
         save_to_file: Union[None, str, Path] = None,
     ) -> Iterator[Article]:
         """Yields articles from initialized scrapers
@@ -173,8 +171,6 @@ class CrawlerBase(ABC):
                 URLs before download. This filter applies on both requested and responded URL. Defaults to None.
             only_unique (bool): If set to True, articles yielded will be unique on the responded URL.
                 Always returns the first encountered article. Defaults to True.
-            ignore_deprecated (bool): If set to True, Publishers marked as deprecated will not be skipped.
-                Defaults to True.
             save_to_file (Union[None, str, Path]): If set, the crawled articles will be collected saved to the
                 specified file as a JSON list.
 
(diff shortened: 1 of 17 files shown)


src/fundus/scraping/crawler.py
```diff
@@ -204,7 +204,10 @@ def build_extraction_filter() -> Optional[ExtractionFilter]:
                         f"is(are) not supported by {publisher.publisher_name}. Skipping publisher"
                     )
                 else:
-                    fitting_publishers.append(publisher)
+                    if not publisher.deprecated:
+                        fitting_publishers.append(publisher)
+                    else:
+                        logger.warning(f"Skipping deprecated publisher: {publisher.publisher_name}")
```
MaxDall, line 210: I would either introduce a parameter to include deprecated publisher, since it is still possible to crawl them depending on IP address and User Agent, or don't skip them during crawling.

MaxDall (changes requested):
Thanks for adding! Looks awesome 👍

src/fundus/scraping/crawler.py
addie9800: Good idea, I added that.

src/fundus/scraping/crawler.py
```diff
@@ -150,6 +150,7 @@ def crawl(
         only_complete: Union[bool, ExtractionFilter] = Requires("title", "body", "publishing_date"),
         url_filter: Optional[URLFilter] = None,
         only_unique: bool = True,
+        ignore_deprecated: bool = False,
```
MaxDall, line 153: I would prefer to set the default to `True`, it doesn't hurt if the deprecated publishers cannot be crawled, but if they can and are excluded by default most won't notice. As an example, crawling Reuters works perfectly fine for me.

src/fundus/scraping/crawler.py
```diff
@@ -209,7 +212,10 @@ def build_extraction_filter() -> Optional[ExtractionFilter]:
                         f"is(are) not supported by {publisher.publisher_name}. Skipping publisher"
                     )
                 else:
-                    fitting_publishers.append(publisher)
+                    if (not publisher.deprecated) or ignore_deprecated:
+                        fitting_publishers.append(publisher)
+                    else:
+                        logger.warning(f"Skipping deprecated publisher: {publisher.publisher_name}")
```
MaxDall, line 218: Here's a logic bug included. If `only_complete` is set to `False`, the `ignore_deprecated` flag doesn't work anymore, because it is the `if isinstance(extraction_filter, Requires)` block.

MaxDall (changes requested):

src/fundus/scraping/crawler.py
addie9800: Thanks for cathing this

MaxDall:
@addie9800 I moved the `ignore_deprecated` flag from the `crawl` method to `Crawler`. The paradigm of `crawl` is to only include parameters affecting all crawlers (forward, backward). Since the depreciation flag does not serve any purpose regarding backward crawling, I decided to move it to the crawler. What do you think about that?

addie9800:
> @addie9800 I moved the `ignore_deprecated` flag from the `crawl` method to `Crawler`. The paradigm of `crawl` is to only include parameters affecting all crawlers (forward, backward). Since the depreciation flag does not serve any purpose regarding backward crawling, I decided to move it to the crawler. What do you think about that?

I think it's a good idea. I also updated the docs a bit. Maybe you can also say what you think about that

MaxDall:
@addie9800 thanks for updating the documentation. Unfortunately the latest documentation commit breaks a lot of links. Also tutorial 4 now seems completely overwhelming and out of focus while tutorial 5 looks like it was cannibalized. The purpose of tutorial 4 is to enable the user to filter for specific articles. `search` or `deprecatrd` isn't as nearly as important here. The latter ones are mostly for convenience. I would let tutorial 4 as it was and add the deprecation mechanica to the former tutorial 5.

addie9800:
Oh, and I thought I had caught the broken links. Well, I undid the previous changes and just appended it to section 5 of the tutorial

MaxDall (approved):
Thanks a lot! Looks really great :)

PR #536, opened 2024-06-18
Add crawl timeout functionality

olaughter:
Adds an argument to the `crawl` method that will end the crawl if the time limit is reached.

I realised I wanted this feature for a project, and wanted to see if adding it back was of interest? I couldn't find an existing issue on it but believe it is useful. Especially in a context where specific filters are being used, as in some cases the crawler will continue running for extended periods without finding new articles. For example if there is a date filter for the current day, previous days will still be crawled and filtered out.

Happy to take advice on anything extra that is needed for this. Just let me know!


src/fundus/scraping/crawler.py
```diff
@@ -179,12 +185,15 @@ def crawl(
...
 
-        if max_articles == 0:
+        if max_articles == 0 or max_seconds == 0:
             return
 
         if max_articles is None:
             max_articles = -1
 
+        if max_seconds is None:
+            max_articles = math.inf
```
addie9800, line 195: Did you mean to overwrite `max_seconds`?
```suggestion
            max_seconds = math.inf
```

docs/1_getting_started.md
```diff
@@ -62,6 +62,12 @@ crawler = Crawler(PublisherCollection.us)
 # crawl one article from a publisher based in the US and print it
 for article in crawler.crawl(max_articles=1):
     print(article)
+
```
addie9800, line 65: I would suggest moving this example (lines 65-70) to line 91, so that this code does not conflict with the example output and possible code variation in the lines 73-89

addie9800 (changes requested):
Thanks a lot for adding this, we haven't thought of it yet, but it's definitely a useful addition. :)

src/fundus/scraping/crawler.py
olaughter: Oops, yes, thank you. Fixed!

docs/1_getting_started.md
olaughter: That makes sense. Fixed!

olaughter:
Think I've fixed the last of those linting issues 🙏

src/fundus/scraping/crawler.py
```diff
@@ -146,6 +148,7 @@ def _build_article_iterator(
     def crawl(
         self,
         max_articles: Optional[int] = None,
+        max_seconds: Optional[float] = None,
```
MaxDall, line 151: In general, I think `Int` as a type is a better fit here.

src/fundus/scraping/crawler.py
```diff
@@ -179,12 +185,15 @@ def crawl(
...
 
-        if max_articles == 0:
+        if max_articles == 0 or max_seconds == 0:
             return
 
         if max_articles is None:
             max_articles = -1
 
+        if max_seconds is None:
+            max_seconds = math.inf
```
MaxDall, line 195: `math.inf` is of type `float`. To resolve the typing issue I would suggest setting `max_seconds` to -1 here.

addie9800 (dismissed):
Thanks a lot for adding this 👍

MaxDall (changes requested):

src/fundus/scraping/crawler.py
olaughter: Thanks @MaxDall, I have switched it to now use `Int`

src/fundus/scraping/crawler.py
olaughter: Thanks, have updated 👍

MaxDall:
@olaughter Sorry for the delay, I'm currently working on this. What concerns me about your original idea is that it depends on receiving an article. Arguing alongside your concerns
> For example if there is a date filter for the current day, previous days will still be crawled and filtered out.

if the crawler stops after `max_seconds` depends on if `max_seconds` is short enough to still receive articles from that date range.
I altered the logic so that the main thread will be interrupted after `max_seconds`. My preferred solution would be to have an actual timeout where the process terminates of there weren't any articles received for `timeout` seconds, but that's another story.

olaughter:
> I will alter the logic so that the main thread will be interrupted after `max_seconds` and commit on your branch.

Thanks @MaxDall, this sounds good and I really like that idea, as you say it makes more sense for it not to be dependant on articles. Look forward to seeing your commit and if there is anymore I can assist with on this please shout!

MaxDall:
@olaughter I finished working on this. I finally decided to go with a timeout mechanic, enabling users to set a timeout. If the crawler does not receive a new article within the specified timeout period, it will terminate automatically. Do you mind give it a try if it benefits your project?

olaughter:
Thanks @MaxDall ! I have tried it out and it works really well. All looking good from my side!

MaxDall (approved):
Thanks @olaughter for the idea and the original implementation. This really helped move the project forward 🚀

MaxDall:
Maybe @addie9800 wanna give it a last look?

src/fundus/utils/timeout.py
```diff
@@ -0,0 +1,158 @@
...
+
+    @property
+    def time(self) -> float:
+        return max(0.0, time.time() - self._start)
+
+    def reset(self):
+        self._start = time.time()
+
+
+class Timer(threading.Thread):
```
addie9800, line 26: I am a bit confused about the naming of the class, since there also is a `threading.Timer` class, which also is referenced in this file and also inherits from `threading.Thread` (refer to other comment). In case the usage of the other implementation was intentional, I would suggest renaming this class to avoid confusion.

src/fundus/utils/timeout.py
```diff
@@ -0,0 +1,158 @@
...
+            Defaults to False.
+        callback: If given, will be called instead of raising KeyboardInterrupt. Defaults to None.
+
+    Returns:
+        Optional[T]: T if <func> returns before <time> seconds, else None if <silent> is set to True.
+    """
+
+    @wraps(func)
+    def wrapper(*args: P.args, **kwargs: P.kwargs) -> Optional[T]:
+        timer = threading.Timer(seconds, callback or _interrupt_handler)
```
addie9800, line 108: Here is the above mentioned usage of `threading.Timer`. Was this intentional, because it feels like you wanted to use your own implementation of `Timer` here.

addie9800 (changes requested):
Thanks for all these great additions and of course the idea in the first place. Sorry for the delay, I was a bit busy the past couple of days :)

src/fundus/utils/timeout.py
MaxDall: You're right, the names are confusing. I'm gonna rename my own implementation to `ResettableTImer`.

src/fundus/utils/timeout.py
MaxDall: That was intentional, but I get it doesn't look like, using ambiguous names.

addie9800 (approved):

alanakbik:
@olaughter and @MaxDall thanks a lot for adding this!

PR #537, opened 2024-06-19
Guard download and streaming of WARC files

MaxDall:
This PR guards the download and streaming of specific WARC files by catching `requests.HTTPError`s and `fastwarc.stream_io.StreamError`s and logging the corresponding WARC path at `error` level. This is in order to prevent Fundus from crashing when something went wrong during the processing of an individual WARC file.

This PR also adds a naive retry mechanic, that tries to reload said WARC file every 30 * current try seconds. By default, it does so three times.

### DIFF
--- a/src/fundus/scraping/crawler.py
+++ b/src/fundus/scraping/crawler.py
@@ -395,7 +395,7 @@ class CCNewsCrawler(CrawlerBase):
         url_filter: Optional[URLFilter] = None,
     ) -> Iterator[Article]:
         retries: int = 0
-        while True and retries <= self.retries:
+        while retries <= self.retries:
             source = CCNewsSource(*publishers, warc_path=warc_path)
             scraper = CCNewsScraper(source)
             try:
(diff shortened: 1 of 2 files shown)


src/fundus/scraping/crawler.py
```diff
@@ -371,19 +383,34 @@ def __init__(
...
         publishers: Tuple[PublisherEnum, ...],
         error_handling: Literal["suppress", "catch", "raise"],
         extraction_filter: Optional[ExtractionFilter] = None,
         url_filter: Optional[URLFilter] = None,
     ) -> Iterator[Article]:
-        source = CCNewsSource(*publishers, warc_path=warc_path)
-        scraper = CCNewsScraper(source)
-        yield from scraper.scrape(error_handling, extraction_filter, url_filter)
+        retries: int = 0
+        while True and retries <= self.retries:
```
addie9800, line 398: Doesn't this simplify to `retries <= self.retries`?

addie9800 (commented):
Thanks for tackling this

src/fundus/scraping/crawler.py
MaxDall: That's `True` *slapstick drum sound*

addie9800 (dismissed):

PR #538, opened 2024-06-19
Slow down WARC path requests

MaxDall:
This adds a random sleep between 0 and 3 seconds in between requests for WARC paths to prevent `503` errors.

### DIFF
--- a/src/fundus/scraping/crawler.py
+++ b/src/fundus/scraping/crawler.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import gzip
 import json
 import os
+import random
 import re
 import time
 from abc import ABC, abstractmethod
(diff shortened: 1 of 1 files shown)


src/fundus/scraping/crawler.py
```diff
@@ -464,14 +465,22 @@ def load_paths(url: str) -> List[str]:
...
+
             if self.processes == 0:
                 nested_warc_paths = [load_paths(url) for url in urls]
             else:
                 # use two threads per process, default two threads per core
                 max_number_of_threads = self.processes * 2
 
                 with ThreadPool(processes=min(len(urls), max_number_of_threads)) as pool:
-                    nested_warc_paths = pool.map(load_paths, urls)
+                    nested_warc_paths = pool.map(random_sleep(load_paths, (0, 3)), urls)
```
addie9800, line 483: I was thinking it might make sense to have a minimum wait of more than 0. Or was your approach that in case it crashes due to the SlowDown from CC that it'll be retried anyway after a short break, in which case it should go through again

MaxDall: My Idea was to spread the requests so we wouldn't bombard the server with min(process*2, 94) concurrent requests. So far it seems to work. Unfortunately, I don't understand the second part of your comment. We do not retry any connection here.

addie9800: Nevermind that, I think I had the PR in mind this was based on. Ok, it does seem to work. I was surprised though, because when I had a 5 second manual delay it still crashed yesterday, but I couldn't reproduce it earlier. Maybe it was my implementation

addie9800 (approved):

PR #539, opened 2024-06-19
Spread parallel requests for `CCNewsCrawler`

MaxDall:
So as it seems `CC-NEWS` we also have to spread the parallel task as well to avoid `503` errors while crawling. I tested this on `Gruenau` and it seems to work just fine.


addie9800 (approved):
That makes sense and looks good. Thanks 👍

PR #540, opened 2024-06-20
Expose loggers and update documentation

MaxDall:
This PR
- exposes all loggers in a dictionary called `loggers` mapping module name to logger
- adds a new utility function to easily add a handler to all existing loggers
- add a new tutorial about logging and update the documentation accordingly

### DIFF
--- a/docs/6_logging.md
+++ b/docs/6_logging.md
@@ -39,6 +39,12 @@ Or find a collection of all existing loggers with their module names here:
 
 ````python
 from fundus.logging import loggers
+
+# print all modules having loggers
+print("\n".join(sorted(loggers.keys())))
+
+# accessing the 'url' logger
+url_logger = loggers["fundus.scraping.url"]
 ````
 
 ## Changing log levels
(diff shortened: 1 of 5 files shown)


docs/6_logging.md
```diff
@@ -0,0 +1,67 @@
...
+You can import a specific logger from the corresponding module like this:
+
+````python
+from fundus.scraping.crawler import logger
+````
+
+Or find a collection of all existing loggers with their module names here:
+
+````python
+from fundus.logging import loggers
```
addie9800, line 38: I think it might be nice to add an example or overview of the module names here. Something along the lines of:

Fundus support logging in the follwoing modules: crawling, scraping...

If you want to access a certain logger, you would achieve this by
crawling_logger = loggers['crawling']

addie9800 (commented):
This is a helpful addition. Thanks 👍

docs/6_logging.md
MaxDall: Yeah, thought about that as well. The thing that kept me away was the maintenance effort. The added modules would be static, so every newly added logger has to be added by hand, which I'm 99% sure we will forget. So I thought it would be better to have no list at all rather than a potentially incomplete list, especially since the modules can still be viewed through the `loggers` variable. What du you think?

addie9800: I think my main concern was that it's a bit of extra work for the users to get access to the loggers names, or at least the format that is being used and with the given example this seems to be partially solved. I don't think though that an incomplete list is a problem as long as it's marked as such. But nevertheless, I think the example should be sufficient as well

addie9800 (approved):
Thanks for this update 👍

PR #542, opened 2024-06-20
Add progress bar for WARC file processing 

MaxDall:
This PR:
- adds a bar to display the progression of WARC file processing
- add a parameter to `CCNewsCrawler` to disable progress bars

### DIFF
--- a/src/fundus/scraping/crawler.py
+++ b/src/fundus/scraping/crawler.py
@@ -381,6 +381,7 @@ class CCNewsCrawler(CrawlerBase):
         end: datetime = datetime.now(),
         processes: int = -1,
         retries: int = 3,
+        disable_tqdm: bool = False,
         server_address: str = "https://data.commoncrawl.org/",
     ):
         """Initializes a crawler for the CC-NEWS dataset.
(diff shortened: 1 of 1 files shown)


addie9800 (commented):
This looks really good. Thanks for adding it. There is one question I have though: do you think it's easily possible to force the progressbar to be at the bottom of the output? Because if there are things printed to stdout, it gets lost pretty quickly. I tried playing around with it a bit, but there doesn't seem to be a quick fix.

MaxDall:
> do you think it's easily possible to force the progressbar to be at the bottom of the output?

No, I can't think of an easy or even practical solution for this. An alternative is always to disable tqdm.
Maybe [this](https://stackoverflow.com/questions/38543506/change-logging-print-function-to-tqdm-write-so-logging-doesnt-interfere-wit) can help?

addie9800:
> > do you think it's easily possible to force the progressbar to be at the bottom of the output?
> 
> No, I can't think of an easy or even practical solution for this. An alternative is always to disable tqdm. Maybe [this](https://stackoverflow.com/questions/38543506/change-logging-print-function-to-tqdm-write-so-logging-doesnt-interfere-wit) can help?

I tried it, but it doesn't really work, maybe because of the multiprocessing, I'm not sure. Disabling wouldn't help either, since I thought it would be nice to always have the current progess visible, even if logging is set to a broader level or the usage involves printing. But it's also not that important.

addie9800 (approved):

PR #544, opened 2024-06-27
Set ˋeval_booleansˋ to True in publisher_coverage

addie9800:
Since in most cases, there is no plaintext available to be parsed I would suggest restricting the publisher_coverage to assess articles that are marked to be free_access. While this doesn't have a big effect on whether or not a complete article can be found, it makes debugging easier, when an article is returned that is actually incomplete and not just a premium article


addie9800:
Closed due to misinterpretation of free_access attribute. (refer to #546 )

PR #545, opened 2024-06-27
Fix Heise

addie9800:
Heise had a minor layout update one week ago. This PR fixes it by changing a minor detail. As it turns out Heise has a gadget blog called techstage and I added support for that as well. Example link: https://www.techstage.de/test/mini-pc-fuer-190-euro-mit-seltenem-extra-ouvis-gk3-plus-mit-vga-anschluss-im-test/lllvhkg

### DIFF
--- a/src/fundus/publishers/de/heise.py
+++ b/src/fundus/publishers/de/heise.py
@@ -47,8 +47,6 @@ class HeiseParser(ParserProxy):
 
         @attribute
         def body(self) -> ArticleBody:
-            # with open(f"E:\\Temp\\{self.precomputed.ld.bf_search('headline')}.html", 'w', encoding='utf-8') as file:
-            #    file.write(self.precomputed.html)
             return extract_article_body_with_selector(
                 self.precomputed.doc,
                 summary_selector=self._summary_selector,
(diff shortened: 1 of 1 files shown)


MaxDall (changes requested):
Since the layout change a new version would be necessary to parse previously crawled articles.

addie9800:
> Since the layout change a new version would be necessary to parse previously crawled articles.

Actually, I did add a new version at first, but then I did decide against it, since the selectors now are just a generalized version of the previous ones and should be completely backwards compatible. The subheadline and summary selectors were just extended and the paragraph selector was also just extended and the existing part generalized.

src/fundus/publishers/de/heise.py
```diff
@@ -30,7 +30,18 @@ class V1(BaseParser):
...
+            "//div[@class='ringCommonDetail ringBlockType-paragraph ']/p[not(@class"
+            " or ((string-length(text()) < 3) and (contains(text(), '(') or contains(span, '(')))"
+            " or contains(text(), '=== Anzeige / Sponsorenhinweis')"
+            " or contains(text(), 'Tipp: Wir sind bei WhatsApp!')"
+            " or contains(a, 'heise+ abonnieren')"
+            " or contains(text(), '► '))"
+            " or @class='antwort rte__abs--antwort'"
+            " or @class='frage rte__abs--frage'"
+            " or @class='json-ld-paid-content-marker'] "
+            " | //div[@class='ringCommonDetail ringBlockType-paragraph ']//ul["
```
MaxDall, line 44: Is this all `techstage` related?

addie9800: Yes, they use a slightly different formatting.

addie9800: But I did realise that the selectors can be simplified, as techstage does not seem to contain as much bloat as heise

MaxDall (approved):

PR #546, opened 2024-06-27
fix free_access for various publishers

addie9800:
it turns out a couple of us publishers do not properly implement the isAccessibleForFree attribute. This PR addresses this.

### DIFF
--- a/src/fundus/publishers/us/la_times.py
+++ b/src/fundus/publishers/us/la_times.py
@@ -37,3 +37,8 @@ class LATimesParser(ParserProxy):
         @attribute
         def title(self) -> Optional[str]:
             return self.precomputed.meta.get("og:title")
+
+        @attribute
+        def free_access(self) -> bool:
+            # LA Times sets the corresponding JSON Element to False by default, but all articles are still available
+            return True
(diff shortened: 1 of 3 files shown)


MaxDall:
Hmm, I'm not sure about this. My concern here is to dilute the meaning of `free_access`. Do the publishers listed here just don't pay attention to the `accesibleForFree` field of the JSON and use a different method to restrict access or aren't they providing free articles at all?

addie9800:
> Hmm, I'm not sure about this. My concern here is to dilute the meaning of `free_access`. Do the publishers listed here just don't pay attention to the `accesibleForFree` field of the JSON and use a different method to restrict access or aren't they providing free articles at all?

From my understanding it is both. I think the LA Times literally just sets the value to false in every article, but does not actually restrict access and The Intercept for example lets you read 2 free articles while browsing the website before you have to enter your email address to continue reading (where my guess is why the mark the article to not be accessible for free). But from the idea I had from free_access was that the attribute marks articles that are physically behind a paywall and cannot be accessed at all or have parts of them hidden.

MaxDall:
> But from the idea I had from free_access was that the attribute marks articles that are physically behind a paywall and cannot be accessed at all or have parts of them hidden.

I thought it was meant to give users the option to respect the publishers' decision which of its articles are usable free of charge.

addie9800:
> I thought it was meant to give users the option to respect the publishers' decision which of its articles are usable free of charge.

Well OK, I see your point then. And considering the actual intention of the isAccessibleForFree tag, I think we should go with your interpretation.

PR #547, opened 2024-06-28
Fix The Mirror

addie9800:
This PR updates the selectors to be more flexible and also includes live tickers.


MaxDall (approved):

MaxDall (changes requested):
Did the layout change here as well or does it just include more articles now?

addie9800:
> Did the layout change here as well or does it just include more articles now?

No, this is just a more precise selector. The previous version assumed that the first two divs contain the summary and article, but this is not the case for articles with an audio version of the article where summary and content are in divs 1 and 3 respectively causing these to not be crawled properly.

MaxDall (approved):

PR #548, opened 2024-07-01
Use `timeout` parameter in coverage script instead of wrapper

MaxDall:


addie9800 (approved):
Nice, this simplifies a lot

PR #549, opened 2024-07-01
Update `EveningStandard` parser

MaxDall:
This adjusts the `EveniingStandard` parser to the latest version.


src/fundus/publishers/uk/evening_standard.py
```diff
@@ -40,3 +41,18 @@ def authors(self) -> List[str]:
         @attribute
         def topics(self) -> List[str]:
             return generic_topic_parsing(self.precomputed.meta.get("keywords"))
+
+    class V1_1(V1):
+        VALID_UNTIL = datetime.date.today()
+        _summary_selector = CSSSelector("div.sc-jgyXzG")
+        _subheadline_selector = CSSSelector("div#main div.sc-dFfFtc > h3")
```
addie9800, line 48: It seems like not all articles use this new layout. [This article](https://www.standard.co.uk/lifestyle/julian-assange-wikileaks-life-stella-moris-marriage-relationships-b1167169.html) was published today and follows the old layout (which would also miss the subheadlines)

addie9800: My guess are that some (but not all) of the lifestyle articles still use the old format for whatever reason: 
- https://www.standard.co.uk/lifestyle/wellness/modern-brain-clinic-dr-lesley-parkinson-clinic-adhd-depression-b1163587.html
- https://www.standard.co.uk/lifestyle/fashion/princess-diana-vivienne-westwood-elton-john-fashion-auction-boom-millions-b1167547.html

addie9800 (changes requested):
Thanks for the update 👍

src/fundus/publishers/uk/evening_standard.py
MaxDall: Hmm, I would refrain from supporting both layouts. For now, I think it's best to drop support for `lifestyle` articles.

addie9800 (approved):
Sounds good

PR #551, opened 2024-07-03
Fix haberturk selectors

MaxDall:
This fixes a bug with `Haberturk` parser selecting subheadlines as summaries.

see [this](https://www.haberturk.com/turkiye-de-orman-yanginlari-yanginlarda-son-durum-guncel-haberler-3700095) article.

I also replaced the test case to cover this problem.


addie9800 (changes requested):
This looks good. Thanks for updating it. When looking through some example articles, I noticed that the publishing date doesn't always get extracted. Perhaps we should switch to using the meta tag or at least use it as backup. This is the article I found for which it fails: https://www.haberturk.com/ingiltere-sandik-basinda-14-yillik-muhafazakar-parti-iktidari-sona-erecek-mi-3700214

addie9800 (approved):

PR #553, opened 2024-07-03
Rework logging and fix overwritten config

MaxDall:
## Fixes a bug when using `spawn` as start method for multiprocessing
Previously, when using `spawn` as a start method, already set logging configurations are overwritten, when starting a new process. This PR introduces a workaround, passing the current configuration to the `Pool` and initializing the logging module with said config when a new process is started for the `CCNewsCrawler`.

## Reworks parts of the logging module
IMO adding handlers and setting levels was still pretty tedious, even after the last rework. With this PR, the logger debug level is set to `DEBUG` by default, and one controls debugging with handler levels instead of the logger level.

### DIFF
--- a/src/fundus/logging.py
+++ b/src/fundus/logging.py
@@ -1,28 +1,110 @@
 import logging
-from typing import Dict
+from typing import Dict, Set, Union, cast
 
-__all__ = ["set_log_level", "add_handler", "create_logger", "loggers"]
+from fundus.utils.serialization import JSONVal
 
-loggers: Dict[str, logging.Logger] = {}
+_default_handler_level = logging.ERROR
 
+__all__ = ["set_log_level", "add_handler", "create_logger", "loggers", "handlers"]
+
+# create std-handler
 _stream_handler = logging.StreamHandler()
+_stream_handler.name = "std-handler"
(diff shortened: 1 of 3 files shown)


src/fundus/logging.py
```diff
@@ -1,28 +1,110 @@
...
+
+    formatters: Set[logging.Formatter] = cast(
+        Set[logging.Formatter], {handler.formatter for handler in handlers.values()}
+    )
+
+    def get_formatter_config(formatter: logging.Formatter) -> JSONVal:
+        return {"format": formatter._fmt}
+
+    def get_handler_config(handler: logging.Handler) -> JSONVal:
+        config: Dict[str, Union[int, str]] = {
```
addie9800, line 85:
```suggestion
        config: Dict[str, JSONVal] = {
```
Is there a reason to limit the values of the dictionary to int and str? Because allowing them to be of JSONVal would allow us to remove the type ignore in line 96

addie9800 (changes requested):
Thanks for the rework 👍

src/fundus/logging.py
MaxDall: No, you're absolutely right! That's way better 😅

addie9800 (approved):

PR #555, opened 2024-07-04
Fix Funke topics

addie9800:
Apparently the newpapers belonging to the Funke group have discontinued using the keywords keyword. For bigger articles they now have suggestions for similar categories, I would propose using these as new topics.


MaxDall:
Thank you very much for fixing this!

What do you think about `re.sub` instead of `.split(X)[0].strip()`

MaxDall (approved):

PR #559, opened 2024-07-12
Rework `Article` class

MaxDall:
### Overview
This PR presents a comprehensive overhaul of the Article class to improve data handling and maintainability.

### Key Changes
- **Introduction of Data Views**: The refactor emphasizes maintaining raw extraction data as the primary source while providing structured `Views` for data access. This approach facilitates easier benchmarking and evaluation.
  
- **Implementation of Custom Attribute Handling**: Custom `__getattribute__` and `__setattr__` methods replicate Descriptor-like behavior, ensuring controlled attribute access without relying on class attributes.

- **Read-Only Attributes**: Initially, all extraction data is read-only to enhance security and stability. This simplifies data integrity management, with provisions for future writable features.

This PR serves as preparation for the upcoming CC-NEWS benchmark!

### DIFF
--- a/src/fundus/scraping/article.py
+++ b/src/fundus/scraping/article.py
@@ -1,11 +1,9 @@
-from dataclasses import dataclass, field, fields
 from datetime import datetime
 from textwrap import TextWrapper, dedent
-from typing import Any, Dict, Iterator, List, Optional, Set, Tuple
+from typing import Any, Dict, List, Mapping, Optional
 
 import langdetect
 import lxml.html
-import more_itertools
 from colorama import Fore, Style
 
 from fundus.logging import create_logger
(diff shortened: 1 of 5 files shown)


src/fundus/scraping/article.py
```diff
@@ -16,37 +14,76 @@
...
-        unvalidated_attributes: Set[str] = set()
-        for attribute, value in extracted_unvalidated:
-            object.__setattr__(article, attribute, value)  # Sets attributes on a frozen dataclass
-            unvalidated_attributes.add(attribute)
-        object.__setattr__(article, "_unvalidated_attributes", unvalidated_attributes)
+        # create descriptors for attributes that aren't pre-defined as properties.
+        for attribute in extraction.keys():
+            if not hasattr(self, attribute):
+                setattr(self, attribute, AttributeView(attribute, self.__extraction__))
+
```
addie9800, line 42: If we're already making properties for all the important properties of the article, I would also consider adding one for the responded url.

src/fundus/scraping/article.py
```diff
@@ -16,37 +14,76 @@
...
+    def free_access(self) -> bool:
+        return self.__extraction__.get("free_access") or False
+
+    @property
+    def publishers(self) -> str:
+        return self.html.source_info.publisher
+
+    def __getattribute__(self, item: str):
+        if (attribute := object.__getattribute__(self, item)) and hasattr(attribute, "__get__"):
+            return attribute.__get__(self, type(self))
```
addie9800, line 73: When does this case come into play? I tried playing around a bit, but didn't find a setting where it is used.

addie9800 (changes requested):
Thank you for the rework. This looks really good 👍

src/fundus/scraping/article.py
```diff
@@ -16,37 +14,76 @@
...
+    def __setattr__(self, key: str, value: object):
+        if hasattr(self, key):
+            # we can't use getattr here, because it would invoke __get__, so unfortunately no default value
+            attribute = object.__getattribute__(self, key)
+            if hasattr(attribute, "__set__"):
+                attribute.__set__(key, value)
+                return
+        object.__setattr__(self, key, value)
+
+    def __getattr__(self, item: str):
```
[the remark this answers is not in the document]
MaxDall: The purpose of `__getattr__` here is to prevent IDEs like pycharm from flagging attribute access for attributes not described in the constructor. In our case, every access on an unvalidated attribute would get flagged by IDEs. I did this in order to underline our intended design.

src/fundus/scraping/article.py
```diff
@@ -16,37 +14,76 @@
...
+    @property
+    def topics(self) -> List[str]:
+        return self.__extraction__.get("topics") or []
+
+    @property
+    def free_access(self) -> bool:
+        return self.__extraction__.get("free_access") or False
+
+    @property
+    def publishers(self) -> str:
```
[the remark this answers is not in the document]
MaxDall: Your right, that's a typo.

src/fundus/scraping/article.py
MaxDall: Hmm, I think then we also have to do one for requested URL and that's getting a bit overwhelming. I totally get your point here and I think you're totally right, dunno yet if that's the best solution.

src/fundus/scraping/article.py
MaxDall: Every unvalidated attribute, more precise, every attribute that is added as `AttributeView` rather than a property is accessed this way. 
You can see this in action in the `test_extraction_view_getter` test case.

src/fundus/scraping/article.py
addie9800: Ok, then we can just keep it in mind for future updates

addie9800 (approved):

PR #563, opened 2024-07-18
Fix a bug with `HTMLTestFile` path generation

MaxDall:


addie9800 (approved):
Thanks for fixing this 👍

PR #565, opened 2024-07-18
Add swiss newspaper `NZZ`

MaxDall:


src/fundus/publishers/ch/__init__.py
```diff
@@ -17,3 +19,13 @@ class CH(metaclass=PublisherGroup):
...
         ],
     )
+
+    NZZ = Publisher(
+        name="Neue Zürcher Zeitung (NZZ)",
+        domain="https://www.nzz.ch/",
+        parser=NZZParser,
+        sources=[
+            NewsMap("https://www.nzz.ch/sitemap/news0.xml"),
+            Sitemap("https://www.nzz.ch/sitemap.xml", sitemap_filter=inverse(regex_filter(r"sitemap/[\d]{4}/[\d]{2}"))),
```
addie9800, line 29: There would also be some available RSS Feeds: https://www.nzz.ch/information/rss-ld.105364

src/fundus/publishers/ch/nzz.py
```diff
@@ -0,0 +1,49 @@
...
+    generic_author_parsing,
+    generic_date_parsing,
+)
+
+
+class NZZParser(ParserProxy):
+    class V1(BaseParser):
+        _summary_selector = CSSSelector("div.headline p.headline__lead")
+        _subheadline_selector = CSSSelector("div.article h2.subtitle, div.article h5.articlecomponent")
+        _paragraph_selector = CSSSelector(
```
addie9800, line 20: This [article](https://www.nzz.ch/international/usa-das-wichtigste-zum-rennen-um-das-weisse-haus-ld.1716364) has a table of contents, that is parsed as well.

addie9800: Also, as it seems this only occurs in English articles, but the last line with `Latest Articles` is parsed as well. See here: https://www.nzz.ch/english/us-election-2024-the-latest-polls-on-trump-and-biden-ld.1829123

addie9800 (changes requested):
Nice addition 👍

src/fundus/publishers/ch/__init__.py
MaxDall: I would argue here that those are not necessary here, judging by the news map.

addie9800 (approved):

PR #569, opened 2024-07-19
Add ˋTimes Of Indiaˋ

addie9800:
This adds the first indian newspaper. Unfortunately the default country code (in) is a python keyword, which is why I chose to use the three letter alternative.


src/fundus/parser/utility.py
```diff
@@ -232,11 +233,16 @@ def generic_author_parsing(
         value:      An input value representing author(s) which get parsed based on type
         split_on:   Only relevant for type(<value>) = str. If set, split <value> on <split_on>,
             else (default) split <value> on common delimiters
+        default_authors: Allows to filter out default authors like 'NewsDesk' or 'Redaktion'
```
MaxDall, line 236: I think having this default parameter overloads the function. There is a utility function `apply_substitution_pattern_over_list` to substitute strings in a list of strings. I think this may be more suitable for this use case.

src/fundus/publishers/ind/times_of_india.py
```diff
@@ -0,0 +1,80 @@
...
+        @attribute
+        def publishing_date(self) -> Optional[datetime.datetime]:
+            return generic_date_parsing(self.precomputed.ld.bf_search("datePublished"))
+
+        @attribute
+        def authors(self) -> List[str]:
+            return generic_author_parsing(
+                self.precomputed.ld.bf_search("author"),
+                default_authors=["TOI Sports Desk", "TOI News Desk", "TOI Tech Desk"],
+            )
```
MaxDall, line 57: Maybe it would be easier to just remove every author including the literal `TOI`

MaxDall (changes requested):
Thanks for adding! 👍 Besides some minor things this looks good

MaxDall:
Looks good 👍 one last thing: I think it would be nice to have a test case including a subheading as well.

MaxDall:
It seems like [this article](https://timesofindia.indiatimes.com/city/delhi/pak-based-terrorist-sounds-train-alarm/articleshow/112846333.cms) fails to get fully extracted. The first paragraph is missing.

[Here](https://timesofindia.indiatimes.com/business/india-business/cabinet-approves-setting-up-of-12-industrial-smart-cities-check-full-list/articleshow/112863664.cms) `Full list of industrial cities ...` is missing.

Also, I am wondering if there is an article having a summary and subheadings since the new test case only has subheadings.

addie9800:
> Collaborator

Thanks, it seems like I was tired at the time. Apparently they added an extra space throwing off my selector and I didn't catch it. I cleaned up a bit as well and it should be better now.

MaxDall:
> [Here](https://timesofindia.indiatimes.com/business/india-business/cabinet-approves-setting-up-of-12-industrial-smart-cities-check-full-list/articleshow/112863664.cms) Full list of industrial cities ... is missing.

Unfortunately the headline is still missing.

addie9800:
> > [Here](https://timesofindia.indiatimes.com/business/india-business/cabinet-approves-setting-up-of-12-industrial-smart-cities-check-full-list/articleshow/112863664.cms) Full list of industrial cities ... is missing.
> 
> Unfortunately, the headline is still missing.

OK, found the issue. Sorry for all this back and forth. This time it should really be good to go.

MaxDall (approved):
Thanks for adding! Looks good 👍

Issue #441, opened 2024-04-22
[Question]: News Publisher has two News site maps, what to do?


BorisKalika:
### Question

Hello :) 

I want to add a new UK publisher to the list. When scouring for the news sitemap of the publisher through robots.txt, I discovered that it has 2 ("part1" && "part2") news sitemaps. 

What should I do in this case?

MaxDall:
Hey @BorisKalika thank you for thinking about contributing :)

Could you provide me with the publisher's domain/name? Then I gonna give it a look and get back to you.

BorisKalika:
Hello @MaxDall :)

It was BBC https://www.bbc.co.uk/ :)

Bots.txt gave me a Newsmap that was (for some reason) split into two parts.

MaxDall:
Just to check if I understand you correctly: You're talking about https://www.bbc.co.uk/sitemaps/https-index-uk-news.xml ?

If so, the map you're referring to is an `indexmap`. There is also a part about those in the documentation.

Just add `NewsMap('https://www.bbc.co.uk/sitemaps/https-index-uk-news.xml')` as a source and Fundus will handle the rest. Per default, the package crawls through sitemaps recursively.

BorisKalika:
Thank you! As I have already added The Sun, anyone is free to add BBC if they feel like it! (I don't want to hog newspapers)

If no one claims it till Thursday, I will add it myself.

Best regards :)!

PR #572, opened 2024-07-25
Add The BBC as a UK publisher

olaughter:
This adds The BBC as a UK news publisher.

I did find an issue for this, it was marked completed but seems the publisher was never added: https://github.com/flairNLP/fundus/issues/441

Do let me know if I have missed anything. I'm looking forward to hearing any feedback you might have or ideas for improvement.


src/fundus/publishers/uk/the_bbc.py
```diff
@@ -0,0 +1,53 @@
...
+class TheBBCParser(ParserProxy):
+    class V1(BaseParser):
+        _subheadline_selector = CSSSelector("div[data-component='subheadline-block']")
+        _paragraph_selector = CSSSelector("div[data-component='text-block']")
+
+        @attribute
+        def body(self) -> ArticleBody:
+            return extract_article_body_with_selector(
+                self.precomputed.doc,
+                summary_selector=self._subheadline_selector,
```
addie9800, line 23: You accidentally added your subheadline selector as your summary selector

src/fundus/publishers/uk/the_bbc.py
```diff
@@ -0,0 +1,53 @@
...
+    extract_article_body_with_selector,
+    generic_author_parsing,
+    generic_date_parsing,
+)
+
+
+class TheBBCParser(ParserProxy):
+    class V1(BaseParser):
+        _subheadline_selector = CSSSelector("div[data-component='subheadline-block']")
+        _paragraph_selector = CSSSelector("div[data-component='text-block']")
```
addie9800, line 17: There are some articles that have a summary, for example this one: [https://www.bbc.com/news/world-us-canada-67285325](https://www.bbc.com/news/world-us-canada-67285325). Could you add a summary_selector for that as well? Then you might have a problem splitting up the paragraph and summary, but it should be fine to consider each p node to be its own paragraph, even if that means in this case that each sentence is it's own paragraph.

addie9800 (changes requested):
Hi, 
thank you for adding the new publisher and for the extension of our `generic_author_parsing`. It looks good so far, I only noticed some minor issues. :)

src/fundus/publishers/uk/__init__.py
```diff
@@ -149,3 +150,12 @@ class UK(metaclass=PublisherGroup):
...
         ],
     )
+
+    BBC = Publisher(
+        name="The BBC",
+        domain="https://www.bbc.co.uk/",
+        parser=TheBBCParser,
+        sources=[
+            NewsMap("https://www.bbc.co.uk/sitemaps/https-index-uk-news.xml"),
+        ],
```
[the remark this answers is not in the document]
olaughter: Thanks @addie9800 ! I've added a Sitemap as well as ending up adding a filter applying to both the Sitemap and the Newsmap

src/fundus/publishers/uk/the_bbc.py
olaughter: Woops, good spot, thanks!

src/fundus/publishers/uk/the_bbc.py
olaughter: Thanks @addie9800, adding a summary_selector makes sense. I'm having some trouble implementing it though. Looking at the html in the example there, the summary text block uses the same data-component and class as the text block elements in the rest of the article. Are you able to elaborate some more on what it would look like to select the summary?

addie9800: Sure thing. What I thought might work is exploiting that the summary is bolded and using a Selector along the lines of `div[data-component=text-block] > p > b`. Then you would only have to make sure to not also select them with the paragraphs. Now I am unsure whether or not this kind of criteria is supported by CSSSelector, but it should definitely be supported by XPath.

olaughter: Thanks for the advice @addie9800, I think I have found a solution that works. It required a bit of care in the end to avoid excluding any extra b tags from paragraphs later on in the article, but this seems to be working consistently well now. Let me know what you think

MaxDall:
@olaughter Thanks a lot for adding the `BBC` to Fundus!

PR #575, opened 2024-08-05
Improve contribution documentation about `ArticleBody`

MaxDall:
I noticed that `subheadline` often gets confused with what Fundus would call `summary`. To address this, I added a part to the contribution tutorial about the explicit parts of `ArticleBody`


docs/images/newspaper_labels_bold.pdf
addie9800, line 1: This seems to be leftover.

addie9800 (changes requested):
This is a great addition, that is something I have observed as well 👍

addie9800 (approved):

PR #584, opened 2024-08-12
Add `NationalPost`

addie9800:


src/fundus/publishers/ca/national_post.py
```diff
@@ -0,0 +1,62 @@
...
+            preliminary_topics = self.precomputed.ld.bf_search("keywords")
+            filter_list = ["Curated", "News", "Newsroom daily", "story", "Canada", "World"]
+            filtered_topics = [
+                topic
+                for topic in preliminary_topics
+                if "NLP Entity Tokens" not in topic
+                and "NLP Category" not in topic
+                and topic not in filter_list
+                and not re.search(r"[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}", topic)
+            ]
```
MaxDall, line 61: Maybe some sort of regex filter would be nice here. Actually, I think you could just use the regular `regex_filter`.

src/fundus/publishers/ca/__init__.py
```diff
@@ -16,3 +17,14 @@ class CA(metaclass=PublisherGroup):
...
         ],
     )
+
+    NationalPost = Publisher(
+        name="National Post",
+        domain="https://nationalpost.com",
+        parser=NationalPostParser,
+        sources=[
+            NewsMap("https://nationalpost.com/sitemap-news.xml"),
+            Sitemap("https://nationalpost.com/sitemap.xml"),
```
MaxDall, line 27:
```suggestion
            Sitemap("https://nationalpost.com/sitemap-old.xml"),
```

tests/resources/parser/test_data/ca/NationalPost.json
```diff
@@ -0,0 +1,38 @@
...
+  "V1": {
+    "authors": [
+      "Special to National Post"
+    ],
+    "body": {
+      "summary": [
+        "A draft policy by the CMA essentially calls for an end to private workplace coverage for virtual health care"
+      ],
+      "sections": [
+        {
```
MaxDall, line 11: I think it would be good to add a test case including at least one subheadline.

MaxDall (changes requested):
Thanks for adding 👍

src/fundus/publishers/ca/national_post.py
addie9800: I do agree. Using regex_filter seems a bit cursed Typing wise, but it does work just fine.

MaxDall (approved):
Thanks for adding! Looks good 👍

PR #588, opened 2024-08-23
Added Krautreporter

dkm1006:
I have two questions: 
1) Is there a way to filter urls in the Publisher specification to avoid requesting pages like impressum, etc.?
2) The JSON-LD here is wrapped in a CDATA tag - I think it might be cleaner to handle this in the _base_setup of the BaseParser (I've prepared an implementation, too)


addie9800:
Hi @dkm1006,
Thank you for adding Krautreporter. To answer your questions:
1. Yes, `Sitemap` and `NewsMap` objects have an attribute called `sitemap_filter` which expects a `regex_filter`. You can find more details [here](https://github.com/flairNLP/fundus/blob/master/docs/how_to_add_a_publisher.md#finishing-the-publisher-specification) about how that can be done.
2. Since this seems to be specific to only this one publisher, I was thinking it may be best to just override `_base_setup` locally for `Krautreporter` to keep potential future fixes for other publishers from aggregating in that function. But maybe we can see what @dobbersc thinks as well.

dkm1006:
@addie9800 thanks for your response! 
1) It’s my understanding that sitemap_filter only filters links to sub sitemaps (i.e. in a recursive call). 
2) Generally, I agree. However, a quick google search though showed that this CDATA enclosed JSON-LD may be (relatively) common. The solution could be extended to stripping the json of all comments.

addie9800:
Ok, my bad. I didn't read your question well enough and you're absolutely correct, that's the way the sitemap_filter works and in this case unfortunately there is no option I am aware of that would filters urls from a the publisher specification. At least I have also struggled with it in the past. But that might be an idea for future work.

I think stripping the json of all comments might be my preferred solution. This would then also cover potential other future issues. I did some research and it seems like `json5` could be an option, since it natively supports and ignores (single and multiline) comments while remaining a superset of `json`.

dkm1006:
Okay, thank you, I‘ll see if I can add something like a url_filter parameter to the Sitemap. 
`json5` in addition to adding an external dependency apparently is very slow and does more than needed - maybe a simple regex would suffice for 99% of cases? I’ll update my PR with a proposal.

MaxDall:
Hey @dkm1006 thanks for the PR ;)

- Regarding your first question: Yes there is. The `Publisher` class has an init parameter `url_filter` which allows you to define a std URL filter used for this specific publisher. I.e. writing something like this:
```python
Publisher(
    ....,
    ....,
    url_filter = regex_filter("/video/")
)
```
would filter out all URLs including a subpath `/video/` for the specified publisher.

- Regarding your second question: I like the idea of having a dedicated LD parsing function. It looks way cleaner now! Thanks a lot. I just made it a bit more generic by capturing only text content inside curly brackets instead of removing the specific `CDATA` tag.

dkm1006:
Thanks for the hint with the `url_filter` parameter! This is almost exactly what I was looking for. Here the RSS feed has a different url pattern than the sitemap (i.e. the RSS feed has links without topics in the path (`https://krautreporter.de/5335-fast-alles-was-du-uber-die-eu-dachtest-stimmt-nicht-mehr`) whereas the sitemap includes topics (`https://krautreporter.de/politik-und-macht/5335-fast-alles-was-du-uber-die-eu-dachtest-stimmt-nicht-mehr`). This makes it impossible to write a global filter that is exact but I think this is a minor issue, as the simple url filter I've added catches 99% of urls that should be filtered out.

I've also changed the json_pattern a little to account for whitespace between `[` and `{` (e.g.  `[ {}, {} ]` would result in an error for the previous pattern).

src/fundus/publishers/de/krautreporter.py
```diff
@@ -0,0 +1,46 @@
...
+from fundus.parser import ArticleBody, BaseParser, ParserProxy, attribute, utility
+
+
+class KrautreporterParser(ParserProxy):
+    class V1(BaseParser):
+        _summary_selector = CSSSelector("p.article-headers-standard__teaser")
+        _subheadline_selector = CSSSelector("div.article-markdown > h2")
+        _paragraph_selector = CSSSelector("div.article-markdown > p")
+        _json_ld_selector = CSSSelector('script[type="application/ld+json"]')
+        _topic_selector = CSSSelector("div.article-headers-shared-topic")
```
addie9800, line 16: Apparently there's a different article type: Newsletter. They have a slightly different structure. [Here]( https://krautreporter.de/sinn-und-konsum/5485-jetzt-sei-gefalligst-glucklich) is an example. Also, if the parsing of the topic fails, it raises an `IndexError`, which should be caught as well

src/fundus/publishers/de/krautreporter.py
```diff
@@ -0,0 +1,46 @@
...
+from lxml.cssselect import CSSSelector
+
+from fundus.parser import ArticleBody, BaseParser, ParserProxy, attribute, utility
+
+
+class KrautreporterParser(ParserProxy):
+    class V1(BaseParser):
+        _summary_selector = CSSSelector("p.article-headers-standard__teaser")
+        _subheadline_selector = CSSSelector("div.article-markdown > h2")
+        _paragraph_selector = CSSSelector("div.article-markdown > p")
```
addie9800, line 14: In most articles the last line is often something like _Redaktion: Brigitte Wenger, Schlussredaktion: Susan Mücke, Bildredaktion: Gabriel Schaefer; Audioversion: Iris Hochberger_ 
I would consider filtering those out. I think it should be relatively easy since the p elements do not have any own text and only contain one em element as a child. At least with XPath it should simple with something like  `p[text()]` or `p[not(em)]`, although the latter may be riskier.

addie9800 (changes requested):
Thank you for all the effort you put into this PR :) I just have two minor comments and then you should be good to go

src/fundus/publishers/de/krautreporter.py
dkm1006: I think this is dangerous, as I’ve seen articles with paragraphs that contain only text wrapped in an em-tag. 
We could maybe filter out those that are the last child of the article-markdown div and only contain an em-tag?

addie9800: Sounds good 👍

MaxDall: Found a solution that builds on the assumption that those paragraphs are proceeded by `hr` tags.

```XPath
//div[contains(@class, 'article-markdown')] /p[not(preceding::hr)]
```

MaxDall: Okay, it seems that this assumption is also false 😅 Looks like we have to use regex.

MaxDall: I adjusted the selector a bit so now every `p` without text directly following an `hr` element isn't selected. There is still a bit of risk but I think that might be our best option besides returning to regexes.

```
//div[contains(@class, 'article-markdown')] /p[text() or not(preceding-sibling::*[1][self::hr])]
```

addie9800 (approved):
This looks good, thank you 👍

PR #595, opened 2024-08-27
Update `BR` parser

MaxDall:


src/fundus/publishers/de/br.py
```diff
@@ -14,6 +15,8 @@
 
 class BRParser(ParserProxy):
```
addie9800, line 16: `publishing_date` parsing fails for Meldungen, since the publishingDate attribute does not exist in that [case]( https://www.br.de/nachrichten/meldung/messerangreifer-von-moers-war-wohl-psychisch-krank%2C3006aa526). It is mentioned loosely in the source, so perhaps it could be parsed from there.

addie9800 (changes requested):
Thanks for updating it :)

addie9800 (approved):
👍

PR #597, opened 2024-08-28
Fix bug in test file generation

addie9800:
When overriding the parser test file data multiple times in a row, the script will delete the saved html every other time. Simply reordering the commands fixes this. Additionally I updated the error message to also reflect the publisher where it occured, even though it is visible by looking though the pytest summary as well, I thought this is simpler because all the necessary information is at the same place.

### DIFF
--- a/scripts/generate_parser_test_files.py
+++ b/scripts/generate_parser_test_files.py
@@ -116,6 +116,10 @@ def main() -> None:
                 if not (article := get_test_article(publisher, url)):
                     logger.error(f"Couldn't get article for {publisher.name}. Skipping")
                     continue
+
+                # remove previous file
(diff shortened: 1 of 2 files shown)


MaxDall:
I have to admit I quite don't understand the problem here. Could you explain it a bit more in detail to me?

addie9800:
> I have to admit I quite don't understand the problem here. Could you explain it a bit more in detail to me?

Sure. If you run this code snippet once: `python -m scripts.generate_parser_test_files -p Tagesschau -o`, it will do just what you would expect it to do. But, in case you run it again on the same date, the script will delete the saved html without replacing it with the "new" html. It's most obvious if you run the command above and look at the edited filed in the git section of PyCharm.

MaxDall (approved):

PR #599, opened 2024-08-29
Guard key access for RSSFeed entries

MaxDall:
This fixes a bug regarding RSS feeds.

While parsing the feeds and yielding resulting URLs, the current implementation is based on the assumption that every feed entry has at least a `link` field, which is not the case.

This PR guards the key access.

See [this](https://github.com/flairNLP/fundus/actions/runs/10616770184/job/29427987839#step:5:145) for further details.


addie9800 (approved):
Thanks for fixing this 👍

PR #603, opened 2024-09-03
Fix `APNews`

MaxDall:


tests/resources/parser/test_data/us/APNews.json
```diff
@@ -25,64 +25,129 @@
...
-      "Activision Blizzard Inc",
+      "Kamala Harris",
+      "Donald Trump",
+      "Pennsylvania",
+      "Georgia",
       "General news",
+      "Domestic News",
+      "NE State Wire",
+      "MN State Wire",
+      "ME State Wire",
```
addie9800, line 123: This seems to be a whole lot of topics. I would consider the filtering the following three types:
- Anything containing `State Wire`
- One letter topics such as: p, s, w, n etc
- Anything containing news
I think this should still leave us with enough meaningful topics.

addie9800 (approved):
Looks good 👍

PR #606, opened 2024-09-04
Add Images

addie9800:


addie9800:
Ah well, sorry about that. I did only intend to push one extra file :)... I cleaned up a bit now, in case you want to have a look at it, but I haven't reached a real milestone yet, since you last had a peek at it. I think the issue I am struggling with most at the moment is the dynamic rescaling of images where some publishers change the path of the url according to the necessary resolution for the given screen, making it difficult to come up with a selector for the corresponding `img` element. If you have any idea, shoot ;)

addie9800:
Here are some examples, of problematic cases:
- The file name is changed to reflect resolution and width of an image: The article: https://www.spiegel.de/netzwelt/apps/instagram-beschraenkt-accounts-von-teenagern-a-3b65f364-e98b-4a31-8cfc-2d7ec4fc19bf#ref=rss has three versions of the same image in it's JSON: https://cdn.prod.www.spiegel.de/images/50b64eef-9d82-4021-93dc-732207c263cc_w1200_r1.778_fpx70_fpy50.99.jpg , https://cdn.prod.www.spiegel.de/images/50b64eef-9d82-4021-93dc-732207c263cc_w1200_r1.33_fpx70_fpy50.99.jpg and https://cdn.prod.www.spiegel.de/images/50b64eef-9d82-4021-93dc-732207c263cc_w1200_r1_fpx70_fpy50.99.jpg . Yet none of these links are actually in an `img` element on the website. In my browser the actual link is: https://cdn.prod.www.spiegel.de/images/50b64eef-9d82-4021-93dc-732207c263cc_w960_r1.778_fpx70_fpy50.99.jpg 
- Similarly, some publishers change the path to the image based on the image resolution: in this article: https://www.welt.de/politik/deutschland/article253552426/Christian-Lindners-Steuerplaene-entlasten-Gutverdiener-am-staerksten.html there is one image in the JSON: https://img.welt.de/img/politik/deutschland/mobile253552538/4727938867-ci16x9-w1200/Session-of-the-lower-house-of-German-parliament-Bundestag-in-Ber.jpg but in my browser, this image is used: https://img.welt.de/img/politik/deutschland/mobile253552538/4727938867-ci23x11-w20/Session-of-the-lower-house-of-German-parliament-Bundestag-in-Ber.jpg

addie9800:
Update: As of now, I have verified the functionality for TheNamibian, DerStandard, ORF, NineNews and CBCNews. They can be used to get an impression of the intended functionality.

MaxDall:
@addie9800 You can eliminate the URL parameter using the following selector on the HTML file.
```python
url_selector = XPath("//meta[@property='og:url']/@content]")
```

addie9800:
> @addie9800 You can eliminate the URL parameter using the following selector on the HTML file.

Thanks, I changed that

> Further, I would like to hear your opinion on this: With this implementation, we get the matching `ImageObject` from the ld for every image candidate we identified, so long as there is an image object present at least 😅. Do you say it is worth the additional effort? Or is solely taking information from the "HTML" sufficient?

I think this: `so long as there is an image object present at least` is the 'springender Punkt'. From what I've seen it's too much of a hassle to actually find the ImageObjects within the JSON, because many publishers have slightly different implementations. Then, not all images in there are necessarily relevant, or have the same URL as the images actually used in the HTML. And, a lot of publishers don't include any extra information, some just list image URLs. Lastly, these lists may also be incomplete, only containing the cover images. Long story short - I think it's too much extra work, with probably no(t much) gain, since all the info from the JSON, at least in the publishers we integrated up till now, is also (easily) available in the HTML.

> Another idea that came to my mind while looking at the `similarity_threshold`: Wouldn't it be more sufficient to use the bounds within `load_images_from_html` as well? Wouldn't this replace the entire merging part? Or at least reduce the number of potential candidates a lot?

You might be right about that. I think it can be a step to integrate the `merge_images` into the `load_images_from_html` function. It's mostly still a relic from when we sourced the images from json and html and had to map them together.

src/fundus/parser/utility.py
```diff
@@ -371,3 +392,218 @@ def parse_title_from_root(root: lxml.html.HtmlElement) -> Optional[str]:
...
+        lower=sorted_indices[-1],
+    )
+
+
+def image_extraction(
+    doc: lxml.html.HtmlElement,
+    paragraph_selector: XPath,
+    image_selector: XPath = XPath("//figure//img"),
+    upper_boundary_selector: XPath = XPath("//main"),
+    lower_boundary_selector: Optional[XPath] = None,
```
addie9800, line 550:
```suggestion
    lower_boundary_selector: Optional[XPath] = XPath("//footer"),
```
@MaxDall I would suggest adding this, because I am encountering Bounds Extraction errors regularly for premium articles. Since they often don't have any paragraphs and if there is no lower_boundary_selector set, also no lower_boundary causing the Bounds Extraction to return 0

MaxDall: The only problem is that this would overwrite the last paragraph as a lower boundary. It could be all the same, could be really bad.

Having working default values from the master branch would mitigate this problem, we could also argue to not raise an exception here but instead return an empty list.

I think having `None` should be fine while working on this.

src/fundus/parser/utility.py
```diff
@@ -567,7 +567,7 @@ def image_extraction(
         "(./ancestor::figure//*[(contains(@class, 'copyright') or contains(@class, 'credit')) and text()])[1]"
     ),
     author_filter: Optional[Pattern[str]] = None,
-    relative_urls: bool = False,
+    relative_urls: Union[bool, XPath] = False,
```
addie9800, line 570: I suggest changing this slightly, to cover unusual cases. Refer to `People` for an examplory use-case.

src/fundus/parser/utility.py
```diff
@@ -637,21 +637,24 @@ def nodes_to_text(nodes: List[Union[lxml.html.HtmlElement, str]]) -> Optional[st
...
         # parse authors
         authors = []
         if isinstance(author_selector, Pattern):
             # author is part of the caption
             if caption and (match := re.search(author_selector, caption)):
                 authors = [match.group("credits")]
                 caption = re.sub(author_selector, "", caption).strip() or None
+            elif description and (match := re.search(author_selector, description)):
+                authors = [match.group("credits")]
+                description = re.sub(author_selector, "", description).strip() or None
```
MaxDall: I would suggest leaving description as is and not applying filters here. If I remember correctly we stated in the documentation, that its the parsed `alt` attribute of the image, so I would argue one would expect the raw data.

src/fundus/parser/utility.py
```diff
@@ -542,8 +542,14 @@ def get_versions_from_node(
...
-    width = float(source.get("width") or 0) or None
-    height = float(source.get("height") or 0) or None
+    try:
+        width = float(source.get("width") or 0) or None
+    except ValueError:
+        width = None
+    try:
+        height = float(source.get("height") or 0) or None
+    except ValueError:
+        height = None
```
MaxDall, line 552: Do you know which case leads to a `ValueError`? In general, I would like to avoid `try ... except` blocks, especially in highly frequented code.

MaxDall (changes requested):

MaxDall (approved):

PR #608, opened 2024-09-04
Add Swiss newspaper Tages Anzeiger

MaxDall:


src/fundus/publishers/ch/ta.py
```diff
@@ -0,0 +1,39 @@
...
+    generic_author_parsing,
+    generic_date_parsing,
+)
+
+
+class TAParser(ParserProxy):
+    class V1(BaseParser):
+        _summary_selector = CSSSelector("p.ContentHead_lead____SsS")
+        _subheadline_selector = CSSSelector("article > h2")
+        _paragraph_selector = CSSSelector("article > p:not(.ContentHead_lead____SsS):not(.Feedback_root__fr_Mi)")
```
addie9800, line 18: I would also filter out p elements with class `ArticleContainer_agencies__g6Lpj ArticleContainer_content-width__FRl7F`. They contain the contributing agencies at the end of an article, like `SDA/ij` in [this article](https://www.tagesanzeiger.ch/sunrise-telekomkonzern-will-bis-ende-jahr-an-boerse-948884565986)

addie9800 (changes requested):
Thanks for adding this new publisher. I only got a minor comment :)

addie9800 (approved):

PR #609, opened 2024-09-10
Patch `RobotsFileParser` and add crawl-delay

MaxDall:
This PR ...
- patches the stdlib `RobotsFileParser` to properly fetch robots.txt files again
- removes `robotspy `dependency
- add crawl-delay parsing

closes #601


addie9800 (approved):
Thanks for updating this. This way I guess is nicer as well, since we don't have the additional dependency

PR #611, opened 2024-09-11
Fix Sportschau

addie9800:
The former summary selector was too broad, causing errors when paring articles like this one: https://www.sportschau.de/fussball/nationsleague/ein-wechselspiel-aus-licht-und-schatten,einzelkritik-dfb-102.html . The parsing crashed because the summary nodes are not restricted to the beginning of the article.

### DIFF
--- a/src/fundus/publishers/de/sportschau.py
+++ b/src/fundus/publishers/de/sportschau.py
@@ -14,7 +14,9 @@ from fundus.parser.utility import (
 
 class SportSchauParser(ParserProxy):
     class V1(BaseParser):
-        _summary_selector = CSSSelector("p >strong")
+        _summary_selector = CSSSelector(
+            "p[class='textabsatz columns twelve  m-ten  m-offset-one l-eight l-offset-two']" " > strong"
+        )
         _paragraph_selector = CSSSelector("article >p.textabsatz:not(p.textabsatz:nth-of-type(1))")
         _subheadline_selector = CSSSelector("article >h2")
 


src/fundus/publishers/de/sportschau.py
```diff
@@ -14,7 +14,9 @@
 
 class SportSchauParser(ParserProxy):
     class V1(BaseParser):
-        _summary_selector = CSSSelector("p >strong")
+        _summary_selector = CSSSelector(
+            "p[class='textabsatz columns twelve  m-ten  m-offset-one l-eight l-offset-two']" " > strong"
+        )
         _paragraph_selector = CSSSelector("article >p.textabsatz:not(p.textabsatz:nth-of-type(1))")
```
MaxDall, line 20: I would suggest these two selectors as it seems the summary is always the first paragraph:

```python
_paragraph_selector = XPath("//article/p[position() > 1]")
_summary_selector = XPath("//article/p[1]")
```

MaxDall (changes requested):

src/fundus/publishers/de/sportschau.py
addie9800: I think that unfortunately that does not always hold, since they also include a wide variety of other sources. E.g. this article from RBB does not include any summary: https://www.sportschau.de/regional/rbb/rbb-fussball-ex-nationalspieler-nico-schulz-spielt-ab-sofort-bei-tuerkischem-zweitligisten-100.html

MaxDall (approved):

PR #612, opened 2024-09-13
Fix a bug with the `plaintext` property of `Article`

MaxDall:
There is a bug in the evaluation of the `plaintext` property of `Article` so that the actual content of `ArticleBody` is not displayed unless it includes a paragraph.


addie9800 (approved):
Thanks 👍

PR #613, opened 2024-09-13
Add `HTML` headers to `Robots.read()`

MaxDall:
As some publishers require `user-agent` set for request headers, this PR add the `HTML` request header to the `Robots` request.


addie9800 (approved):
Good catch, that explains why doe of those sites weren't working.

PR #614, opened 2024-09-16
Implement XPath queries for LDs

MaxDall:
This PR enables developers to query a `LinkedDataMapping` using XPath expressions by adding a method called `xpath_search` to `LinkedDataMapping`.

For an example, please take a look at the [test cases](https://github.com/flairNLP/fundus/pull/614/files#diff-60fb0a496fe1cd67d38a04f3cfde59201fe9bff091439e2d07eb834e1ba62f3a).


addie9800 (dismissed):
This is a nice addition. Perhaps we should consider starting a new doc file with the utility functions, so that one can easily have an overview of what's available

PR #615, opened 2024-09-16
Add The West Australian

MaxDall:
This PR is based on #614 and should be reviewed afterward.


src/fundus/publishers/au/__init__.py
```diff
@@ -15,3 +16,13 @@ class AU(metaclass=PublisherGroup):
...
         ],
     )
+
+    WestAustralian = Publisher(
+        name="The West Australian",
+        domain="https://thewest.com.au/",
+        parser=WestAustralianParser,
+        sources=[
+            NewsMap("https://thewest.com.au/news-sitemap.xml"),
+            Sitemap("https://thewest.com.au/sitemap.xml", reverse=True),
```
addie9800, line 26: There's also a RSS feed containing the most recent articles: https://thewest.com.au/rss

src/fundus/publishers/au/west_australian.py
```diff
@@ -0,0 +1,67 @@
...
+        def publishing_date(self) -> Optional[datetime.datetime]:
+            return generic_date_parsing(self.precomputed.ld.bf_search("datePublished"))
+
+        @attribute
+        def authors(self) -> List[str]:
+            return generic_author_parsing(self.precomputed.ld.bf_search("author"))
+
+        @attribute
+        def topics(self) -> List[str]:
+            return generic_topic_parsing(self.precomputed.ld.bf_search("keywords"))
```
addie9800, line 67: For some reason TheWestAustralian seems to include a lot of duplicate topics. I think we should filter them to be unique (maybe it makes sense to implement it in `generic_topic_parsing`?). This is already visible in the testcase or in [This](https://thewest.com.au/news/wa/tanya-plibersek-has-half-heartedly-supported-nigel-satterleys-contentious-north-stoneville-townsite-c-16087647) article. Also, they add the topic `premium` to their plus articles, I would consider filtering that as well

src/fundus/publishers/au/west_australian.py
```diff
@@ -0,0 +1,67 @@
...
+)
+
+
+class WestAustralianParser(ParserProxy):
+    class V1(BaseParser):
+        _page_data_selector = XPath(
+            "string(//script[re:test(text(), 'window.PAGE_DATA')])",
+            namespaces={"re": "http://exslt.org/regular-expressions"},
+        )
+        _json_type_pattern = re.compile(r"^(?P<type>.*)=\s*{")
```
addie9800, line 24: This seems to be leftover. At least I couldn't find any usage in the project files.

addie9800 (changes requested):
Thanks for adding this publisher, seems like it was quite some effort. I only got some minor suggestions :)

src/fundus/publishers/au/west_australian.py
MaxDall: Good Point! I will add this in a separate PR

addie9800 (approved):

PR #618, opened 2024-09-24
Remove `<br>` tags in html

addie9800:
Occasionally, a `<br>` tag is used to start a new paragraph instead of the usual new `<p>` element. I have modified the `_base_setup` to replace these tags prior to processing the article. As an extreme example consider this article: https://www.namibian.com.na/namibian-disability-federation-urges-official-recognition-of-sign-language-for-deaf-inclusion/
The test case also contains one example.

### DIFF
--- a/src/fundus/publishers/na/the_namibian.py
+++ b/src/fundus/publishers/na/the_namibian.py
@@ -51,7 +51,12 @@ class TheNamibianParser(ParserProxy):
         _paragraph_selector = XPath("//div[contains(@class, 'entry-content')]/p[(text() or strong) and position()>1]")
         _summary_selector = XPath("//div[contains(@class, 'entry-content')]/p[(text() or strong) and position()=1]")
 
-        def _base_setup(self, html: str) -> None:
-            html = re.sub(r"(<br>)+", "<p>", html)
+        @attribute
+        def body(self) -> ArticleBody:
+            html = re.sub(r"(<br>)+", "<p>", self.precomputed.html)
             doc = lxml.html.document_fromstring(html)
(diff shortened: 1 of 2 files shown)


src/fundus/publishers/na/the_namibian.py
```diff
@@ -44,5 +48,10 @@ def authors(self) -> List[str]:
...
         VALID_UNTIL = datetime.today().date()
-        _paragraph_selector = XPath("//div[contains(@class, 'entry-content')]/p[position()>1]")
-        _summary_selector = XPath("//div[contains(@class, 'entry-content')]/p[position()=1]")
+        _paragraph_selector = XPath("//div[contains(@class, 'entry-content')]/p[(text() or strong) and position()>1]")
+        _summary_selector = XPath("//div[contains(@class, 'entry-content')]/p[(text() or strong) and position()=1]")
+
+        def _base_setup(self, html: str) -> None:
+            html = re.sub(r"(<br>)+", "<p>", html)
+            doc = lxml.html.document_fromstring(html)
+            self.precomputed = Precomputed(html, doc, get_meta_content(doc), get_ld_content(doc))
```
MaxDall, line 57: Two things here:
1. The intended way to do something like this would be to use the `function` decorator instead of overwriting the `_base_setup`
2. `html` should still reflect the original HTML. We have to find some way around this.

MaxDall (changes requested):

src/fundus/publishers/na/the_namibian.py
addie9800: I changed it to only modify the html for the purpose of extracting the body. This way it's also not necessary to overwrite the HTML.

MaxDall (approved):

PR #620, opened 2024-09-30
Adjust `generic_topic_parsing` to return only unique topics

MaxDall:


addie9800 (approved):
Thanks for adressing this :+1:

PR #623, opened 2024-10-02
Fix `Focus`

addie9800:
This PR adresses two issues:
- Some articles such as this one: https://www.focus.de/finanzen/karriere/gen-z-in-den-usa-zieht-handwerkliche-berufe-hochbezahlten-buerojobs-vor_eaf8f700-90ab-45ec-a1f3-8d1ac74faa86.html are not extracted at all, due to the layout being slightly different
- QSE is having trouble using the focus articles, due to topics being rarely used. I implemented a rough fallback, to use section keywords, what in my opinion is somewhat satisfactory, but perhaps @MaxDall you can also check, if it's worth adding them.

### DIFF
--- a/src/fundus/publishers/de/focus.py
+++ b/src/fundus/publishers/de/focus.py
@@ -19,7 +19,7 @@ class FocusParser(ParserProxy):
         _summary_selector = CSSSelector("div.leadIn > p, div.Article-Description ")
         _subheadline_selector = CSSSelector("div.textBlock > h2, div[data-qa-article-content-text] > h2")
         _snippet_selector = XPath(
-            'string(//script[@type="text/javascript"][contains(text(), "window.bf__bfa_metadata") or contains(text(), "pageAdKeyword")])'
+            'string(//script[@type="text/javascript"][contains(text(), "window.bf__bfa_metadata")])'
         )
 
         # regex patterns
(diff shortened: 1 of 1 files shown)


MaxDall:
@addie9800 Awesome that you caught these layout edge cases!

Regarding the topic part: I wouldn't consider section keywords as useful topics especially since it introduces inconsistency to the actual topic field either for this particular publisher or all publishers as a whole. I would suggest leaving this to downstream tasks like QSE.

src/fundus/publishers/de/focus.py
```diff
@@ -15,11 +15,11 @@
...
     class V1(BaseParser):
-        _paragraph_selector = CSSSelector("div.textBlock > p")
-        _summary_selector = CSSSelector("div.leadIn > p")
-        _subheadline_selector = CSSSelector("div.textBlock > h2")
+        _paragraph_selector = CSSSelector("div.textBlock > p , div[data-qa-article-content-text] > p")
+        _summary_selector = CSSSelector("div.leadIn > p, div.Article-Description ")
+        _subheadline_selector = CSSSelector("div.textBlock > h2, div[data-qa-article-content-text] > h2")
         _snippet_selector = XPath(
-            'string(//script[@type="text/javascript"][contains(text(), "window.bf__bfa_metadata")])'
+            'string(//script[@type="text/javascript"][contains(text(), "window.bf__bfa_metadata") or contains(text(), "pageAdKeyword")])'
```
MaxDall, line 22: Is this still part of the new layout or a leftover of the former topic parsing fallback?

addie9800: No, that was also a leftover. It's fixed now

MaxDall (approved):

PR #625, opened 2024-10-03
Cleaner code for LD and JSON parsing

MaxDall:
This PR aims to make the code about LD and JSON parsing a bit less complex. Further, I removed some unused imports for `TheNamibian`

### DIFF
--- a/src/fundus/parser/utility.py
+++ b/src/fundus/parser/utility.py
@@ -7,7 +7,6 @@ from dataclasses import dataclass, field
 from datetime import datetime
 from functools import total_ordering
 from typing import (
-    Any,
     Callable,
     ClassVar,
     Dict,
@@ -18,6 +17,7 @@ from typing import (
     Pattern,
     Type,
     Union,
+    cast,
 )
 
 import lxml.html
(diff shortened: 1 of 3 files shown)


src/fundus/parser/utility.py
```diff
@@ -154,25 +155,34 @@ def extract_nodes(selector: XPath, node_type: Type[Node]) -> List[Node]:
...
-        json_content = sanitize_json(node.text_content()) or ""
-        try:
-            jsons.append(json.loads(json_content))
-        except json.JSONDecodeError as error:
-            logger.debug(f"Encountered {error!r} during JSON parsing")
+    if match is None or not (sanitized := match.group("json")):
+        return None
+
+    # substitute "bad" values
+    sanitized = re.sub(_json_undefined, r"\g<key>:null", sanitized)
```
addie9800, line 168: This is possible? 🤯 You don't want to know how often I worked around that to avoid selecting text I don't want to loose due to substitution...

addie9800 (approved):
Nice modification. It makes a lot more sense to move that to utility

src/fundus/parser/utility.py
MaxDall: I feel you 😄 regex is a rabbit hole and every time I have to use it I find myself half an hour and 30 open tabs later scratching my head.

PR #626, opened 2024-10-03
Deprecate `get_value_by_key_path` and replace with `xpath_search`

MaxDall:
This PR deprecates `get_value_by_key_path` class method of `LinkedDataMapping` and replaces all existing occasions with `xpath_search`.

To talk about efficiency I provide some additional runtime comparison and a script to reproduce.
The script loads one article for `The West Australian`, since it by far includes the largest `LD` and loads a specific value (or key path) 100000 times. For more details click on `Script`.

<details>
  <summary>Script</summary>
  
  ```python
from timeit import timeit

setup = '''
from timeit import timeit

from lxml.etree import XPath

test_xpath = XPath("NewsArticle/datePublished")

crawler = Crawler(PublisherCollection.au.WestAustralian)

for article in crawler.crawl(max_articles=1):
    ld = article.ld
'''

stmt1 = 'ld.get_value_by_key_path(["NewsArticle", "datePublished"])'
stmt2 = 'ld.xpath_search("NewsArticle/datePublished")'
stmt3 = 'ld.xpath_search(test_xpath)'

length = 100000

time1 = timeit(setup=setup, stmt=stmt1, number=length)
time2 = timeit(setup=setup, stmt=stmt2, number=length)
time3 = timeit(setup=setup, stmt=stmt3, number=length)

print(f"Loading a value {length} times")
print(f"key_path took {time1:.2f} seconds, with an avg of {time1/length:.2e}")
print(f"xpath_search without pre-compiled XPath took {time2:.2f} seconds, with an avg. of {time2/length:.2e}")
print(f"xpath_search with pre-compiled XPath took {time3:.2f} seconds, with an avg. of {time3/length:.2e}")
print(f"using pre compiled xpath is {time2/time3:.2f} times faster")
print(f"using key_path is {time3/time1:.2f} times faster than xpath_search")
  ```
</details>

### Output
```
Loading a value 100000 times
key_path took 0.07 seconds, with an avg of 6.98e-07
xpath_search without pre-compiled XPath took 2.42 seconds, with an avg. of 2.42e-05
xpath_search with pre-compiled XPath took 1.51 seconds, with an avg. of 1.51e-05
using pre-compiled XPath is 1.60 times faster
using key_path is 21.65 times faster than xpath_search
```

For better typing, this PR adds a `scalar` parameter to `xpath_search`, indicating that one expects an optional scalar value in return.

### DIFF
--- a/src/fundus/parser/data.py
+++ b/src/fundus/parser/data.py
@@ -120,7 +120,7 @@ class LinkedDataMapping:
         ...
 
     @overload
-    def xpath_search(self, query: Union[XPath, str], scalar: Literal[True] = True) -> Any:
+    def xpath_search(self, query: Union[XPath, str], scalar: Literal[True] = True) -> Optional[Any]:
         ...
 
     def xpath_search(self, query: Union[XPath, str], scalar: bool = False) -> Union[Any, List[Any]]:
@@ -190,10 +190,12 @@ class LinkedDataMapping:
         values = list(results.values())
 
         if scalar:
-            if len(values) != 1:
-                raise ValueError(f"Got multiple values when expecting a single scalar value")
-            else:
+            if not values:
+                return None
+            elif len(values) == 1:
                 return values.pop()
+            else:
+                raise ValueError(f"Got multiple values when expecting a single scalar value")
         else:
             return values
 
--- a/src/fundus/parser/data.py
+++ b/src/fundus/parser/data.py
@@ -22,7 +22,7 @@ import more_itertools
 import xmltodict
 from dict2xml import dict2xml
 from lxml.etree import XPath, tostring
-from typing_extensions import Self, TypeAlias
+from typing_extensions import Self, TypeAlias, deprecated
 
 from fundus.utils.serialization import replace_keys_in_nested_dict
 
@@ -81,6 +81,7 @@ class LinkedDataMapping:
                 self.__dict__[self.__UNKNOWN_TYPE__] = []
             self.__dict__[self.__UNKNOWN_TYPE__].append(ld)
 
+    @deprecated("Use xpath_search() instead")
     def get_value_by_key_path(self, key_path: List[str], default: Any = None) -> Optional[Any]:
         """
         Works like get() except this one assumes a path is given as list of keys (str).
--- a/src/fundus/parser/data.py
+++ b/src/fundus/parser/data.py
@@ -123,7 +123,7 @@ class LinkedDataMapping:
     def xpath_search(self, query: Union[XPath, str], scalar: Literal[True] = True) -> Optional[Any]:
         ...
 
-    def xpath_search(self, query: Union[XPath, str], scalar: bool = False) -> Union[Any, List[Any]]:
+    def xpath_search(self, query: Union[XPath, str], scalar: bool = False):
         """Search through LD using XPath expressions
 
         Internally, the content of the LinkedDataMapping is converted to XML and then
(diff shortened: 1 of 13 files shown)


src/fundus/parser/data.py
```diff
@@ -113,7 +115,15 @@ def to_unicode_characters(text: str) -> str:
...
+        ...
+
+    @overload
+    def xpath_search(self, query: Union[XPath, str], scalar: Literal[True] = True) -> Optional[Any]:
+        ...
+
+    def xpath_search(self, query: Union[XPath, str], scalar: bool = False) -> Union[Any, List[Any]]:
         """Search through LD using XPath expressions
 
         Internally, the content of the LinkedDataMapping is converted to XML and then
```
addie9800, line 129: I would also add some documentation for `scalar`

src/fundus/parser/data.py
```diff
@@ -113,7 +115,15 @@ def to_unicode_characters(text: str) -> str:
...
-    def xpath_search(self, query: XPath) -> List[Any]:
+    @overload
+    def xpath_search(self, query: Union[XPath, str], scalar: Literal[False] = False) -> List[Any]:
+        ...
+
+    @overload
+    def xpath_search(self, query: Union[XPath, str], scalar: Literal[True] = True) -> Optional[Any]:
+        ...
+
+    def xpath_search(self, query: Union[XPath, str], scalar: bool = False) -> Union[Any, List[Any]]:
```
addie9800, line 126: Not being that familiar with typing, is this third variation necessary for the case where a variable value is passed to scalar? So it's undetermined at compilation, whether it will be True or False?

Also, if this is the case, shouldn't the return value also be `Union[Optional[Any], List[Any]]`, to be consistent with the case of Literal[True]?

src/fundus/publishers/shared/euronews.py
```diff
@@ -28,8 +28,7 @@ def body(self) -> ArticleBody:
 
         @attribute
         def authors(self) -> List[str]:
-            key_path = ["NewsArticle", "author", "name"]
-            author_string = self.precomputed.ld.get_value_by_key_path(key_path)
+            author_string = self.precomputed.ld.xpath_search("NewsArticle/author/name", scalar=True)
```
addie9800, line 31: Is the `scalar=True` on purpose here? Shouldn't multiple author elements be possible as well?

src/fundus/publishers/us/occupy_democrats.py
```diff
@@ -41,7 +41,7 @@ def title(self) -> Optional[str]:
 
         @attribute
         def topics(self) -> List[str]:
-            return generic_topic_parsing(self.precomputed.ld.get_value_by_key_path(["Article", "keywords"]))
+            return generic_topic_parsing(self.precomputed.ld.xpath_search("Article/keywords", scalar=True))
```
addie9800, line 44: Also here, is the `scalar=True` intentional?

addie9800 (changes requested):
Looks like a good change. Thanks 👍

src/fundus/parser/data.py
MaxDall: No, you don't have to type hint the original function, it's just for IDEs like Pycharm.

You're right, I forgot the `Optional` there.

src/fundus/publishers/shared/euronews.py
MaxDall: I just mimicked the previous behavior bit your right, this should also be possible to take multiple authors into account.

src/fundus/publishers/us/occupy_democrats.py
MaxDall: Yes. Here we're searching for just one field `Article/keywords`. The field can still be a list. Scalar doesn't make any assumption about the type of the returned result, only about the number of results. With `scalar` we basically say that we expect only one field to match the expression.

addie9800 (approved):
Looks good 👍

PR #627, opened 2024-10-13
Add `The Japan News`

addie9800:


MaxDall (changes requested):
Thanks a lot for adding 👍 Maybe we should specify the language a news paper uses at one point, what do you think?

MaxDall (approved):

PR #629, opened 2024-10-14
Update lint.yml

addie9800:
Apparently there is an open issue with the isort-action not working properly with the latest ubuntu release: https://github.com/isort/isort-action/issues/94 This change should fix this until the library is patched.


addie9800:
The issue was fixed

PR #631, opened 2024-10-15
Handle multiple ld types

addie9800:
As of now, the parsing of LD-JSON raises an error, when encountering an object with multiple types defined. I would suggest that, instead of prohibiting multiple values for ld_type, we combine them in a comma separated list. This (or another solution) is necessary for adding the Spanish Newspaper El Pais


MaxDall (approved):
Thanks 👍

PR #632, opened 2024-10-15
Add `El Pais`

addie9800:


src/fundus/publishers/es/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+from fundus.scraping.url import RSSFeed
+
+
+class ES(metaclass=PublisherGroup):
+    ElPais = Publisher(
+        name="El País",
+        domain="https://elpais.com/",
+        parser=ElPaisParser,
+        sources=[
+            RSSFeed("https://feeds.elpais.com/mrss-s/pages/ep/site/elpais.com/portada"),
```
MaxDall, line 12: It seems like you can request sitemaps with this schema `https://elpais.com/sitemaps/2015/10/sitemap_0.xml`. Everything greater than `0` seems to result in the same sitemap.

MaxDall (changes requested):
Really nice as well 🎉 First Spanish publisher

src/fundus/publishers/es/__init__.py
addie9800: Nice, how did you find that?

MaxDall: Often it helps googeling the site + sitemap, in this case `www.elpais.com sitemap`

MaxDall (approved):

PR #633, opened 2024-10-15
Add default return values for attributes

MaxDall:
There was a long-lasting bug when using `suppress` as error handling, disabling its functionality. Because of the lack of default values for `Attribute`s, whenever an error occurred during the parsing of a specific attribute the exception was passed to the scraper, skipping the entire article.

This PR adds a property `__default__` to the attribute. The property returns a default value for the specific attribute based on its annotation, thus enabling us to use default values in case of an exception.

Further, I changed the type of `body` in the guidelines to `Optional[ArticleBody]` to be more in tune with `Article`.

The only review-relevant file here is `base_parser.py`


addie9800 (approved):
Thanks for fixing this 👍

PR #650, opened 2024-10-24
Improve encoding detection

MaxDall:
This adds an additional selector to the HTML encoding detection to avoid requests build-in detection. This is because the latter uses chardet which can lead to seriously long execution times (>3s) for a single article.


addie9800 (approved):
Sounds like a great speed-up 👍

PR #652, opened 2024-10-28
Fix `Frankfurter Rundschau`

addie9800:
This PR adds parsing for this article: https://www.fr.de/ratgeber/tiere/zeitumstellung-tiere-haustiere-hunde-katzen-wildtier-landwirtschaft-tagesrhythmus-zr-93379931.html

### DIFF
--- a/src/fundus/publishers/de/frankfurter_rundschau.py
+++ b/src/fundus/publishers/de/frankfurter_rundschau.py
@@ -14,11 +14,14 @@ from fundus.parser.utility import (
 
 class FrankfurterRundschauParser(ParserProxy):
     class V1(BaseParser):
-        _paragraph_selector = XPath("//p[@class='id-StoryElement-paragraph'] | "
-              "//p[contains(@class,'id-Article-content-item-paragraph') and text()] |"
-              "//div[@class='id-Article-body']//ul/li[not(@class='id-AuthorList-item ')]")
-        _summary_selector = CSSSelector("p.id-StoryElement-leadText, "
-                                        "p[class='id-Article-content-item id-Article-content-item-summary']")
+        _paragraph_selector = XPath(
+            "//p[@class='id-StoryElement-paragraph'] | "
+            "//p[contains(@class,'id-Article-content-item-paragraph') and text()] |"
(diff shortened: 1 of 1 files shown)


src/fundus/publishers/de/frankfurter_rundschau.py
```diff
@@ -13,9 +14,15 @@
 
 class FrankfurterRundschauParser(ParserProxy):
     class V1(BaseParser):
-        _paragraph_selector = CSSSelector("p.id-StoryElement-paragraph, article > ul > li")
-        _summary_selector = CSSSelector("p.id-StoryElement-leadText")
-        _subheadline_selector = CSSSelector("h2.id-StoryElement-crosshead")
+        _paragraph_selector = XPath(
+            "//p[@class='id-StoryElement-paragraph'] | "
+            "//p[contains(@class,'id-Article-content-item-paragraph') and text()] |"
+            "//div[@class='id-Article-body']//ul/li[not(@class='id-AuthorList-item ')]"
```
MaxDall, line 20: This one differs from the original `article > ul > li` is there a reason for that?

src/fundus/publishers/de/frankfurter_rundschau.py
```diff
@@ -13,9 +14,15 @@
...
+        )
+        _summary_selector = CSSSelector(
+            "p.id-StoryElement-leadText, " "p[class='id-Article-content-item id-Article-content-item-summary']"
```
MaxDall, line 23: There are unnecessary quotation marks

MaxDall (changes requested):

src/fundus/publishers/de/frankfurter_rundschau.py
addie9800: Yes, they added another list item in some articles called "Meine News". You can see an example here: https://www.fr.de/politik/ampel-regierung-koalition-news-lindner-fdp-habeck-gruene-spd-wirtschaft-plan-papier-zr-93394586.html

src/fundus/publishers/de/frankfurter_rundschau.py
addie9800: Thanks, seems to be left-over from formatting.

MaxDall (approved):

PR #653, opened 2024-10-29
Minor Changes to Utility Functions

addie9800:
I suggest the following changes:
- sort by float value (I had a case using 1.5x)
- use the ancestor function of XPath to get the parent picture element, should it exist. This has the advantage that if it exist, it will actually be found. In the previous version, `.getParent()` would select the predecessor, which is not necessarily the parent, as one would expect it to be (thinking of a tree structure).

_Only the utility.py should be worth going through. If you look at the commit with the changes to utility, you will see the difference the second bullet point causes._

### DIFF
--- a/docs/supported_publishers.md
+++ b/docs/supported_publishers.md
@@ -526,9 +526,7 @@
           <span>www.golem.de</span>
         </a>
       </td>
-      <td>
-        <code>images</code>
-      </td>
+      <td>&#160;</td>
       <td>&#160;</td>
     </tr>
     <tr>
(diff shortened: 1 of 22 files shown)


MaxDall:
> sort by float value (I had a case using 1.5x)

Totally support this!

> In the previous version, .getParent() would select the predecessor

Where did you get that info from, in the lxml [doku](https://lxml.de/api/lxml.etree._Element-class.html) it states that `getparent` returns the parent element or None if the node is `root`

PR #655, opened 2024-10-30
Fix pickling problem in `LinkedDataMapping`

addie9800:
As of now CC-Crawling is crashing due to a pickling error, with a message along the lines of `cannot pickle lxml.etree._Element`. I traced the error back to this article: https://de.euronews.com/2023/02/17/chinas-botschaft-in-paris-twittert-falsche-behauptungen-uber-das-erdbeben-in-der-turkei where the issue seems to be the use of the `__as_xml__()` function, which creates an attribute in the `LinkedDataMapping` object of type `lxml.etree._Element`. To avoid crashing, I suggest overwriting the `__getstate__()` and `__setstate__()` functions converting the Elements to a string before pickling.

### DIFF
--- a/src/fundus/parser/data.py
+++ b/src/fundus/parser/data.py
@@ -22,7 +22,7 @@ import lxml.etree
 import more_itertools
 import xmltodict
 from dict2xml import dict2xml
-from lxml.etree import XPath, tostring
+from lxml.etree import XPath, fromstring, tostring
 from typing_extensions import Self, TypeAlias, deprecated
 
 from fundus.utils.serialization import JSONVal, replace_keys_in_nested_dict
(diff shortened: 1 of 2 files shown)


MaxDall:
@addie9800 Oh wow, thanks a lot for catching this. That's actually a rather big thing, since it effects all parser who make use of the now officially supported `xpath_search` function.

I'm not quite sure about the road I want to go here, but that aside, your suggestion looks like a perfect solution if we're going to keep the XML persistent to save some time.

MaxDall (changes requested):
@addie9800 Thanks a lot for fixing this! I completely overlooked that. That's a huge miss on my side.

MaxDall (approved):

PR #660, opened 2024-11-05
Add `octet-stream` to decompressor

MaxDall:
Adds a new content-type to the decompressor to decompress `octet-streams`. Previously sitemaps like for `lt.LRT` couldn't be read.

### DIFF
--- a/src/fundus/scraping/url.py
+++ b/src/fundus/scraping/url.py
@@ -2,6 +2,7 @@ import gzip
 import itertools
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
+from enum import Enum
 from functools import cached_property
 from typing import Callable, ClassVar, Dict, Iterable, Iterator, List, Optional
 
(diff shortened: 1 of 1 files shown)


src/fundus/scraping/url.py
```diff
@@ -13,19 +14,51 @@
...
 
 from fundus.logging import create_logger
+from fundus.parser.utility import generic_nodes_to_text
 from fundus.scraping.filter import URLFilter, inverse
 from fundus.scraping.session import _default_header, session_handler
 
 logger = create_logger(__name__)
 
 
+class CompressionFormats(Enum):
```
addie9800, line 24: From what I see, `bz2`, `zip` and `lzma` would raise a `NotImplementedError`. Without having any background knowledge it seems odd to me that you specifically chose these four compression formats and only implemented the decompression for one of them.

MaxDall: I asked ChatGPT about the most common compression formats and added them as templates.

Update: I added the missing formats (except ZIP) as I have no clue how to handle ZIP archives unless we stumble upon one.

addie9800 (approved):
This looks good. Thanks for adding this 👍

PR #661, opened 2024-11-13
add version parsing

MaxDall:
This PR reworks the versioning system of images, adds extended serialization and parses size for image versions.

### DIFF
--- a/src/fundus/parser/utility.py
+++ b/src/fundus/parser/utility.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import itertools
 import json
 import re
@@ -19,6 +21,7 @@ from typing import (
     Optional,
     Pattern,
     Sequence,
+    Set,
     Type,
     Union,
 )
@@ -36,10 +39,13 @@ from fundus.parser.data import (
     DOM,
     ArticleBody,
     ArticleSection,
+    Dimension,
     Image,
+    ImageVersion,
     LinkedDataMapping,
     TextSequence,
 )
+from fundus.utils.regex import _get_match_dict
 
 logger = create_logger(__name__)
 
@@ -417,20 +423,8 @@ def image_author_parsing(authors: Union[str, List[str]], author_filter: Optional
     return generic_author_parsing(authors)
 
 
-class Bounds(NamedTuple):
-    upper: int
-    first_paragraph: Optional[int]
-    lower: int
-
-
-class IndexedImageNode(NamedTuple):
-    position: int
-    content: lxml.html.HtmlElement
-    is_cover: bool
-
-
 # https://regex101.com/r/MplUXL/2
-_srcset_pattern = re.compile(r"(?P<url>[^\s]+)\s*(?P<descriptor>[0-9.]+[wx])?(,?\s*)")
+_srcset_pattern = re.compile(r"(?P<url>\S+)\s*(?P<descriptor>[0-9.]+[wx])?(,?\s*)")
 
 
 def parse_srcset(srcset: str) -> Dict[str, str]:
@@ -444,7 +438,7 @@ def parse_srcset(srcset: str) -> Dict[str, str]:
     return dict(sorted(urls.items(), key=lambda item: float(item[0][:-1])))
 
 
-def parse_urls_from_image(node: lxml.html.HtmlElement) -> Optional[Dict[str, str]]:
+def parse_urls(node: lxml.html.HtmlElement) -> Optional[Dict[str, str]]:
     if srcset := (node.get("data-srcset") or node.get("srcset")):
         return parse_srcset(srcset)
     elif src := (node.get("data-src") or node.get("src")):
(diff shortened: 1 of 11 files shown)


lattaai9:
Hello, I tried to solve the issue.

This is what I did:

This PR reworks the versioning system of images, adds extended serialization and parses size for image versions in the Article class. The changes include:

1. Adding an 'images' property to store image versions
2. Implementing image version parsing with size information
3. Extending the to_json method to include image serialization
4. Adding a new method parse_image_versions to handle image versioning

You can review changes in this commit: https://github.com/lattaai9/flairNLP-fundus-661/commit/b662984e1fb32cac10644cb0b79c48736231b1a2.


> [!CAUTION]
> Disclaimer: The commit was created by Latta AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---
This issue was tried to solve for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

src/fundus/parser/utility.py
```diff
@@ -453,7 +447,95 @@ def parse_urls_from_image(node: lxml.html.HtmlElement) -> Optional[Dict[str, str
...
+    ) -> Optional[Dimension]:
+        if not (width or height):
+            width = self.width
+            height = self.height
+        if dimension := Dimension.from_ratio(width, height, self.ratio):
+            return dimension * (dpr or 1)
+        return None
+
+
+_min_width_pattern = re.compile(r"min-width:\s*(?P<minwidth>[0-9]*)(?P<descriptor>[pxem]+)")
```
addie9800, line 469: According to some css docs, possible units are also `%` and `ch`

src/fundus/parser/utility.py
```diff
@@ -453,7 +447,95 @@ def parse_urls_from_image(node: lxml.html.HtmlElement) -> Optional[Dict[str, str
...
+_width_x_height_pattern = re.compile(r"(?P<width>[0-9]+)x(?P<height>[0-9]+)")
```
addie9800, line 470: It might make sense to also match `max-width`, which is sometimes used instead of `min-width` for smaller images.
e.g. https://www.zeit.de/gesundheit/2023-11/neues-alzheimer-medikament-zulassung-lecanemab-leqembi-demenz

src/fundus/parser/utility.py
```diff
@@ -453,7 +447,95 @@ def parse_urls_from_image(node: lxml.html.HtmlElement) -> Optional[Dict[str, str
...
+        return set()
+
+    # get min width
+    min_width = None
+    if media := source.get("media"):
+        for value, descriptor in re.findall(_min_width_pattern, media):
+            if descriptor != "px":
+                logger.debug(f"Pixel calculation not implemented for {descriptor}")
+            else:
+                min_width = int(value)
```
addie9800, line 486: I feel like an iteration over all matches is not the best solution here. Granted, we would only expect one match, but in case there are more, overwriting all previous matches seems not ideal.

src/fundus/parser/utility.py
```diff
@@ -453,7 +447,95 @@ def parse_urls_from_image(node: lxml.html.HtmlElement) -> Optional[Dict[str, str
...
+        ratio = width / height
+    calculator = _DimensionCalculator(width, height, ratio)
+
+    versions = set()
+    for descriptor, url in urls.items():
+        kwargs: Dict[str, float] = {}
+        if descriptor is not None:
+            if match := re.search(r"(?P<multiplier>[0-9.]+)x", descriptor):
+                kwargs["dpr"] = float(match.group("multiplier"))
+            elif match := re.search(r"(?P<width>[0-9]+)(px|w)", descriptor):
```
addie9800, line 501: Other technically allowed units of measurement are `%` and `em`, also `w` doesn't seem to be a valid unit (https://www.w3schools.com/cssref/css_units.php)

src/fundus/parser/utility.py
```diff
@@ -453,7 +447,95 @@ def parse_urls_from_image(node: lxml.html.HtmlElement) -> Optional[Dict[str, str
...
+            return dimension * (dpr or 1)
+        return None
+
+
+_min_width_pattern = re.compile(r"min-width:\s*(?P<minwidth>[0-9]*)(?P<descriptor>[pxem]+)")
+_width_x_height_pattern = re.compile(r"(?P<width>[0-9]+)x(?P<height>[0-9]+)")
+
+
+def get_versions_from_node(
+    source: lxml.html.HtmlElement, ratio: Optional[float], size_patter: Optional[Pattern[str]]
```
addie9800, line 474: Typo in `size_pattern`

src/fundus/utils/regex.py
```diff
@@ -0,0 +1,40 @@
...
+    pattern: Pattern[str], string: str, conversion: Optional[Callable[[str], _T]] = None, keep_none: bool = False
+) -> Dict[str, Union[str, _T, None]]:
+    matches = {}
+    for match in re.finditer(pattern, string):
+        match_dict = match.groupdict()
+        for key, value in match_dict.items():
+            if value is not None:
+                matches[key] = conversion(value) if conversion is not None else value
+            elif keep_none:
+                matches[key] = match_dict[key] or value
```
addie9800, line 39: Are `match_dict[key]` and `value` not equivalent?

src/fundus/parser/utility.py
MaxDall: The URLs coming are coming from the `srcset` attribute. Allowed descriptors are `x`, `w`, I also found some `px`

src/fundus/utils/regex.py
MaxDall: You`re right, that's a type and should be `match`

addie9800 (approved):
Looks good, thanks for adding this 👍

addie9800 (changes requested):
By fixing a merge conflict I simultaneously triggered the full test program and it turns out, there seems to be an issue with the serialization of the versions. As far as I can see by now the main problem is the deserialization of `ImageVersion`

MaxDall:
> By fixing a merge conflict I simultaneously triggered the full test program and it turns out, there seems to be an issue with the serialization of the versions. As far as I can see by now the main problem is the deserialization of `ImageVersion`

The test cases from the image branch are not compatible with this branch.

addie9800:
> > By fixing a merge conflict I simultaneously triggered the full test program and it turns out, there seems to be an issue with the serialization of the versions. As far as I can see by now the main problem is the deserialization of `ImageVersion`
> 
> The test cases from the image branch are not compatible with this branch.

I know, I had a look at it locally and exemplarily overwrote the data for `TheNamibian`. When loading the test data the versions attribute is loaded as a `List[dict]` instead of `List[ImageVersion]`, which then later causes an issue when trying to access `version.url` in data.py:414

MaxDall:
> I know, I had a look at it locally and exemplarily overwrote the data for `TheNamibian`. When loading the test data the versions attribute is loaded as a `List[dict]` instead of `List[ImageVersion]`, which then later causes an issue when trying to access `version.url` in data.py:414

Hmm something seems odd here. I just did what you described:
- overwritten the test case with `python -m scripts.generate_parser_test_files -p TheNamibian -oj`
- loaded the test case with the following code:
```python
from tests.utility import get_test_articles

for article in get_test_articles(PublisherCollection.na.TheNamibian):
    for image in article.images:
        print(image)
```
and everything seems to work just fine.

Could you provide an example where it breaks for you so I can reproduce the error?

addie9800:
The first step is just as you described. Then run this code:
```python
from tests.utility import load_test_case_data

data = load_test_case_data(PublisherCollection.na.TheNamibian)
```

MaxDall:
@addie9800 Fixed it

addie9800 (approved):
Looks good 👍

PR #663, opened 2024-11-26
Update `FreiePresse` 

addie9800:
This PR splits up the FreiePresse Parser into two separate parser versions. I established the dates using the wayback machine. Verifying changes using CCNewsCrawler might take a couple of minutes.


addie9800:
This is a good test article for the new version: https://www.freiepresse.de/sport/rodeln-deutsch-schweizer-doppelzimmer-julia-taubitz-und-natalie-maag-will-auch-im-neuen-mixed-wettbewerb-angreifen-artikel13611492?ref=rss

MaxDall:
@addie9800 Thanks for addressing this and especially for finding a date for the layout change. It seems like the image extractor for `V1` can't parse an image for the test case, although there seems to be an image included in the article.

MaxDall (approved):

PR #664, opened 2024-11-29
Merge Funke parsers

addie9800:
I noticed some bloat in `Hamburger Abendblatt`. Since `Hamburger Abendblatt`, `Braunschweiger Zeitung` and `Berliner Morgenpost` belong to the same organization (and using the same layout), I figured it would be more efficient to merge the three parsers into one single parser.


MaxDall (changes requested):
@addie9800 Thanks for the code clean-up. Makes everything easier to maintain 👍 
Just one thing, the shared Funke parser would be better off in the `de` directory, because it's only shared for German publishers.

addie9800:
Moved to #676 due to merge conflicts

PR #682, opened 2025-01-13
Add `AsahiShimbun`

MaxDall:


src/fundus/publishers/jp/asahi_shimbun.py
```diff
@@ -0,0 +1,53 @@
...
+        @attribute
+        def topics(self) -> List[str]:
+            return generic_topic_parsing(self.precomputed.meta.get("keywords"))
+
+        @attribute
+        def images(self) -> List[Image]:
+            return image_extraction(
+                doc=self.precomputed.doc,
+                paragraph_selector=self._paragraph_selector,
+                relative_urls=True,
```
addie9800, line 52: Sometimes, there is an author attribution. From my Google Translate understanding the regex should look something like this:
`r"、(?P<credits>.*?)撮影"`
As the photo by part of the attribution seems to be the last to characters following the actual name of the photographer.

addie9800: This is an example article: https://www.asahi.com/articles/AST1F4445T1FUTIL02SM.html

src/fundus/publishers/jp/asahi_shimbun.py
```diff
@@ -0,0 +1,53 @@
...
+        def publishing_date(self) -> Optional[datetime.datetime]:
+            return generic_date_parsing(self.precomputed.meta.get("article:published_time"))
+
+        @attribute
+        def authors(self) -> List[str]:
+            return generic_author_parsing(self.precomputed.ld.bf_search("author"))
+
+        @attribute
+        def topics(self) -> List[str]:
+            return generic_topic_parsing(self.precomputed.meta.get("keywords"))
```
addie9800, line 45: I would add a filter for the following words: `朝日新聞デジタル` (Asahi Shimbun Digital), `朝日新聞` (Asahi Shimbun) , `ニュース`(News) , `新聞` (Newspaper), `その他・話題` (Other/Topics)

addie9800 (changes requested):
Thanks for adding the next Japanese publisher 👍

addie9800 (approved):
Looks good 👍

PR #683, opened 2025-01-13
Add `ChunichiShimbun` and `TokyoShimbun`

MaxDall:


src/fundus/publishers/jp/tokyo_chunichi_shimbun.py
```diff
@@ -0,0 +1,59 @@
...
+        def body(self) -> Optional[ArticleBody]:
+            return extract_article_body_with_selector(
+                self.precomputed.doc,
+                paragraph_selector=self._paragraph_selector,
+                subheadline_selector=self._subheadline_selector,
+            )
+
+        @attribute
+        def title(self) -> Optional[str]:
+            return self.precomputed.meta.get("og:title")
```
addie9800, line 33: I would get the title from the JSON instead, since `：東京新聞デジタル` (Tokyo Shimbun Digital) is appended in the `og:title` attribute

src/fundus/publishers/jp/tokyo_chunichi_shimbun.py
```diff
@@ -0,0 +1,59 @@
...
+    generic_author_parsing,
+    generic_date_parsing,
+    generic_topic_parsing,
+    image_extraction,
+)
+
+
+class TokyoChunichiShimbunParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = XPath("//div[@class='block' and not(descendant::div or descendant::h2)]")
```
addie9800, line 20: [This article](https://www.tokyo-np.co.jp/article/379018) has a line with just `　　　◇`. I am not sure, if it should be filtered or not, since technically it's not really any content, but just some formatting.

src/fundus/publishers/jp/tokyo_chunichi_shimbun.py
```diff
@@ -0,0 +1,59 @@
...
+        @attribute
+        def images(self) -> List[Image]:
+            return image_extraction(
+                doc=self.precomputed.doc,
+                image_selector=CSSSelector("main div.image img, main div.thumb img"),
+                caption_selector=XPath(
+                    "./ancestor::div[@class='wrap']//p[@class='caption'] | "
+                    "./ancestor::div[@class='thumb']//p[@class='thumb-caption']"
+                ),
+                author_selector=re.compile(r"（(?P<credits>[^）]+)）\s*$"),
```
addie9800, line 56: The selector should be modified to not select `撮影` which roughly translated to photo by. Note that this is placed behind the credits to be selected.

src/fundus/publishers/jp/tokyo_chunichi_shimbun.py
addie9800: On that note, they seem to be using black diamonds as formatting: `◆`. Also here, I'm not sure what to do with that. I am 80% sure it plays no role in the semantics of the language. It can be found in the previous article as well as [here](https://www.tokyo-np.co.jp/article/379028)

src/fundus/publishers/jp/tokyo_chunichi_shimbun.py
```diff
@@ -0,0 +1,59 @@
...
+        def title(self) -> Optional[str]:
+            return self.precomputed.meta.get("og:title")
+
+        @attribute
+        def publishing_date(self) -> Optional[datetime.datetime]:
+            return generic_date_parsing(self.precomputed.ld.bf_search("datePublished"))
+
+        @attribute
+        def authors(self) -> List[str]:
+            return generic_author_parsing(self.precomputed.ld.bf_search("author"))
```
addie9800, line 41: I would remove the substring `記者`, which translates to `Reporter`

src/fundus/publishers/jp/tokyo_chunichi_shimbun.py
```diff
@@ -0,0 +1,59 @@
...
+
+        @attribute
+        def topics(self) -> List[str]:
```
addie9800, line 44: I would filter the following words: `ニュース` (News) (and perhaps all strings containing it as well, since we also have topics like `Tokyo News`), `話題` (Topics)

Also, I would remove the substring `話題・` (`Topics/`). It is used like `話題・深掘り` (`Topics/Deep Digging`)

addie9800 (changes requested):
Thanks for adding these as well 👍

src/fundus/publishers/jp/tokyo_chunichi_shimbun.py
MaxDall: I would suggest to just leave it as is

addie9800 (dismissed):
I noticed two more minor issues and took the liberty of fixing them. I hope that's OK with you. The author_selector needed to be modified since everything that is selected in the caption is removed. Looks good to me though 👍

MaxDall:
> I noticed two more minor issues and took the liberty of fixing them. I hope that's OK with you. The author_selector needed to be modified since everything that is selected in the caption is removed. Looks good to me though 👍

Of course 👍 Thanks for doing that!

PR #686, opened 2025-01-14
add `Nikkei`

MaxDall:


src/fundus/publishers/jp/nikkei.py
```diff
@@ -0,0 +1,41 @@
...
+        @attribute
+        def publishing_date(self) -> Optional[datetime.datetime]:
+            return generic_date_parsing(self.precomputed.ld.bf_search("datePublished"))
+
+        @attribute
+        def images(self) -> List[Image]:
+            return image_extraction(
+                doc=self.precomputed.doc,
+                paragraph_selector=self._paragraph_selector,
+                lower_boundary_selector=CSSSelector("div.paywall_pzomzzc"),
```
addie9800, line 40: The lower_boundary needs to be set a bit higher: the image of the magazine is extracted as well: https://www.nikkei.com/article/DGXZQOUB1588V0V10C25A1000000/
They seem to occasionally provide an author (see image 1 of the same article)

src/fundus/publishers/jp/nikkei.py
```diff
@@ -0,0 +1,41 @@
...
+from fundus.parser.utility import (
+    extract_article_body_with_selector,
+    generic_date_parsing,
+    image_extraction,
+)
+
+
+class NikkeiParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = CSSSelector("section[data-track-article-content] > p")
```
addie9800, line 16: From my feeling, [this](https://www.nikkei.com/article/DGXZQOUB021II0S5A100C2000000/) could be a summary of the article

src/fundus/publishers/jp/nikkei.py
```diff
@@ -0,0 +1,41 @@
...
+            )
+
+        @attribute
+        def title(self) -> Optional[str]:
+            return self.precomputed.ld.bf_search("headline")
+
+        @attribute
+        def publishing_date(self) -> Optional[datetime.datetime]:
+            return generic_date_parsing(self.precomputed.ld.bf_search("datePublished"))
+
```
addie9800, line 34: One could consider using the elements at `/html/body/div[5]/main/article/header/div[2]/a` as topics. One might have to filter: 日経ヴェリタス which is just Nikkei Veritas or sth similar. But there are some topics like: Economy, Southwest Oceania, Science

addie9800 (changes requested):
Thanks for adding this 👍

src/fundus/publishers/jp/nikkei.py
MaxDall: Could you provide an example article that includes these elements? I couldn't find one.

src/fundus/publishers/jp/nikkei.py
MaxDall: Didn't manage to find a reliable selector which does not clash with [this](https://www.nikkei.com/article/DGXZQOUB148MY0U5A110C2000000/) article

src/fundus/publishers/jp/nikkei.py
MaxDall: Didn't manage to find a reliable selector which does not clash with [this](https://www.nikkei.com/article/DGXZQOUB148MY0U5A110C2000000/) article

src/fundus/publishers/jp/nikkei.py
addie9800: This one for example: https://www.nikkei.com/article/DGXZQOUA1649Y0W5A110C2000000/ has the topic Disasters and Weather
In this one and a couple of others the topic is the area that is relevant to the article: https://www.nikkei.com/article/DGXZQOCC2328F0T20C25A1000000/ (here for example Gunma)

addie9800 (approved):

PR #687, opened 2025-01-20
Fix a bug in `publisher_coverage.py`

MaxDall:
This resolves a bug in the publisher coverage script where an `Exception` used as an attribute value for an article was not accounted for, causing the script to crash.


addie9800 (approved):
Thanks for catching this

PR #688, opened 2025-01-21
Add `SankeiShimbun`

MaxDall:


addie9800 (changes requested):
Thanks for also adding this publisher 👍

addie9800 (approved):

PR #689, opened 2025-01-22
Add `NikkanGeadai`

MaxDall:
This PR also adds a new utility function to transform `<br>` into `<p>` elements. Maybe @addie9800 wants to give it a shot for `TheNamibian` and also `TimesOfIndia`.


src/fundus/parser/utility.py
```diff
@@ -271,6 +271,55 @@ def get_meta_content(root: lxml.html.HtmlElement) -> Dict[str, str]:
     return metadata
 
 
+def transform_breaks_to_paragraphs(element: lxml.html.HtmlElement, **attribs: str) -> lxml.html.HtmlElement:
```
addie9800, line 274: This is a really good idea, but it isn't (yet) viable for `TimesOfIndia`. Not necessarily all paragraphs are separated with `<br>` blocks, but also follow `<div>` blocks for example, which is currently producing issues.

addie9800 (changes requested):
Thanks for this addition 👍

src/fundus/parser/utility.py
MaxDall: Thanks for giving it a try! I will look into this in another PR.

addie9800 (approved):

PR #691, opened 2025-01-23
Add Tanzanian Publishers

addie9800:


src/fundus/publishers/tz/daily_news_tz.py
```diff
@@ -0,0 +1,57 @@
...
+    utility,
+)
+from fundus.parser.utility import image_extraction
+
+
+class DailyNewsTZParser(ParserProxy):
+    class V1(BaseParser):
+        _summary_selector = CSSSelector("div.cs-entry__subtitle")
+        _subheadline_selector = XPath("//div[@class='entry-content']//p[not(text() or position()=1)]//span//strong")
+        _paragraph_selector = XPath("//div[@class='entry-content']//p[text() or position()=1]")
```
MaxDall, line 23: The paragraph selector might be too restrictive: Most of [this](https://habarileo.co.tz/musk-usaid-ni-shirika-la-kihalifu/) doesn't get extracted.

MaxDall (changes requested):
Thanks a lot for adding the first tanzanian publishers 🚀

src/fundus/publishers/tz/daily_news_tz.py
addie9800: Thanks for catching that, there were some more things off with the paragraph selector. Should be good to go now.

MaxDall (approved):

PR #695, opened 2025-02-02
Update `Dagbladet`

addie9800:
Apologies for this PR, it's a bit messy, but unfortunately, Dagbladet decided to keep its old layout on some of the subpages like https://borsen.dagbladet.no/nyheter/tror-kina-kan-velge-tesla-hevn/82617028. Another issue we should discuss is in this article: https://www.dagbladet.no/nyheter/bolgneset-utbygging-legges-pa-is/82591507. They have the normal summary, as always, and an AI-generated one. I would argue that we should also extract the summary; where I am unsure is whether or not to include the AI disclaimer in the plaintext. I left it in for now, but I would also agree to remove it.

### DIFF
--- a/src/fundus/publishers/no/dagbladet.py
+++ b/src/fundus/publishers/no/dagbladet.py
@@ -1,6 +1,6 @@
 import datetime
 import re
-from typing import List, Optional
+from typing import List, Optional, Union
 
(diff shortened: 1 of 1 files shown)


MaxDall (approved):
@addie9800 Thanks for fixing this! LGTM.

PR #696, opened 2025-02-02
Add `DK`

addie9800:
Adds the first Danish publisher


MaxDall (changes requested):
@addie9800 Thanks a lot for adding our first danish publisher ❤️

MaxDall (approved):

PR #699, opened 2025-02-03
Add `PT`

addie9800:


MaxDall (changes requested):
@addie9800 thanks for adding 🚀

src/fundus/publishers/pt/the_portugal_news.py
```diff
@@ -0,0 +1,54 @@
...
+    generic_date_parsing,
+    image_extraction,
+)
+
+
+class ThePortugalNewsParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = XPath("//div[@class='article-body']//p[string-length(text())>1]")
+        _subheadline_selector = XPath("//div[@class='article-body']/p/b")
+        _summary_selector = XPath("//div[@class='fs-4 font-semibold mb-3']")
```
MaxDall, line 20: In [this](https://www.theportugalnews.com/news/2025-02-03/algarve-breaks-records/95305) article the `Related Topics` at the end gets extracted as well.

addie9800: Thanks for catching this, it should be fixed now

MaxDall (approved):

PR #700, opened 2025-02-04
Support for 2 new italian newspapers - Corriere della Sera & Il Giornale

ruggsea:
Following up on my previous contribution of La Repubblica, I've added support for two more major Italian newspapers to expand the coverage of Italian media:

## Additions:
- Created a new parser for Corriere della Sera (www.corriere.it)
- Created a new parser for Il Giornale (www.ilgiornale.it)
- Updated the IT publisher group with the new newspapers
- Set up RSS feeds and sitemap crawling for both sources

## What can the parsers extract?
Both parsers can extract:
- Article titles
- Full article body (including handling of advertisements)
- Author information
- Publishing dates
- Topics

## Testing:
- Created and passed unit tests for both newspapers
- Updated the Supported Sources documentation
- All linting and mypy checks pass

This addition brings the total number of supported Italian newspapers to three, providing a broader coverage of Italian news sources.


ruggsea:
Apparently Corriere della Sera was already implemented (in d225bc20b368e68233fef0147f6f52bd95de5d37) while I was doing it too! In any case, my implementation has many different sitemaps as I've noticed that the "items" sitemaps referenced in https://www.corriere.it/rss/sitemap_v2.xml are very incomplete

src/fundus/publishers/it/__init__.py
```diff
@@ -22,3 +24,90 @@ class IT(metaclass=PublisherGroup):
...
+            "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/121.0.0.0 Safari/537.36",
+            "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8",
+            "Accept-Language": "en-US,en;q=0.9",
+            "Accept-Encoding": "gzip, deflate, br",
+            "Connection": "keep-alive",
+            "Upgrade-Insecure-Requests": "1",
+        },
+        sources=[
+            RSSFeed("https://www.ilgiornale.it/feed.xml"),
+            RSSFeed("https://www.ilgiornale.it/feed/rss.xml"),
```
addie9800, line 109: This link returns a 404 error

src/fundus/publishers/it/__init__.py
```diff
@@ -22,3 +24,90 @@ class IT(metaclass=PublisherGroup):
...
     )
+
+    CorriereDellaSera = Publisher(
+        name="Corriere Della Sera",
+        domain="https://www.corriere.it",
+        parser=CorriereDellaSeraParser,
+        sources=[
+            # Main RSS feeds
+            RSSFeed("https://www.corriere.it/feed-hp/homepage.xml"),
+            RSSFeed("https://www.corriere.it/rss/ultimora.xml"),
```
addie9800, line 35: This Feed seems to be empty

src/fundus/publishers/it/__init__.py
```diff
@@ -22,3 +24,90 @@ class IT(metaclass=PublisherGroup):
...
+            RSSFeed("https://www.corriere.it/dynamic-feed/rss/section/Dataroom.xml"),
+            RSSFeed("https://www.corriere.it/dynamic-feed/rss/section/lettere-al-direttore.xml"),
+            RSSFeed("https://www.corriere.it/dynamic-feed/rss/section/lo-dico-al-corriere.xml"),
+            RSSFeed("https://www.corriere.it/dynamic-feed/rss/section/frammenti-di-ferruccio-de-bortoli.xml"),
+            # Main sitemaps
+            Sitemap("https://www.corriere.it/rss/sitemap_v2.xml"),
+            Sitemap("https://www.corriere.it/salute/sitemap-dizionario-corriere-salute.xml"),
```
addie9800, line 42: This sitemap seems to not contain any articles, as far as I can see

src/fundus/publishers/it/__init__.py
```diff
@@ -22,3 +24,90 @@ class IT(metaclass=PublisherGroup):
...
+            Sitemap("https://www.corriere.it/dynamic-sitemap/sitemap-last-100/Sette.xml"),
+            Sitemap("https://www.corriere.it/dynamic-sitemap/sitemap-last-100/Moda.xml"),
+            Sitemap("https://www.corriere.it/dynamic-sitemap/sitemap-last-100/BuoneNotizie.xml"),
+            Sitemap("https://www.corriere.it/dynamic-sitemap/sitemap-last-100/lettere__al__direttore.xml"),
+            Sitemap("https://www.corriere.it/dynamic-sitemap/sitemap-last-100/lo__dico__al__corriere.xml"),
+            Sitemap("https://www.corriere.it/dynamic-sitemap/sitemap-last-100/Frammenti-ferruccio-de-bortoli.xml"),
+            # Section sitemaps
+            Sitemap("https://www.corriere.it/rss/sitemap/Motori.xml"),
+            Sitemap("https://www.corriere.it/rss/sitemap/Cultura.xml"),
+            Sitemap("https://vivimilano.corriere.it/sitemap_index.xml"),
```
addie9800, line 80: This one seems to lead to an entire different subpage, which is not supported by the current parser (and also doesn't necessarily need to)

src/fundus/publishers/it/__init__.py
```diff
@@ -22,3 +24,90 @@ class IT(metaclass=PublisherGroup):
...
+            Sitemap("https://www.corriere.it/oroscopo/sitemap.xml"),
+            Sitemap("https://www.corriere.it/elezioni/sitemap/sitemap.xml"),
+            Sitemap("https://www.corriere.it/sport/risultati-live/sitemap.xml"),
+            Sitemap("https://www.corriere.it/salute/il-medico-risponde/sitemap.xml"),
+            Sitemap("https://www.corriere.it/rss/sitemap/lettere-al-direttore.xml"),
+            Sitemap("https://www.corriere.it/rss/sitemap/lo-dico-al-corriere.xml"),
+            Sitemap("https://www.corriere.it/rss/sitemap/Cook-Last.xml"),
+            Sitemap("https://www.corriere.it/economia/chiedi-esperto/sitemap.xml"),
+            Sitemap("https://www.corriere.it/economia/chiedi-esperto/news/sitemap.xml"),
+            Sitemap("https://www.corriere.it/studio/sitemap-studio.xml"),
```
addie9800, line 91: This sitemap no longer exists

src/fundus/publishers/it/__init__.py
ruggsea: My idea was that you can never know what they are actually going to use, so better safe than sorry (also to make up for lack of coverage of the main index)

addie9800: Fair enough, I would also be fine with leaving it in

src/fundus/publishers/it/__init__.py
addie9800: While I agree that including it in the extracted articles could be beneficial, the current parser does not make those entries parsable.

ruggsea:
Thank you @addie9800 for the thorough review! I think I should have addressed all comments (and rebased off current master and handled the image attribute):

For Il Giornale:
- Removed redundant `free_access` method (already in BaseParser)
- Removed unused logger
- Simplified selectors as suggested:
  - Updated summary selector to work without typography--content restriction
  - Simplified paragraph selector to the suggested XPath
- Fixed type errors using proper HtmlElement handling
- Used the article provided for test genearion
- Removed the 404 RSS feed, keeping only the working one

For Corriere della Sera:
- Kept the existing implementation and added my topics extraction from breadcrumbs
- Cleaned up the sitemaps configuration:
  - Removed health dictionary sitemap (not parseable)
  - Removed vivimilano sitemap (different subpage)
  - Kept only working dynamic sitemaps
  - Better organized into categories
  
  Let me know what you think!

src/fundus/publishers/it/il_giornale.py
```diff
@@ -0,0 +1,118 @@
...
+            return []
+
+        @attribute
+        def images(self) -> List[Image]:
+            # Extract images using the utility function
+            images = image_extraction(
+                doc=self.precomputed.doc,
+                paragraph_selector=self._paragraph_selector,
+                image_selector=self._image_selector,
+                author_selector=re.compile(r"(?:Foto:?\s*)?(?P<credits>[^()]+)(?:\s*\([^)]+\))?$"),
```
addie9800, line 99: The pattern needs to be modified to select only if an attribution is being made. What I mean by that can exemplarily be seen in this image:
```
Fundus-Article Cover-Image:
-URL:			 'https://img.ilgcdn.com/sites/default/files/styles/xl/public/foto/2025/02/01/1738434667-azs3g51p0mzhzbpw1kqf-ansa.jpeg?_=1738434667&format=webp'
-Description:	 'Si fingono sordomute per derubarli. Ma le vittime sono due carabinieri'
-Caption:		 None
-Authors:		 ['Si fingono sordomute per derubarli. Ma le vittime sono due carabinieri']
-Versions:		 [300x169, 500x281, 800x450, 1200x675]
```
Especially if the author attribution is done in the caption, the utility function would remove the match from the caption and save it as the author. But did you see any case, where there was an author attribution done by the publisher? I searched through their website but didn't see any instances. If you also didn't find any instance, we can, I think, safely remove the custom author_selector and modify it accordingly, should we at some point stumble across an article image author attribution.

addie9800 (changes requested):
Thank you for addressing this so quickly. I just have a couple comments about the images attribute and then we should be good to go 🚀

src/fundus/publishers/it/il_giornale.py
ruggsea: Ok makes sense!

ruggsea:
Thanks for the fast feedback. I should have fixed/made more broad the image selection for IlGiornale
- Fixed image selector to capture all images (it should now work with the automotive article)
- Removed fallback case for cover images
- Removed unused author_selector since no attributions found

Also, now the <cite> tags are being included as well!

MaxDall:
@ruggsea Thanks a lot for adding this ❤️ and thanks a lot for the quick and in-depth review @addie9800. I'm gonna merge this now and prepare a new minor release.

Issue #702, opened 2025-02-05
[Question]: Json serialization inquiries


ruggsea:
### Question

Is there a particular reason for why when saving articles to json (for example by specifying `save_to_file` in the crawling or by using the `Article.to_json()` method) things like the URL of the articles and/or the specific publisher the articles came from are not included? Or maybe they are and I am being mistaken?

In any case, I was also wondering if the Image attribute is also already supported in the serialization, because I've got some errors when trying to serialize articles containing that attribute (but I am not filing it as a bug cause I am not sure whether it was a mistake in my devel env)

MaxDall:
@ruggsea Thanks for catching that. While the `Image` object is serializable, we missed a bug in the articles' `to_json` method causing these issues. I will work on a fix for this.

PR #703, opened 2025-02-11
Fix article serialization for `images`

MaxDall:
There was a bug when serializing the `images` attribute described in #702. This PR provides a quick fix so that the `to_json` method also works for `images`.

### DIFF
--- a/src/fundus/scraping/article.py
+++ b/src/fundus/scraping/article.py
@@ -9,7 +9,7 @@ from colorama import Fore, Style
 from fundus.logging import create_logger
 from fundus.parser import ArticleBody, Image
 from fundus.scraping.html import HTML
-from fundus.utils.serialization import JSONVal, is_jsonable
+from fundus.utils.serialization import JSONVal
 
 logger = create_logger(__name__)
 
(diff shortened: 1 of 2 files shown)


addie9800:
I am having some issues with this. 

```
from fundus import PublisherCollection, Crawler,
from tests.utility import get_test_articles

publisher = PublisherCollection.de.Tagesschau
crawler = Crawler(publisher)

if __name__ == "__main__":
    for article in crawler.crawl(max_articles=1, only_complete=False, error_handling='suppress', save_to_file="E:\\Temp\\test.json"):
        print(article)
```
Running this code gives me an error: `TypeError: Attribute 'topics' of type <class 'lxml.etree._ElementUnicodeResult'> is not JSON serializable`. This can be easily fixed by modifying your code slightly:
```
        def serialize(v: Any) -> JSONVal:
            if hasattr(v, "serialize"):
                return v.serialize()  # type: ignore[no-any-return]
            elif isinstance(v, datetime):
                return str(v)
            elif isinstance(v, str):
                return v
            raise TypeError(f"Attribute {attribute!r} of type {type(v)!r} is not JSON serializable")
```
Yet, this causes this error: `line 342, in <lambda>
    json.dumps(crawled_articles, default=lambda o: o.to_json(), ensure_ascii=False, indent=4)
AttributeError: 'datetime.datetime' object has no attribute 'to_json'` which causes me to scratch my head. I have played around with it for some time now, but can't lay my finger on the issue. Perhaps you see can see it.

addie9800 (approved):

PR #704, opened 2025-02-11
Add `MX` + minor bug fixes

addie9800:
This PR:
- adds `TheMexicoNews` - the first Mexican publisher
- Fixes the Decoding of HTML Unicode characters in the parsing of `LinkedDataMapping`
- Fixes an issue in `text_content()` where text was omitted, if contained within the tail of an excluded tag. To reproduce, check this article: https://mexiconewsdaily.com/news/bad-idea-mexico-pushes-back-trump-universal-steel-tariffs/ without the change, the sentence `The tariffs are set to take effect on March 12 and, as things stand, will apply to all countries’ steel and aluminum exports to the United States.` is not extracted.


addie9800:
This PR has been reopened as #734 to support latest changes in `master`

PR #710, opened 2025-02-24
Add `max_articles_per_publisher` parameter to `crawl`

MaxDall:
This PR adds an option to specify the maximum number of articles that should be crawled per publisher to enable users to balance corpora.


addie9800 (dismissed):
Very nice, this also now saves the corresponding publisher with the article. 🚀

PR #712, opened 2025-02-25
Extend timeout in publisher coverage

addie9800:
The Tanzanian publishers at least need a larger timeout. They fail, because the Sitemaps take too long to load. Since this may also apply to other publishers, I would suggest bumping up the timout.

### DIFF
--- a/scripts/publisher_coverage.py
+++ b/scripts/publisher_coverage.py
@@ -15,7 +15,7 @@ from fundus.scraping.article import Article
 
 def main() -> None:
     failed: int = 0
-    timeout_in_seconds: int = 20
+    timeout_in_seconds: int = 30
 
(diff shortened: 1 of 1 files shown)


scripts/publisher_coverage.py
```diff
@@ -15,7 +15,7 @@
 
 def main() -> None:
     failed: int = 0
-    timeout_in_seconds: int = 10
+    timeout_in_seconds: int = 20
```
MaxDall, line 18:
```suggestion
    timeout_in_seconds: int = 30
```
It seems that I accidentally reduced the timeout to 10 seconds here ca41bb1afc89e792d5b69b5e7897bf311f79c074

MaxDall (changes requested):

PR #713, opened 2025-02-27
Properly release resources  

MaxDall:
This PR addresses a longstanding issue with the Crawler class. Previously, when exiting the ThreadPool, individual publisher threads were not properly closed. As a result, joining the pool required waiting for every URLSource to fully exhaust. This update introduces an event-based system, allowing events to be registered in a thread-safe dictionary and checked within threads. Additionally, the WebSource class now includes EventDict as a class attribute, allowing the main thread to terminate individual web sources more effectively.

### DIFF
--- a/src/fundus/scraping/crawler.py
+++ b/src/fundus/scraping/crawler.py
@@ -55,7 +55,7 @@ from fundus.publishers.base_objects import Publisher, PublisherGroup
 from fundus.scraping.article import Article
 from fundus.scraping.delay import Delay
 from fundus.scraping.filter import ExtractionFilter, Requires, RequiresAll, URLFilter
-from fundus.scraping.html import CCNewsSource
+from fundus.scraping.html import CCNewsSource, WebSource
 from fundus.scraping.scraper import CCNewsScraper, WebScraper
 from fundus.scraping.session import session_handler
 from fundus.scraping.url import URLSource
@@ -337,6 +337,10 @@ class CrawlerBase(ABC):
                     tuple(fitting_publishers), error_handling, build_extraction_filter(), url_filter
                 ):
                     if max_articles_per_publisher and article_count[article.publisher] == max_articles_per_publisher:
+                        if isinstance(self, Crawler) and not WebSource.__EVENTS__.is_event_set(
+                            "stop", article.publisher
+                        ):
+                            WebSource.__EVENTS__.set_event("stop", article.publisher)
                         if sum(article_count.values()) == len(self.publishers) * max_articles_per_publisher:
                             break
                         continue
@@ -426,6 +430,8 @@ class Crawler(CrawlerBase):
         self.ignore_robots = ignore_robots
         self.ignore_crawl_delay = ignore_crawl_delay
 
+        WebSource.__EVENTS__.clear_all()
+
     def _fetch_articles(
         self,
         publisher: Publisher,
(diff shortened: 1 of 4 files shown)


src/fundus/scraping/crawler.py
```diff
@@ -464,15 +470,20 @@ def _single_crawl(
...
-        with ThreadPool(processes=len(publishers) or None) as pool, session_handler.context(len(publishers), 1):
-            yield from pool_queue_iter(pool.map_async(wrapped_article_task, publishers), result_queue)
+        try:
+            with ThreadPool(processes=len(publishers) or None) as pool, session_handler.context(len(publishers), 1):
+                yield from pool_queue_iter(pool.map_async(wrapped_article_task, publishers), result_queue)
+        finally:
+            logger.debug(f"Shutting down {type(self).__name__!r} ...")
+            WebSource.__EVENTS__.set_all("stop")
+            pool.join()
+            logger.debug("Shutdown done")
```
addie9800, line 486: Or alternatively, we could clear all events here after successfully shutting down the pool.

src/fundus/scraping/scraper.py
```diff
@@ -95,6 +95,9 @@ def __init__(
         parser_mapping: Dict[str, ParserProxy] = {publisher.name: publisher.parser}
         super().__init__(*html_sources, parser_mapping=parser_mapping)
 
+        WebSource.__EVENTS__.register_event("stop", publisher.name)
+        WebSource.__EVENTS__.register_event("bridge", publisher.name)
```
addie9800, line 99: What does this event do? I think it is unused as far as I can see.

src/fundus/utils/events.py
```diff
@@ -0,0 +1,80 @@
...
+        self._lock = threading.Lock()
+
+    @staticmethod
+    def _get_identifier() -> int:
+        return threading.get_ident()
+
+    def _resolve(self, key: Union[int, str, None]) -> int:
+        """Resolves a given key to a thread identifier
+
+        Should only be used within a Lock!
```
addie9800, line 25: I am not sure what the possible consequences are when not used within a lock, but if they are severe enough, maybe we should enforce this. ChatGPT suggests something like this: [https://chatgpt.com/share/67c2341a-13ac-8000-b4f1-ae69ab7e1685](https://chatgpt.com/share/67c2341a-13ac-8000-b4f1-ae69ab7e1685)

src/fundus/utils/events.py
```diff
@@ -0,0 +1,80 @@
...
+                    events[event].set()
+                else:
+                    for flag in events.values():
+                        flag.set()
+
+    def clear_all(self, event: Optional[str] = None):
+        with self._lock:
+            for events in self._events.values():
+                if event is not None and event not in events:
+                    events[event].set()
```
addie9800, line 69: I think this might be a copy-paste error. If not, it is a bit unintuitive to set() in the clear_all() function

src/fundus/utils/events.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def set_event(self, event: str, key: Union[int, str, None] = None):
+        with self._lock:
+            self._events[self._resolve(key)][event].set()
+            logger.debug(f"Set event {event!r} for {self._resolve(key)}")
+
+    def set_all(self, event: Optional[str] = None):
+        with self._lock:
+            for events in self._events.values():
+                if event is not None and event not in events:
```
addie9800, line 59: I think I am not a fan of this part and the corresponding section in clear_all(). The first idea I got from the name is that all events are set for all threads, but that is not necessarily the case. I think maybe a short docstring here could improve the understanding. Perhaps the function could also be renamed to `set_for_all()` and then the docstring explains the behavior if `event=None`

addie9800: Also, I think that there is a typo and the correct condition should be `event is not None and event in events`

addie9800 (changes requested):
Hey, thank you so much for addressing this, this is a great addition 🚀  . I just have a couple questions and remarks.

src/fundus/scraping/html.py
```diff
@@ -130,17 +133,27 @@ def __init__(
...
+    @property
+    def _is_stopped(self):
+        return self.__EVENTS__.is_event_set("stop")
+
     def fetch(self, url_filter: Optional[URLFilter] = None) -> Iterator[HTML]:
         combined_filters: List[URLFilter] = ([self.url_filter] if self.url_filter else []) + (
             [url_filter] if url_filter else []
         )
 
+        if self._is_stopped:
```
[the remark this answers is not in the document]
MaxDall: I replaced the for with a while loop to get rid of one check.

src/fundus/scraping/crawler.py
```diff
@@ -426,6 +430,8 @@ def filter_publishers(publisher: Publisher) -> bool:
         self.ignore_robots = ignore_robots
         self.ignore_crawl_delay = ignore_crawl_delay
 
+        WebSource.__EVENTS__.clear_all()
```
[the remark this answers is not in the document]
MaxDall: Good point. I will move it to the shutdown procedure as you suggested in a later comment.

src/fundus/scraping/scraper.py
MaxDall: Thanks for catching that, it's a debug leftover.

src/fundus/utils/events.py
MaxDall: There are several issues with the approach ChatGPT suggested, the main one IMO being that there is no way to ensure which thread currently locks the lock. The possible consequence is a write/read race condition. Since the method is private and meant to be used only within the class itself, IMO it is fine to not enforce the lock and and only leave a hint in the doc string.

src/fundus/utils/events.py
MaxDall: Thanks for catching

addie9800 (approved):
Looks good 👍

PR #715, opened 2025-03-25
add url_filter to voa

addie9800:
This PR:
- removes the NewsMap that only contains irrelevant pages
- filters the irrelevant pages from the remaining sitemaps
The pages filtered are similar to this page: https://voanews.com/a/8003948.html (will be redirected)


MaxDall (approved):
@addie9800 Thanks a lot for cleaning the sources. Helps a lot.

PR #717, opened 2025-04-01
Update ˋBusinessInsiderˋ

addie9800:
This PR:
- adds Parser Version ˋV2ˋ and removes ˋimagesˋ supportˋ for the new version. The reason is that BI is using a custom attribute type to save the image source data, which is unsupported as of now


MaxDall (approved):
@addie9800 Thanks a lot for addressing the layout changes. What would we have to do to support image extraction again?

addie9800:
> @addie9800 Thanks a lot for addressing the layout changes. What would we have to do to support image extraction again?

In this specific case, they have added a custom attribute named ˋdata-srcsˋ, within that they use a custom object which contains the link to the image and some additional data. It usually looks like this:
`data-srcs="{&quot;https://i.insider.com/67e5850b69253ccddf9a7c4f&quot;:{&quot;contentType&quot;:&quot;image/png&quot;,&quot;aspectRatioW&quot;:606,&quot;aspectRatioH&quot;:455}}"`

PR #721, opened 2025-04-03
Register default `stop` event for WevSource

MaxDall:


addie9800 (approved):
Looks good. Thanks for fixing this 👍

PR #723, opened 2025-04-04
Make network connections interruptible 

MaxDall:
This PR ...
- fixes a race condition that prevented exceptions from being propagated
- makes open connections interruptible so the pool can properly shut down
- fixes the publisher scipt to run again

### DIFF
--- a/src/fundus/scraping/session.py
+++ b/src/fundus/scraping/session.py
@@ -1,15 +1,53 @@
+import socket
 import threading
+from concurrent.futures import ThreadPoolExecutor, TimeoutError
 from contextlib import contextmanager
-from typing import Iterator, Optional
+from dataclasses import dataclass
+from typing import Optional
 
 import requests.adapters
-from typing_extensions import Self
 
 from fundus.logging import create_logger
+from fundus.utils.events import __EVENTS__
 
 logger = create_logger(__name__)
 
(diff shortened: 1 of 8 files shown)


src/fundus/scraping/session.py
```diff
@@ -1,17 +1,55 @@
...
+        else:
+            url = kwargs.get("url")
+
+        with ThreadPoolExecutor(max_workers=1) as executor:
+            future = executor.submit(_req)
+            while True:
+                try:
+                    response = future.result(timeout=1)
+                except TimeoutError:
+                    if __EVENTS__.is_event_set("stop"):
```
addie9800, line 38: I am a bit confused by this or at least I don't understand the theory behind it. I have a couple of questions:
- Why does it make sense to start a ThreadPoolExecutor if we limit it to one worker, it seems odd.
- Also a one second timeout feels a bit too little, when you try to load larger sitemaps (but then again, I might be misunderstanding the way it works). 
- Why does it work to check for the stop event in the except block. Even though it get's called regularly, when I played around with it a bit, I don't see why that happens, since it should only occur on a timeout and most websites respond faster than within one second. 
Sorry for the load of questions, I am just really unsure, what's going on :)

addie9800 (approved):
Thanks for addressing this issue. This looks good to me. 👍

PR #725, opened 2025-04-08
Add `CZ`

horychtom:
Adding CZ Publishers Group for Czech Republic's publishers.

Starting with SeznamZpravy publisher


src/fundus/publishers/cz/seznamzpravy.py
```diff
@@ -0,0 +1,66 @@
...
+)
+
+
+class SeznamZpravyParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = XPath(
+            "//article//div[contains(@class, 'speakable')"
+            "and not(ancestor::div[contains(@class, 'mol-post-card__body')])]"
+            "/p[contains(@class, 'e_bg')]"
+        )
```
addie9800, line 25: I think we need to determine, if the red boxes, that provide additional information, are a part of the article. For example in this article: https://www.seznamzpravy.cz/clanek/ekonomika-byznys-reality-airbnb-znovu-obsazuje-stred-prahy-dopad-muze-byt-citelny-273635 only the subheadlines Amsterdam, Barcelona... are extracted, but the rest isn't. Since I assume you speak Czech, I would leave it up to you to evaluate whether you would consider the box as additional info that fits within the article or if it really is just additional and would not be missing in the article. Nevertheless, whichever way you choose, the paragraph_selector needs to be refined to include either all or nothing of the info box

addie9800: Also, if you choose to include the boxes, then consequently you would also have to refine the image_selector to also include the images in those boxes as well. I have a feeling though that including the boxes might break up the structure of the article

addie9800 (changes requested):
Thank you so much for adding our first Czech publisher. This is great work :rocket: I just have one comment regarding your paragraph_selector

src/fundus/publishers/cz/seznamzpravy.py
horychtom: Thank you for the feedback! 
I have decided to not include the info boxes, as they often include wikipedia-type of facts that disrupt the flow of the article and are not generally part of the articles story.
I went back to the selectors and made them more simple. Only direct children of main article class are now selected, therefore the info box is excluded implicitely.
For the images, I added image selector that excludes the images with the info-box parent. As they could also come a bit out of nowhere without the context. Finally I updated the test case, excluding the images from the info-boxes.

addie9800 (approved):
Sorry, for the delay. This looks great. Thank you 👍

PR #726, opened 2025-04-12
Rework language attribution

MaxDall:
This reworks parts of the language attribution PR, in particular, adding a new "aware" publisher object called `FilteredPublisher`


addie9800 (approved):
This looks good. Thanks for the rework.

PR #730, opened 2025-04-15
Add `IL`

addie9800:
This PR adds the first Israeli Publisher. Note that it seems that the publisher did not publish any new articles in the last four years, but has a substantial archive, which is why I think it might be an asset nevertheless.


addie9800:
This PR has been reopened as #736 to address the latest changes in `master`

PR #734, opened 2025-04-16
Add `MX` + minor bug fixes

addie9800:
This PR:
- adds `TheMexicoNews` - the first Mexican publisher
- Fixes the Decoding of HTML Unicode characters in the parsing of `LinkedDataMapping`
- Fixes an issue in `text_content()` where text was omitted, if contained within the tail of an excluded tag. To reproduce, check this article: https://mexiconewsdaily.com/news/bad-idea-mexico-pushes-back-trump-universal-steel-tariffs/ without the change, the sentence `The tariffs are set to take effect on March 12 and, as things stand, will apply to all countries’ steel and aluminum exports to the United States.` is not extracted.


src/fundus/parser/utility.py
```diff
@@ -552,9 +554,9 @@ def get_longest_string(strings: List[str]) -> str:
         return sorted(strings, key=len)[-1]
 
     if srcset := cast(List[str], _srcset_selector(node)):
-        return parse_srcset(get_longest_string(srcset))
+        return parse_srcset(normalize_whitespace(get_longest_string(srcset)))
     elif src := cast(List[str], _src_selector(node)):
-        return {"1x": get_longest_string(src)}
+        return {"1x": normalize_whitespace(get_longest_string(src))}
```
MaxDall: Is this added on purpose again?

src/fundus/publishers/__init__.py
```diff
@@ -8,7 +8,6 @@
 from fundus.publishers.ca import CA
 from fundus.publishers.ch import CH
 from fundus.publishers.cn import CN
-from fundus.publishers.cz import CZ
```
MaxDall: How come that CZ is removed in this PR

src/fundus/publishers/mx/__init__.py
```diff
@@ -0,0 +1,22 @@
+from fundus.publishers.base_objects import Publisher, PublisherGroup
+from fundus.scraping.url import Sitemap
+
+from ...scraping.filter import inverse, regex_filter
+from .mexico_news_daily import MexicoNewsDailyParser
```
MaxDall, line 5: For imports outside the publisher directory, we normally use package imports `from fundus ...`

src/fundus/parser/utility.py
```diff
@@ -189,8 +190,9 @@ def sanitize_json(text: str) -> Optional[str]:
 
     # substitute "bad" values
     sanitized = re.sub(_json_undefined, r"\g<key>:null", sanitized)
+    removed_unicode = html.unescape(re.sub(r"&quot;?", "\\&quot;", sanitized))
```
MaxDall: Is this a general problem or specific to `&quot;?`? I don't quite understand what's the problem here.

MaxDall (changes requested):

src/fundus/parser/utility.py
addie9800: The issue is that the unescaping breaks the JSON, because it inserts unescaped quotation marks. You can observe this when trying to overwrite the JSON of ElMundo, without previously replacing all instances of `&quot;` with `\\&quot;`. It's so far back, I honestly don't remember why I added the ? operator and since it seems to work just as well without it, I will remove it.

src/fundus/publishers/__init__.py
addie9800: I am just as confused as you are... It's back now

src/fundus/parser/utility.py
addie9800: I vaguely remember yes, but I honestly can't remember what it was and I can't reproduce any reason as of now, why I added it. But whatever the issue was, doesn't seem to be there anymore, so I'll remove it.

src/fundus/parser/utility.py
MaxDall: Thanks for the explanation. After understanding what's going on, I like the solution you came up with. Thanks a lot.

MaxDall (approved):
@addie9800 Thanks a lot for adding the first Mexican publisher and fixing the unescaped HTML 🚀

PR #737, opened 2025-04-16
Add `IL`

addie9800:
This PR adds the first Israeli Publisher. Note that it seems that the publisher did not publish any new articles in the last four years, but has a substantial archive, which is why I think it might be an asset nevertheless.


src/fundus/publishers/il/__init__.py
```diff
@@ -0,0 +1,23 @@
+from fundus.publishers.base_objects import Publisher, PublisherGroup
+from fundus.scraping.url import Sitemap
+
+from ...scraping.filter import inverse, regex_filter
+from .israel_nachrichten import IsraelNachrichtenParser
```
MaxDall, line 5: We normally use package imports here.

addie9800: Ah yes, my bad. That happened through automatic importing in PyCharm

PR #744, opened 2025-04-17
Add unit test if `default_language` is ISO 639 language code

MaxDall:


addie9800:
Your addition looks good to me, coincidentally was thinking to do the same. I think it might be useful to do the same sanity check for possibly set custom languages in the URLSource. What do you think?

MaxDall:
@addie9800 Are there still any open questions regarding this PR? If not, I would proceed with merging it.

addie9800 (approved):

PR #745, opened 2025-04-22
Deprecate Attributes

addie9800:
As by per the suggestion from @MaxDall, I looked into the possibility to deprecate attributes for publishers that no longer support a certain attribute. This has the advantage that articles are no longer marked as incomplete, if a deprecated attribute is missing. In short, this PR:

- Adds a `deprecated` flag to the `attribute` decorator
- Modifies the `extraction_filter` logic, i.e. `Requires` to exclude deprecated attributes unless specifically wished by user
- Modifies the way the data extracted by the parser is passed back to the crawler (see Note)
- Modifies test cases to be compatible with this new representation of data
- Marks `topics` attribute of `Focus` as deprecated

*Note:* I have been thinking back and forth whether to modify the way the data is passed back from the parser or to propagate the `force_deprecated` flag. I decided to go for the first for two reasons: simpler `crawl` and subsequent functions and we would need a consider the `force_deprecated` flag within the `extraction_filter` in any case, to decide whether a missing attribute is deprecated and whether or not this is accepted.


MaxDall (commented):
@addie9800 Thanks a lot for taking the time to address this long-lasting issue 🙏 

This looks really good! But before I go into detail with my review, I would like to answer a question about the solution in general: Would an implementation, which includes a timestamp to deprecate attributes, and prevent them from being executed, rather than annotate the extraction, be more in line with our current design? What do you think?

addie9800:
> Would an implementation, which includes a timestamp to deprecate attributes, and prevent them from being executed, rather than annotate the extraction, be more in line with our current design?

That is a good point. It would automatically include backwards compatibility, i.e. still parsing the attributes automatically, when they still exist, which is a big bonus. 
I do wonder though, if that would require a modification of the base_parser concept, as we do the attribute collection outside the context of the article or if it could also be realized as a post-processing solution.

PR #746, opened 2025-04-24
Bug Fix in Source Restriction

addie9800:
This bug causes Fundus to crash, when `restrict_sources_to` contains a `URLSource` which is not supported by one of the publishers. This PR handles that case.


MaxDall (approved):
@addie9800 thanks a lot for catching and fixing this

PR #757, opened 2025-05-14
Add `RU`

addie9800:
This PR adds the first Russian Newspaper. Note that due to a bug in the native RobotsParser, you need to ignore the robots file to crawl this publisher.


MaxDall (dismissed):
@addie9800 Thanks a lot for adding this 🚀 I additionally added a temporary option to suppress the rules parsed from the `robots.txt` file for a specific publisher.

PR #760, opened 2025-05-22
Add `ZA`

addie9800:
This PR adds `ZA`


src/fundus/parser/data.py
```diff
@@ -540,7 +540,11 @@ class Image(DataclassSerializationMixin):
     position: int
 
     def __post_init__(self):
-        for url in [version.url for version in self.versions]:
+        for version in self.versions:
+            url = version.url
+            if url.startswith("//"):
+                url = "https:" + url
+                version.url = url
```
MaxDall: Can't we omit that part and set `relative_urls=True`?

src/fundus/parser/utility.py
```diff
@@ -189,16 +190,27 @@ def sanitize_json(text: str) -> Optional[str]:
...
+
+
 def parse_json(text: str) -> Optional[Dict[str, JSONVal]]:
     if not (json_content := sanitize_json(text)):
         return None
 
     try:
-        return cast(Dict[str, JSONVal], json.loads(json_content))
+        parsed_json = json.loads(json_content)
+        return cast(Dict[str, JSONVal], unescape_json_values(parsed_json))
```
MaxDall: Is this a major difference from #734? But anyway, we should either decide to go with one solution or merge them properly.

src/fundus/publishers/tw/__init__.py
```diff
@@ -1,5 +1,5 @@
 from fundus.publishers.base_objects import Publisher, PublisherGroup
-from fundus.publishers.tw.TaipeiTimes import TaipeiTimesParser
+from fundus.publishers.tw.taipei_times import TaipeiTimesParser
```
MaxDall, line 2: Thanks for catching that 👍

tests/resources/parser/test_data/jp/NikkanGeadai.json
```diff
@@ -74,7 +74,7 @@
       }
     ],
     "publishing_date": "2025-01-31 11:30:00+09:00",
-    "title": "フジテレビ問題でスポーツ界にも大激震！協賛企業&quot;総スカン&quot;で各競技団体のビジネスモデル完全崩壊へ",
+    "title": "フジテレビ問題でスポーツ界にも大激震！協賛企業\"総スカン\"で各競技団体のビジネスモデル完全崩壊へ",
```
MaxDall, line 77: Okay, there is something I'm missing here. Do you have an idea why this case is solved by this solution but not with #734?

MaxDall (changes requested):
@addie9800 Thanks a lot for adding this 🚀

src/fundus/parser/data.py
addie9800: Yes that is true. I guess I just assumed it wouldn't work, since it's not hosted on the same domain. But the tests go through without a problem

src/fundus/parser/utility.py
addie9800: There is one critical difference. Before unescaping, you need to ensure that the the quotation marks are handled appropriately. I have now merged the two solutions.

Update: this solution works better. I'll keep this for both cases

tests/resources/parser/test_data/jp/NikkanGeadai.json
addie9800: The problem here is, that they encode `&` as \u0026 which throws off the regex in #734, I think. But I am not entirely sure. I think it shouldn't but that's the only difference I see

src/fundus/parser/utility.py
MaxDall: Thanks for the explanation

MaxDall (approved):

PR #762, opened 2025-05-22
Add `LS`

addie9800:
This PR adds Lesotho to our `PublisherCollection`


MaxDall (approved):
@addie9800 Thanks a lot for adding `Lesotho` as a new country and `Lesotho Times` as it's first publisher, respectively. Looks good 🚀

PR #765, opened 2025-05-28
Add KR with MBN

zxxxv:
This PR introduces support for **Maeil Business Newspaper**, South Korean news source integrated into the dataset.

**Publisher:** _Maeil Business Newspaper_


src/fundus/publishers/kr/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+class KR(metaclass=PublisherGroup):
+    default_language = "kr"
+
+    MBN = Publisher(
+        name="MaeilBusinessNewspaper",
+        domain="https://www.mk.co.kr/",
+        parser=MBNParser,
+        sources=[
+            RSSFeed("https://www.mk.co.kr/rss/40300001/"),
+            Sitemap("https://www.mk.co.kr/sitemap/latest-articles/"),
```
addie9800, line 16: This seems to be what we would consider a ˋNewsMapˋ as it only contains the moste recent articles

addie9800: But, you can add this as a ˋSitemapˋ object as well. This contains a detailed overview of all the articles that mbn has released: https://www.mk.co.kr/sitemap/daily-articles/

src/fundus/publishers/kr/__init__.py
```diff
@@ -0,0 +1,18 @@
+from fundus.publishers.base_objects import Publisher, PublisherGroup
+from fundus.scraping.url import RSSFeed, Sitemap
+
+from .mbn import MBNParser
+
+
+class KR(metaclass=PublisherGroup):
+    default_language = "kr"
```
addie9800, line 8:
```suggestion
    default_language = "ko"
```
I think the ISO 639-1 language code for Korean should be ˋkoˋ

src/fundus/publishers/kr/mbn.py
```diff
@@ -0,0 +1,58 @@
...
+    generic_topic_parsing,
+    image_extraction,
+)
+
+
+class MBNParser(ParserProxy):
+    class V1(BaseParser):
+        _summary_selector = XPath("//div[@class='mid_title']//div")
+        _paragraph_selector = XPath(
+            "//div[@itemprop='articleBody']//p | //div[@itemprop='articleBody']//div[normalize-space(text())]"
```
addie9800, line 22: This _paragraph_selector also captures the summary, hence it is selected twice. As an example, you can observe: https://www.mk.co.kr/news/realestate/11330333

src/fundus/publishers/kr/mbn.py
```diff
@@ -0,0 +1,58 @@
...
+
+        @attribute
+        def images(self) -> List[Image]:
+            return image_extraction(
+                doc=self.precomputed.doc,
+                paragraph_selector=self._paragraph_selector,
+                upper_boundary_selector=XPath("//div[@itemprop='articleBody']"),
+                image_selector=XPath("//div[@itemprop='articleBody']//div[@class='thumb_area img']//img"),
+                caption_selector=XPath("./ancestor::div[@class='thumb_area img']//span[@class='thum_figure_txt']"),
+                author_selector=re.compile(r"(?!.*\.)(?P<credits>.*)"),
```
addie9800, line 57: In this article, the image author is credited to ˋ 사진설명ˋ which seems to be a mistake. Maybe as background for you, if there is no caption the regex is applied to the description.

addie9800: Also, MBN sometimes encloses the image authors with square brackets, as they did in this article: https://www.mk.co.kr/news/society/11330332 . They should also be removed, ideally using the ˋauthor_selectorˋ. It would also be good, if ˋ사진 =ˋ were removed by the selector.

src/fundus/publishers/kr/mbn.py
```diff
@@ -0,0 +1,58 @@
...
+    extract_article_body_with_selector,
+    generic_author_parsing,
+    generic_date_parsing,
+    generic_topic_parsing,
+    image_extraction,
+)
+
+
+class MBNParser(ParserProxy):
+    class V1(BaseParser):
```
addie9800, line 19: Unfortunately, it seems like MBN is using multiple site layouts for different subsections. As of now, the parser cannot parse the economy section: https://www.mk.co.kr/economy/view/2025/366669

src/fundus/publishers/kr/mbn.py
```diff
@@ -0,0 +1,58 @@
...
+        @attribute
+        def publishing_date(self) -> Optional[datetime]:
+            return generic_date_parsing(self.precomputed.ld.get_value_by_key_path(["NewsArticle", "datePublished"]))
+
+        @attribute
+        def title(self) -> Optional[str]:
+            return self.precomputed.ld.get_value_by_key_path(["NewsArticle", "headline"])
+
+        @attribute
+        def topics(self) -> List[str]:
```
addie9800, line 46: I feel like the section the article is published in, is a bit too vague as a topic and since the website does not yet provide any good alternatives, I would probably suggest removing ˋtopicsˋ support for this publisher altogether.

src/fundus/publishers/kr/mbn.py
addie9800: > the parser cannot parse the economy section

It seems like they use a very unconventional formatting for the economy section, which can be quite 
a hassle to deal with manually. We have had 2 similar cases in the past and have established the following helper function, which might make your life easier: `transform_breaks_to_paragraphs`. If you want to have a look at an example, it is used in ˋIlGriornaleˋ and NikkanGeadaiˋ.

src/fundus/publishers/kr/mbn.py
```diff
@@ -0,0 +1,58 @@
...
+        def body(self) -> Optional[ArticleBody]:
+            return extract_article_body_with_selector(
+                self.precomputed.doc,
+                paragraph_selector=self._paragraph_selector,
+                summary_selector=self._summary_selector,
+            )
+
+        @attribute
+        def authors(self) -> List[str]:
+            return generic_author_parsing(self.precomputed.ld.get_value_by_key_path(["NewsArticle", "author"]))
```
addie9800, line 35:
```suggestion
            return generic_author_parsing(self.precomputed.ld.xpath_search("//NewsArticle//author"))
```
Please replace all instances of `get_value_by_key_path` with `xpath_search` as the first is deprecated and going to be removed in one of the newer versions (then there will also not be any issues with the testing). The switch is pretty straightforward, you can just prepend all values of the list with a double forward slash and combine them into one string as the XPath.

.github/workflows/tests.yml
```diff
@@ -2,7 +2,7 @@ name: tests
 
 on:
   push:
-    branches: [ master ]
```
addie9800, line 5: This file should remain unchanged. With the changes described below, you should also no longer have any issues with Warnings.

addie9800 (changes requested):
Thank you so much for adding our first Korean publisher to Fundus 🚀. As a matter of fact, I have the next one lined up, as soon as this one is merged. I have a couple of requests, that need to be addressed and then this PR is ready to go.

zxxxv:
I haven’t been able to add the economy section yet, and I’ve noticed there’s an error in the image handling. I’ll go ahead and fix it.

src/fundus/publishers/kr/mbn.py
zxxxv: Fixed in effc939

zxxxv: Additionally, I’ve fixed the <br>-related issue using `transform_breaks_to_paragraphs`.

zxxxv:
I believe I’ve addressed all of your requests—could you please take another look when you have a chance? (～￣▽￣)～

addie9800 (approved):
Hey, I'm sorry I missed your comment asking me to take a look again. I still noticed issues, when parsing this article: https://www.mk.co.kr/economy/view/2025/366669, which is why I added some modification. Apart from that, this looks really good. I'm going to go ahead and approve and merge this PR. Thank you so much for adding this. I am now also going go ahead and finalize the other South Korean publisher, in case you need it :)

PR #768, opened 2025-05-29
Add Spanish publishers

Finiluh:
Add elDiario.es and Publico


src/fundus/publishers/es/publico.py
```diff
@@ -0,0 +1,62 @@
...
+    generic_author_parsing,
+    generic_date_parsing,
+    generic_topic_parsing,
+    image_extraction,
+)
+
+
+class PublicoParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = XPath("//div[contains(@class, 'body-modules w-full')]//p")
```
addie9800, line 19: In this article the authors of the articles that are in the middle of the article are parsed as well: https://www.publico.es/economia/sondeo-bce-bajara-tipos-5-junio-pero-julio.html For example Agencias

addie9800 (changes requested):
This looks great. I only have one minor comment and then this is ready to go :)

src/fundus/publishers/es/publico.py
Finiluh: _paragraph_selector updated

addie9800 (approved):
This looks great. Thanks for adding it 🚀

PR #772, opened 2025-06-12
Add `skip_publishers_disallowing_training`

addie9800:
This pull request introduces functionality to filter publishers based on their policies regarding AI training and enhances error handling in the crawling process. 

### AI Training Filtering Enhancements:
* **README and Documentation Updates**: Added a disclaimer in `README.md` and detailed instructions in `docs/5_advanced_topics.md` about verifying publisher policies before using their content for AI training. Introduced a new section, "Filtering publishers for AI training," explaining the `skip_publishers_disallowing_training` parameter. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R28-R32) [[2]](diffhunk://#diff-388fd61818e9122fb9c4fdacba77e888c402e917d98be3fd757023123a57306fR7) [[3]](diffhunk://#diff-388fd61818e9122fb9c4fdacba77e888c402e917d98be3fd757023123a57306fR37-R43)
* **`robots.txt` Parsing**: Enhanced the `CustomRobotFileParser` class in `src/fundus/publishers/base_objects.py` to detect AI training disallowance keywords in comments and added the `disallow_training` attribute to indicate such restrictions. The `Robots` class was also modified to self-ensure the readyness of the Robots File Parser instead of requiring us to check if the parser is ready and calling the `read()` function if necessary. 
* **Publisher Object Updates**: Added a `disallows_training` attribute to the `Publisher` class so that known disallowances can be hardcoded into the `Publisher` setup. [[1]](diffhunk://#diff-c632716248f3f842390bbd6269b98aa9170863e65c162ca7fe83c06ff3869e29R137) [[2]](diffhunk://#diff-c632716248f3f842390bbd6269b98aa9170863e65c162ca7fe83c06ff3869e29L111-R169)

### Crawling Logic Improvements:
* **Skipping Restricted Publishers**: Updated the `crawl()` and `_fetch_articles()` methods in `src/fundus/scraping/crawler.py` to support the `skip_publishers_disallowing_training` parameter, ensuring publishers with AI training restrictions are excluded. [[1]](diffhunk://#diff-284dbcb0188b7c0d168b0de668707fec3b2c1e68fbe26a93a9a0f800f3891cf0R241) [[2]](diffhunk://#diff-284dbcb0188b7c0d168b0de668707fec3b2c1e68fbe26a93a9a0f800f3891cf0R479) [[3]](diffhunk://#diff-284dbcb0188b7c0d168b0de668707fec3b2c1e68fbe26a93a9a0f800f3891cf0R632-R644)
* **Error Handling for Interruptions**: A previous change solving the deadlock issue by calling `task_done()` technically solved the problem, but only by causing the thread to crash. While this did have the intended effect, it is a messy solution. The new solution introduces a `RequestInterruptedError` exception in `src/fundus/scraping/session.py` and updates the crawling logic to handle interrupted requests gracefully. [[1]](diffhunk://#diff-2da385ff8c2e9039263d3444ad5b924f921a951f156efc3be3a85809b8ab149fR18-R23) [[2]](diffhunk://#diff-2da385ff8c2e9039263d3444ad5b924f921a951f156efc3be3a85809b8ab149fL28-R44) [[3]](diffhunk://#diff-b15ba138c72e9c6446eb54088a357823cf2d80a06450f1aa5a5ca179ec3321f4L194-R195)

These updates provide better compliance with publisher policies and improve the robustness of the crawling process.

### DIFF
--- a/src/fundus/scraping/crawler.py
+++ b/src/fundus/scraping/crawler.py
@@ -58,7 +58,7 @@ from fundus.scraping.delay import Delay
 from fundus.scraping.filter import ExtractionFilter, Requires, RequiresAll, URLFilter
 from fundus.scraping.html import CCNewsSource
 from fundus.scraping.scraper import CCNewsScraper, WebScraper
-from fundus.scraping.session import CrashThread, SessionHandler, session_handler
+from fundus.scraping.session import CrashThread, session_handler
 from fundus.scraping.url import URLSource
 from fundus.utils.events import __EVENTS__
 from fundus.utils.timeout import Timeout


src/fundus/scraping/crawler.py
```diff
@@ -481,6 +492,24 @@ def constant_delay() -> float:
             else:
                 raise TypeError("param <delay> of <Crawler.__init__>")
 
+        # register default events
+        __EVENTS__.register_event("stop")
```
MaxDall: I have to admit, I'm not entirely comfortable with moving the event register here. It feels somewhat disconnected from the rest of the code. While I understand the necessity behind this choice, it does strike me as a potential design flaw that we should aim to address.

More broadly, enforcing strict access to events, requiring them to be registered before they can be checked, might not be the best approach. The original intention was to introduce a level of control, ensuring that accessing an unregistered event would result in an error. However, I'm not sure whether this tradeoff is worth it. What do you think?

src/fundus/scraping/crawler.py
```diff
@@ -597,10 +628,19 @@ def _fetch_articles(
...
         retries: int = 0
         while True:
+            if skip_publishers_disallowing_training:
+                publishers = tuple(
+                    [
+                        publisher
+                        for publisher in publishers
+                        if not (publisher.disallows_training or publisher.robots.disallows_training())
+                    ]
+                )
```
MaxDall: Performing this check for every WARC path introduces unnecessary overhead. It would be far more efficient to verify the publishers once during initialization and then propagate the result. However, this should be implemented in a way that doesn’t significantly slow down the `CCNewsCrawlers` initialization (i.e., not running sequentially) and doesn’t depend on the availability of each publisher’s `robots.txt`.

src/fundus/scraping/session.py
```diff
@@ -53,15 +62,13 @@ def _req():
...
                 response = response_queue.get(timeout=1)
-            except Empty:
-                if __EVENTS__.is_event_set("stop"):
-                    logger.debug(f"Interrupt request for {url!r}")
-                    response_queue.task_done()
-                    exit(1)
-            else:
                 if isinstance(response, Exception):
                     raise response
                 return response
```
MaxDall: What's the reason to add that to the try-except block?

src/fundus/scraping/session.py
```diff
@@ -15,6 +15,12 @@
 _default_header = {"user-agent": "Fundus"}
 
 
+class RequestInterruptedError(Exception):
+    """Raised when a request is interrupted by a stop event."""
+
+    pass
```
MaxDall: I use this part of the code to formulate my general concerns.

1. Thanks a lot for the stability improvements. Getting threading under control seems to be a real hassle.
2. I would appreciate implementing a best practice to keep pull requests functionally isolated whenever possible. Combining new crawler functionality with stability improvements (like threading changes) makes the review process significantly harder, and now the features depend on each other.
3. In general, I prefer your solution, introducing a custom error for the interrupt and catching it when possible to control code flow and refrain from intentionally crashing threads with uncought errors. But I see a problem here, as it would require us to handle interruptions at every potential iteration point.

For example, take this part of the `BaseScraper`:

```python
for html in source.fetch(url_filter=url_filter):
    parser = self.parser_mapping[html.source_info.publisher]
```
To prevent it from looping over every source (which it would do now), we would have to check for the stop event here as well, introducing the flag to the scope of the `CCNewsCrawler`, which I would like to prevent.
Having this in mind, I would prefer crashing the thread intentionally by raising uncought `RequestInterruptedError`s, but what's your opinion on this?

MaxDall (changes requested):

src/fundus/scraping/session.py
addie9800: It was probably included there by accident.

src/fundus/scraping/crawler.py
addie9800: Can you perhaps give me an example, where it might cause problems accessing unregistered events? Because especially with regard to the stop event, I feel like the process of registering is devalued by the fact that, we do it by default for all threads when crawling. I could imagine in that it becomes more important for future events, that could be added. So I would also tend towards finding a solution that perhaps automatically registers the stop event, because we will likely need it in most situations.

src/fundus/scraping/crawler.py
addie9800: Did you already have an idea of how we would implement that? If we want to have it non-blocking, I have one idea for now. During initialization of the CCNewsCrawler, we start getting the publishers preferences in the background. We can then start crawling as normally, especially since we have to stream the entire WARC file anyway. We can then already start parsing articles for the supported publishers and either caching them until we have the results from the permission check or start yielding/discarding them, if we have the results. I don't really see a way to do that without crawling the robots.txt. I guess caching the preferences might also be an option, but I don't know, if that makes too much sense, since we'll have to crawl it at least once anyway and if people want to comply with the preferences, I think we should ensure they are up to date.

src/fundus/scraping/session.py
addie9800: You are definitely correct, sorry for the extra confusion with the combined PR. I would also very much appreciate to have them separate at the moment.

While I see the issue you raised, I'm not sure I prefer the alternative of purposely crashing the thread. Maybe there is some alternative to handle it a bit more gracefully? What is your opinion on a similar solution as in the queue setting, where we pass the exception back through the queue, we could do something similar for the iterator. Then we would avoid reusing the flag out of scope, but we can cleanly exit the loop with a break, instead of crashing out.

addie9800:
Realization Ideas:
- Register Default Events like 'stop'
- CrashThread Exception
- Move publisher robots testing away from _fetch_articles in CCNewsCrawler, ideally in background

addie9800:
> There’s just one remaining issue: the try-catch and future timeout logic for the skip publisher part in `CCNewsCrawler` isn’t working as intended.

Could you maybe elaborate on what exactly is not working as intended? I have tested it locally using the following code: 

```python
from logging import DEBUG

from fundus import PublisherCollection, CCNewsCrawler
from fundus.logging import set_log_level

from datetime import datetime

crawler = CCNewsCrawler(PublisherCollection.uk, start=datetime(2025, 10, 1))
set_log_level(DEBUG)

if __name__ == "__main__":
    for article in crawler.crawl(max_articles=10, error_handling="suppress", skip_publishers_disallowing_training=True):
        print(article)
```

and I get the following logs:

```txt
Loading WARC Paths:   0%|          | 0/1 [00:00<?, ?it/s]
2025-10-23 14:21:37,767 - fundus.scraping.session - DEBUG - Creating new session
2025-10-23 14:21:37,968 - fundus.scraping.session - DEBUG - (200) <GET 'https://www.dailymail.co.uk/robots.txt'> took 0.174027 second(s)
2025-10-23 14:21:38,026 - fundus.scraping.session - DEBUG - (200) <GET 'https://www.independent.co.uk/robots.txt'> took 0.22954 second(s)
Loading WARC Paths: 100%|██████████| 1/1 [00:00<00:00,  3.46it/s]
2025-10-23 14:21:38,050 - fundus.scraping.session - DEBUG - (200) <GET 'https://inews.co.uk/robots.txt'> took 0.252384 second(s)
2025-10-23 14:21:38,063 - fundus.scraping.session - DEBUG - (200) <GET 'https://www.standard.co.uk/robots.txt'> took 0.271241 second(s)
2025-10-23 14:21:38,068 - fundus.scraping.session - DEBUG - (200) <GET 'https://www.theguardian.com/robots.txt'> took 0.261279 second(s)
2025-10-23 14:21:38,071 - fundus.scraping.crawler - WARNING - Skipping publisher 'The Guardian' because it disallows training.
2025-10-23 14:21:38,082 - fundus.scraping.session - DEBUG - (200) <GET 'https://www.express.co.uk/robots.txt'> took 0.290059 second(s)
2025-10-23 14:21:38,174 - fundus.scraping.session - DEBUG - (200) <GET 'https://www.euronews.com/robots.txt'> took 0.372205 second(s)
2025-10-23 14:21:38,277 - fundus.scraping.session - DEBUG - (200) <GET 'https://www.telegraph.co.uk/robots.txt'> took 0.475296 second(s)
2025-10-23 14:21:38,289 - fundus.scraping.session - DEBUG - (200) <GET 'https://www.dailystar.co.uk/robots.txt'> took 0.489381 second(s)
2025-10-23 14:21:38,317 - fundus.scraping.session - DEBUG - (200) <GET 'https://www.bbc.co.uk/robots.txt'> took 0.514271 second(s)
2025-10-23 14:21:38,320 - fundus.scraping.session - DEBUG - (200) <GET 'https://www.mirror.co.uk/robots.txt'> took 0.518113 second(s)
2025-10-23 14:21:38,386 - fundus.scraping.session - DEBUG - (200) <GET 'https://metro.co.uk/robots.txt'> took 0.413273 second(s)
2025-10-23 14:21:38,389 - fundus.scraping.session - DEBUG - (200) <GET 'https://www.thesun.co.uk/robots.txt'> took 0.598786 second(s)
2025-10-23 14:21:38,395 - fundus.scraping.crawler - WARNING - Skipping publisher 'The Sun' because it disallows training.
```
where all publishers are checked, and the two publishers that should be blocked are also blocked.

PR #774, opened 2025-06-25
Add logging for source restriction

addie9800:
```
crawler = Crawler(PublisherCollection.us.FreeBeacon, restrict_sources_to=[NewsMap, RSSFeed])
for article in crawler.crawl(max_articles=10, only_complete=False):
    print(article)
```
The code above will exit without any output with the current Fundus version. This leads to quite a headache, when debugging, which is why I propose adding a warning-level log if a publisher is found to not have any sources that fit the specified filters.


MaxDall (approved):
@addie9800 Thanks for addressing this 👍

PR #775, opened 2025-06-30
Add `LU`

addie9800:


src/fundus/publishers/lu/tageblatt.py
```diff
@@ -0,0 +1,57 @@
...
+    extract_article_body_with_selector,
+    generic_author_parsing,
+    generic_date_parsing,
+    image_extraction,
+)
+
+
+class TageblattParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = XPath("//div[@class='text-content']/p[@class='text']")
```
MaxDall: In [this article](https://www.tageblatt.lu/headlines/so-will-luxemburg-kuenftig-gegen-paedo-hunting-vorgehen/) some boilerplate is extracted at the end.

[Here](https://www.tageblatt.lu/headlines/ein-gesetzesprojekt-fuer-die-muelltonne-aenderungen-am-platzverweis-renforce-unzureichend/) as well.

MaxDall (changes requested):
@addie9800 Thanks a lot for adding our first publisher from Luxembourg 🚀 
I found some boilerplate in the extraction, but besides that, it looks excellent.
I also reabsed the branch instead of merging, so before working on it, you should fetch it again locally.

src/fundus/publishers/lu/tageblatt.py
addie9800: Ah, thanks. I took care of it :)

PR #776, opened 2025-06-30
Add `LuxemburgerWort`

addie9800:


src/fundus/publishers/lu/luxemburger_wort.py
```diff
@@ -0,0 +1,60 @@
...
+
+        @attribute
+        def images(self) -> List[Image]:
+            return image_extraction(
+                doc=self.precomputed.doc,
+                paragraph_selector=self._paragraph_selector,
+                upper_boundary_selector=CSSSelector("h1"),
+                caption_selector=XPath("./ancestor::figure//div[contains(@class, 'ImageCaption')]"),
+                author_selector=re.compile(r"(?i)Foto:\s*(?P<credits>.*)"),
+            )
```
MaxDall: In [this article](https://www.wort.lu/panorama/tv-fuer-hunde-was-vierbeiner-vor-dem-bildschirm-erleben/78617450.html), the little preview image of the related article is extracted as well.

MaxDall (changes requested):
@addie9800 Thanks a lot for adding that!

src/fundus/publishers/lu/luxemburger_wort.py
addie9800: Thanks for catching that, seems like I also missed that in the test case :/

MaxDall (approved):

PR #788, opened 2025-09-10
Update user-agent

addie9800:
As iNews is blocking us from accessing their sitemap, I looked for the issue and it seems like the issue is our user-agent. Hence I updated our default user-agent to be a bit more descriptive and also provide a way to contact us.

### DIFF
--- a/src/fundus/scraping/session.py
+++ b/src/fundus/scraping/session.py
@@ -12,7 +12,7 @@ from fundus.utils.events import __EVENTS__
 
 logger = create_logger(__name__)
 
-_default_header = {"user-agent": "Fundus"}
+_default_header = {"user-agent": "Fundus/2.0 (contact: github.com/flairnlp/fundus)"}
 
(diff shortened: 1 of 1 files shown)


src/fundus/scraping/session.py
```diff
@@ -12,7 +12,7 @@
 
 logger = create_logger(__name__)
 
-_default_header = {"user-agent": "Fundus"}
+_default_header = {"user-agent": "Fundus/2.0 (contact: github.com/flairnlp/fundus)"}
```
MaxDall, line 15: Just out of curiosity: Could you pinpoint what exactly was the problem with `iNews`? The missing version?

addie9800: I actually don't know for sure. I tried a standard User-Agent which worked and then I asked ChatGpt what a suitable User-Agent for us would be. And that one worked immediately.

PR #790, opened 2025-09-14
Update sources ThePortugalNews

addie9800:
The sitemap URLs are outdated. I have now also added the sources for various available languages.


MaxDall (approved):
@addie9800 Thanks so much for adding all these sources! Just one small nitpick—it’s not a blocker for this PR, but I noticed that when someone crawls this publisher, they first get all the English articles, and so on, which makes the articles a bit out of order date-wise.

PR #791, opened 2025-09-14
Update blocklist for Publisher Coverage

addie9800:


.github/workflows/publisher_coverage.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
         # We skip the Kicker APNews publishers, because they are IP-blocked
         run: |
           set -o pipefail
-          timeout 25m python -u scripts/publisher_coverage.py --skip Kicker APNews | tee publisher_coverage.txt
+          timeout 25m python -u scripts/publisher_coverage.py --skip Kicker APNews Tageblatt | tee publisher_coverage.txt
```
MaxDall, line 35: Hmm i have the feeling i missed something: Why aren't we just deprecate this Publishers? Or is it that they IP blocked the GitHub server?

addie9800: > Or is it that they IP blocked the GitHub server?

Exactly. The crawling works normally, when running locally.

PR #792, opened 2025-09-22
Add tag filter to text extraction

MaxDall:
This PR implements functionality to omit tags during text extraction.


addie9800 (commented):
This is a very elegant solution. I like it 👍

PR #796, opened 2025-10-15
Fix race conditions and rework exception handling

MaxDall:
This PR implements some proposals for #772 
- Fix a race condition where events where registered after the Pool already tries to shutdown
- Fix a bug where events weren't registered when aliased and therefore not considered with `set_for_all`
- Makes `_events` mapping bi-directional
- Reworks `EventDict` and `ThreadEventDict` default logic
- Adds missing doc strings to `EventDict` and `ThreadEventDict`


addie9800 (approved):
This looks really good. Especially the extra documentation is a nice addition. 👍

PR #798, opened 2025-10-19
New publisher added: Der Freitag

bresslem:


src/fundus/publishers/de/der_freitag.py
```diff
@@ -0,0 +1,52 @@
...
+        def publishing_date(self) -> Optional[datetime]:
+            return generic_date_parsing(self.precomputed.ld.bf_search("datePublished"))
+
+        @attribute
+        def topics(self) -> List[str]:
+            return self.precomputed.ld.bf_search("keywords")
+
+        @attribute
+        def images(self) -> List[Image]:
+            return image_extraction(
```
addie9800, line 49: This seems to be too broad. In this [article](https://www.freitag.de/autoren/der-freitag/geplatzte-gruene-traeume-warum-sich-wasserstoff-nicht-lohnt) for example, the title picture is extracted and 3 other images that are only thumbnails for other articles. Also, the caption includes the author, whereas the image author is not selected at all.

addie9800:
Thank you very much for addressing the comments and opening an issue about the lazy loading problem. From my perspective, unfortunately, there is no quick fix to it without using some sort of headless browser, which would also require quite some restructuring. If @MaxDall does not have a better idea, I would go ahead and approve this PR, since this is nothing that can be fixed within this scope. Thanks for taking the time and adding this publisher 🚀

addie9800 (approved):
Thanks for adding Der Freitag!

PR #800, opened 2025-10-20
Add SE Expressen

ghostsshadow:


src/fundus/publishers/se/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+
+    Expressen = Publisher(
+        name="Expressen",
+        domain="https://www.expressen.se/",
+        parser=ExpressenParser,
+        sources=[
+            RSSFeed("https://feeds.expressen.se/nyheter/"),
+            RSSFeed("https://feeds.expressen.se/sport/"),
+            RSSFeed("https://feeds.expressen.se/noje/"),
+            Sitemap("https://www.expressen.se/sitemap.xml"),
```
addie9800: You should add `reverse=True` to crawl the articles in order.

src/fundus/publishers/se/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class SE(metaclass=PublisherGroup):
+    default_language = "sv"
+
+    Expressen = Publisher(
+        name="Expressen",
+        domain="https://www.expressen.se/",
+        parser=ExpressenParser,
+        sources=[
```
addie9800, line 15: There are also some RSSFeeds missing from here: https://www.expressen.se/om-expressen/rss-smidigt-satt-att-fa-nyheter-fran-expressen-9/

src/fundus/publishers/se/expressen.py
```diff
@@ -0,0 +1,54 @@
...
+        def publishing_date(self) -> Optional[datetime.datetime]:
+            return generic_date_parsing(self.precomputed.ld.bf_search("datePublished"))
+
+        @attribute
+        def authors(self) -> List[str]:
+            return generic_author_parsing(self.precomputed.ld.bf_search("author"))
+
+        @attribute
+        def images(self) -> List[Image]:
+            return image_extraction(
```
addie9800, line 47: You should refine the caption selector. As of now, it selects the image author as well: https://www.expressen.se/nyheter/varlden/ryssarnas-nya-hot-fran-norra--ishavet-flygande-tjernobyl/

src/fundus/publishers/se/expressen.py
```diff
@@ -0,0 +1,54 @@
...
+                doc=self.precomputed.doc,
+                paragraph_selector=self._paragraph_selector,
+                image_selector=XPath("//figure//img"),
+                upper_boundary_selector=CSSSelector("div.article__body-text"),
+            )
+
+        @attribute
+        def topics(self) -> List[str]:
+            return generic_topic_parsing(self.precomputed.ld.bf_search("articleSection"))
```
addie9800: We would consider this too broad in general. It would be better to take the `keywords` from the `ld_json` with some post-processing to remove everything before the `/`.

src/fundus/publishers/se/__init__.py
```diff
@@ -0,0 +1,20 @@
+from fundus.publishers.base_objects import Publisher, PublisherGroup
+from fundus.scraping.url import NewsMap, RSSFeed, Sitemap
+
+from .expressen import ExpressenParser
+
+
+class SE(metaclass=PublisherGroup):
+    default_language = "sv"
+
+    Expressen = Publisher(
```
addie9800, line 11: We should add a `url_filter` here, filtering out `/tv/`, since they only contain videos and cannot be parsed anyway.

src/fundus/publishers/se/expressen.py
```diff
@@ -0,0 +1,54 @@
...
+    generic_date_parsing,
+    generic_topic_parsing,
+    image_extraction,
+)
+
+
+class ExpressenParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = CSSSelector("div.article__body-text p")
+        _summary_selector = CSSSelector("div.article__preamble")
```
addie9800, line 21: You're missing a subheadline selector here. This [article](https://www.expressen.se/nyheter/sverige/elever-atalas-for-misshandel-pa-lundsberg/), for example, has a subheadline. _When you add a subheadline selector, please also change the testcase to an article with subheadlines_

src/fundus/publishers/se/expressen.py
```diff
@@ -0,0 +1,54 @@
...
+        def body(self) -> Optional[ArticleBody]:
+            return extract_article_body_with_selector(
+                self.precomputed.doc,
+                paragraph_selector=self._paragraph_selector,
+                summary_selector=self._summary_selector,
+            )
+
+        @attribute
+        def title(self) -> Optional[str]:
+            return self.precomputed.meta.get("og:title")
```
addie9800: It turns out that the meta tag is unreliable. In this [article](https://www.expressen.se/nyheter/politik/bohlin-till-attack-mot-s-hon-forlorar/), the title in the meta tag does not coincide with the actual title of the article. You should switch to the title named in the JSON

src/fundus/publishers/se/__init__.py
addie9800: We can also filter urls with `expressen-direkt`, which are live tickers and not yet supported by Fundus

src/fundus/publishers/se/expressen.py
ghostsshadow: Did you mean refinement like this?
```python
caption_selector=XPath("./ancestor::figure//figcaption//div[@class='rich-image__description']"),
author_selector=XPath("./ancestor::figure//figcaption//div[@class='rich-image__credit']"),
```

addie9800: Yes, exactly 👍

PR #801, opened 2025-10-21
add Rest Of World Publisher

marten-ti:
Added Rest Of World Publisher.

Rest of World is a nonprofit news organization focused on global technology stories, headquartered in the US but covering tech developments worldwide (e.g., Asia, Latin America, Africa). They publish articles on topics like AI, startups, policy, and innovation in emerging markets.

### DIFF
--- a/src/fundus/publishers/us/rest_of_world.py
+++ b/src/fundus/publishers/us/rest_of_world.py
@@ -9,6 +9,7 @@ from fundus.parser.utility import (
     extract_article_body_with_selector,
     generic_author_parsing,
     generic_date_parsing,
+    generic_topic_parsing,
     image_extraction,
 )
 
@@ -42,8 +43,7 @@ class RestOfWorldParser(ParserProxy):
 
         @attribute
         def topics(self) -> List[str]:
-            keywords: List[str] = self.precomputed.ld.xpath_search("NewsArticle/keywords")
-            return [keyword for keyword in keywords if not keyword.startswith("Subject: ")]
+            return generic_topic_parsing(self.precomputed.ld.xpath_search("NewsArticle/keywords", scalar=True))
 
         @attribute
         def images(self) -> List[Image]:


src/fundus/publishers/us/__init__.py
```diff
@@ -274,3 +275,13 @@ class US(metaclass=PublisherGroup):
...
         ],
     )
+
+    RestOfWorld = Publisher(
+        name="Rest of World",
+        domain="https://restofworld.org/",
+        parser=RestOfWorldParser,
+        sources=[
+            RSSFeed("https://restofworld.org/feed/latest/"),
+            Sitemap("https://restofworld.org/sitemap.xml"),
```
addie9800, line 286: In order to refrain from crawling unnecessary pages, we should add a `url_filter` filtering something along the lines of `restofworld.org/20xx/`

src/fundus/publishers/us/rest_of_world.py
```diff
@@ -0,0 +1,61 @@
...
+    extract_article_body_with_selector,
+    generic_author_parsing,
+    generic_date_parsing,
+    image_extraction,
+)
+
+
+class RestOfWorldParser(ParserProxy):
+    class V1(BaseParser):
+        VALID_UNTIL = date.today()
```
addie9800: This can be omitted for now, since this is the default value.

src/fundus/publishers/us/rest_of_world.py
```diff
@@ -0,0 +1,61 @@
...
+            keywords: List[str] = self.precomputed.ld.xpath_search("NewsArticle/keywords")
+            return [keyword for keyword in keywords if not keyword.startswith("Subject: ")]
+
+        @attribute
+        def images(self) -> List[Image]:
+            return image_extraction(
+                doc=self.precomputed.doc,
+                paragraph_selector=self._paragraph_selector,
+                image_selector=XPath("//figure//img | //img[@src or @data-src]"),
+                caption_selector=XPath("./ancestor::figure//*[self::figcaption or contains(@class, 'caption')]"),
```
addie9800: The captions seem to contain the image author as well. You have two options here: either refine the XPath selector to only select the caption or capture the author using a regex from the caption + author, then it is also removed in the backend.

addie9800 (changes requested):
Thank you for adding `RestOfWorld`. This looks really good already. There are only a couple of minor changes that should be addressed.

PR #802, opened 2025-10-22
Add VN publisher (VnExpress)

bachthyaglx:
Hi, I’ve added a new VN publisher (VnExpress).
Please review when you have time. Thank you!

### DIFF
--- a/src/fundus/publishers/vn/vnexpress.py
+++ b/src/fundus/publishers/vn/vnexpress.py
@@ -50,7 +50,7 @@ class VnExpressIntlParser(ParserProxy):
             return image_extraction(
                 doc=self.precomputed.doc,
                 paragraph_selector=self._paragraph_selector,
-                author_selector=re.compile(r"(?i)Ảnh:\s*(?P<credits>.+)$"),
+                author_selector=re.compile(r"(?i)(Ảnh:|Video:)\s*(?P<credits>.+)$"),
                 upper_boundary_selector=XPath("//h1"),
             )
 


src/fundus/publishers/vn/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+  default_language = "vi"
+
+  VnExpress = Publisher(
+    name="VnExpress",
+    domain="https://vnexpress.net/",
+    parser=VnExpressIntlParser,
+    sources=[
+      RSSFeed("https://vnexpress.net/rss/tin-moi-nhat.rss"),
+      Sitemap("https://vnexpress.net/sitemap.xml"),
+      NewsMap("https://vnexpress.net/google-news-sitemap.xml"),
```
addie9800: Seems like they tricked you with the sitemap links, They all redirect you to home

addie9800: Instead you can add the other RSSFeeds from here: https://vnexpress.net/rss as sources

src/fundus/publishers/vn/vnexpress.py
```diff
@@ -0,0 +1,108 @@
...
+
+class VnExpressIntlParser(ParserProxy):
+  class V1(BaseParser):
+    _summary_selector = CSSSelector("p.description")
+    _paragraph_selector = CSSSelector("article.fck_detail > p")
+    _subheadline_selector = CSSSelector("article.fck_detail > h2")
+      
+    @attribute
+    def title(self) -> Optional[str]:
+      title_list: List[Any] = self.precomputed.ld.xpath_search("//NewsArticle/headline")
```
addie9800: You can use `scalar=True` here, and it will not return a List. Have you observed `self.precomputed.ld.xpath_search("//NewsArticle/headline")` to be unreliable? Usually, relying on the JSON should be sufficient.

src/fundus/publishers/vn/vnexpress.py
```diff
@@ -0,0 +1,108 @@
...
+      title_nodes = CSSSelector("h1.title-detail")(self.precomputed.doc)
+      if title_nodes:
+        title_text: str = title_nodes[0].text_content().strip()
+        return title_text
+      
+      return None
+    
+    @attribute
+    def authors(self) -> List[str]:
+      author_data_list: List[Any] = self.precomputed.ld.xpath_search("//NewsArticle/author")
```
addie9800: You can pass in whatever you get back directly into `generic_author_parsing`. It is designed to work with various inputs. Have you observed self.precomputed.ld.xpath_search("//NewsArticle/author") to be unreliable? Usually, relying on the JSON should be sufficient.

src/fundus/publishers/vn/vnexpress.py
```diff
@@ -0,0 +1,108 @@
...
+      
+      author_nodes = CSSSelector("p.author_mail strong")(self.precomputed.doc)
+      if author_nodes:
+        return [node.text_content().strip() for node in author_nodes if node.text_content().strip()]
+      
+      return []
+    
+    @attribute
+    def publishing_date(self) -> Optional[datetime]:
+      date_list: List[Any] = self.precomputed.ld.xpath_search("//NewsArticle/datePublished")
```
addie9800: Here, you can use `scalar=True` as well. And it should be sufficient to use the JSON value.

src/fundus/publishers/vn/vnexpress.py
```diff
@@ -0,0 +1,108 @@
...
+      if section and isinstance(section, str):
+        return [section]
+      meta_keywords = self.precomputed.meta.get("keywords")
+      if meta_keywords and isinstance(meta_keywords, str):
+        return [k.strip() for k in meta_keywords.split(',') if k.strip()]
+      return []
+
+    @attribute
+    def topics(self) -> List[str]:
+      ld_topics = self._parse_ld_keywords()
```
addie9800: You can simplify this greatly by just using `generic_topic_parsing(self.precomputed.meta.get("keywords")`, which essentially does the same thing your custom helper methods do.

src/fundus/publishers/vn/vnexpress.py
```diff
@@ -0,0 +1,108 @@
...
+from fundus.parser import ArticleBody, BaseParser, ParserProxy, attribute
+from fundus.parser.utility import (
+  extract_article_body_with_selector,
+  generic_author_parsing,
+  generic_date_parsing,
+)
+
+
+class VnExpressIntlParser(ParserProxy):
+  class V1(BaseParser):
```
addie9800: The `images` attribute seems to be missing in this parser.

src/fundus/publishers/vn/vnexpress.py
```diff
@@ -0,0 +1,108 @@
...
+    _summary_selector = CSSSelector("p.description")
+    _paragraph_selector = CSSSelector("article.fck_detail > p")
```
addie9800: In this [article](https://vnexpress.net/nam-phi-coi-viet-nam-la-nguoi-ban-than-thiet-o-chau-a-4955104.html), the author is also extracted from the bottom of the article.

src/fundus/publishers/vn/vnexpress.py
```diff
@@ -0,0 +1,108 @@
...
+      if meta_keywords and isinstance(meta_keywords, str):
+        return [k.strip() for k in meta_keywords.split(',') if k.strip()]
+      return []
+
+    @attribute
+    def topics(self) -> List[str]:
+      ld_topics = self._parse_ld_keywords()
+      if ld_topics:
+        return ld_topics
+      return self._parse_meta_topics()
```
addie9800: There are also some bloat topic like `Tin nóng` (= hot news), which should be removed.

src/fundus/scraping/filter.py
```diff
@@ -83,8 +83,7 @@ def url_filter(url: str) -> bool:
 
 
 class SupportsBool(Protocol):
-    def __bool__(self) -> bool:
-        ...
+    def __bool__(self) -> bool: ...
```
addie9800: You probably have a different `black` version installed. This PR should normally not edit these files.

MaxDall (approved):
@bachthyaglx Thanks for adding this to fundus; @addie9800 Thanks for finishing!

PR #804, opened 2025-10-22
Add Media Indonesia Publisher

vrdhn91:
To add Media Indonesia publisher, a new python file is created (media_indonesia.py) for the parser class. Extracted information are title, author, published date, and article.


src/fundus/publishers/id/media_indonesia.py
```diff
@@ -0,0 +1,41 @@
...
+    extract_article_body_with_selector,
+    generic_author_parsing,
+    generic_date_parsing,
+    generic_topic_parsing,
+    image_extraction,
+)
+
+
+class MediaIndonesiaParser(ParserProxy):
+    class V1(BaseParser):
```
addie9800, line 23: The `images` attribute seems to be missing

addie9800: `topics` also seem to be missing. They can be accessed using the meta attribute `keywords`.

src/fundus/publishers/id/media_indonesia.py
```diff
@@ -0,0 +1,41 @@
...
+        _paragraph_selector = CSSSelector("div.article")
```
addie9800: This paragraph selector selects just the entire article as one big paragraph. You should consider something like `div.article > p`. Make sure it is consistent with the subheadlines of this [article](https://mediaindonesia.com/ekonomi/823652/pertamina-jadi-sasaran-hoaks-masif-mafindo-minta-masyarakat-selalu-verifikasi). I would recommend switching to XPaths, I personally prefert them for these more subtle cases.

src/fundus/publishers/id/media_indonesia.py
```diff
@@ -0,0 +1,41 @@
...
+        _subheadline_selector = CSSSelector("div.article > h2")
```
addie9800: This [article](https://mediaindonesia.com/ekonomi/823652/pertamina-jadi-sasaran-hoaks-masif-mafindo-minta-masyarakat-selalu-verifikasi) uses a different formatting for the subheadlines.

PR #808, opened 2025-10-23
Add Stuttgarter Zeitung Parser

myoncee:
This PR adds support for Stuttgarter Zeitung (stuttgarter-zeitung.de), a newspaper from Baden-Württemberg, Germany.
Initially attempted to add Süddeutsche Westpresse (SWP), but most articles are behind a paywall.

### DIFF
--- a/src/fundus/publishers/de/__init__.py
+++ b/src/fundus/publishers/de/__init__.py
@@ -39,8 +39,8 @@ from .rheinische_post import RheinischePostParser
 from .rn import RuhrNachrichtenParser
 from .spon import SPONParser
 from .sportschau import SportSchauParser
-from .stuttgarter_zeitung import StuttgarterZeitungParser
 from .stern import SternParser
+from .stuttgarter_zeitung import StuttgarterZeitungParser
 from .sz import SZParser
 from .tagesschau import TagesschauParser
 from .tagesspiegel import TagesspiegelParser
--- a/src/fundus/publishers/de/stuttgarter_zeitung.py
+++ b/src/fundus/publishers/de/stuttgarter_zeitung.py
@@ -1,6 +1,9 @@
 import datetime
 from typing import List, Optional
 
+from lxml.cssselect import CSSSelector
+from lxml.etree import XPath
+
 from fundus.parser import ArticleBody, BaseParser, Image, ParserProxy, attribute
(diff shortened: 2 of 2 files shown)


src/fundus/publishers/de/__init__.py
```diff
@@ -93,6 +94,16 @@ class DE(metaclass=PublisherGroup):
         ],
     )
 
+    StuttgarterZeitung = Publisher(
+        name="Stuttgarter Zeitung",
+        domain="https://www.stuttgarter-zeitung.de/",
+        sources=[
+            NewsMap("https://www.stuttgarter-zeitung.de/docs.newsmap_stuttgarter_zeitung.xml"),
+            Sitemap("https://www.stuttgarter-zeitung.de/docs.universal_sitemap_stuttgarter_zeitung.xml"),
```
MaxDall: It seems like the `NewsMap` and the `Sitemap` contain exactly the same URLs, rendering the `Sitemap` useless, so just having the `NewsMap` should be fine.

src/fundus/publishers/de/stuttgarter_zeitung.py
```diff
@@ -0,0 +1,49 @@
...
+            summary = TextSequence([summary_text]) if summary_text else TextSequence([])
+                    
+            paragraph_elements = self._paragraph_selector(self.precomputed.doc)
+            paragraph_texts = [normalize_whitespace(elem.text_content()) for elem in paragraph_elements]
+
+            subheadline_elements = self._subheadline_selector(self.precomputed.doc)
+
+            sections = [ArticleSection(headline=TextSequence([]), paragraphs=TextSequence(paragraph_texts))]
+
+            return ArticleBody(summary=summary, sections=sections)
```
MaxDall: Thanks for taking the extra effort to include the summary as well! However, for the `body` attribute in Fundus, our goal is to mirror the article’s "body" or text as it appears in the original HTML as closely as possible, only adding minimal structure to the unstructured HTML text. This ensures that the extracted text can be accurately mapped back to the original HTML for annotation purposes.  

Sorry if that wasn’t clear from the documentation, I’ll see if i can make that section more precise.  

Could you please fall back to using the `extract_article_body_with_selector` function?

PR #811, opened 2025-10-23
Add BE Publisher (Politico EU)

rascaria:
Adds a new publisher Politico EU.

### DIFF
--- a/src/fundus/publishers/be/__init__.py
+++ b/src/fundus/publishers/be/__init__.py
@@ -1,6 +1,7 @@
 from fundus.publishers.base_objects import Publisher, PublisherGroup
 from fundus.publishers.be.nieuwsblad import NieuwsbladParser
 from fundus.publishers.be.politico_eu import PoliticoEuParser
+from fundus.scraping.filter import regex_filter
 from fundus.scraping.url import NewsMap, RSSFeed, Sitemap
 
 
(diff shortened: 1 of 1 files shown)


src/fundus/publishers/be/politico_eu.py
```diff
@@ -0,0 +1,44 @@
...
+        def publishing_date(self) -> Optional[datetime.datetime]:
+            return generic_date_parsing(self.precomputed.ld.bf_search("datePublished"))
+
+        @attribute
+        def title(self) -> Optional[str]:
+            return self.precomputed.ld.bf_search("headline")
+
+        @attribute
+        def topics(self) -> List[str]:
+            keywords: Optional[List[str]] = self.precomputed.ld.bf_search("keywords")
```
MaxDall: Judging by the articles i've seen so far, the `keyword` attribute from the articles meta attributes seems to be a better fit here. You can access the articles meta attribute, through `precomputed`.

src/fundus/publishers/be/politico_eu.py
```diff
@@ -0,0 +1,44 @@
...
+from fundus.parser.utility import (
+    extract_article_body_with_selector,
+    generic_author_parsing,
+    generic_date_parsing,
+)
+
+
+class PoliticoEuParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = CSSSelector(".article__content > p")
```
MaxDall: `Politico` seems to use sub-headlines as well as seen in [this article](https://www.politico.eu/article/volodymyr-zelenskyy-ukraine-warned-corruption-eu-readies-enlargement-report-card/?utm_source=RSS_Feed&utm_medium=RSS&utm_campaign=RSS_Syndication)

PR #843, opened 2025-12-10
Update `SeznamZpravy` parser

MaxDall:
I noticed, that there are some `NBSP` characters in our test files while working on this. I would prefer a generalized solution to maybe normalize all strings in the extraction and suggest doing so in a separate PR.


src/fundus/publishers/cz/seznamzpravy.py
```diff
@@ -73,10 +73,16 @@ def images(self) -> List[Image]:
...
-        VALID_UNTIL = datetime.date.today()
+        VALID_UNTIL = datetime.date(2025, 11, 24)
 
         _paragraph_selector = XPath("//div[@class='h_f7 h_bZ h_bZ']/div/p/span[@class='atm-text-decorator' and text()]")
         _subheadline_selector = XPath(
             "//div[@class='h_f7 h_bZ h_bZ']/div/p/span[@class='atm-text-decorator']/span | "
             "//div[@class='h_f7 h_bZ h_bZ']/h2"
         )
+
+    class V1_2(V1_1):
```
addie9800, line 84: Did you mean to change the base branch to `check-coverage-script`?

src/fundus/publishers/cz/seznamzpravy.py
```diff
@@ -73,10 +73,16 @@ def images(self) -> List[Image]:
...
+        VALID_UNTIL = datetime.date.today()
+
+        _paragraph_selector = XPath("//article[@role='article'] //div[contains(@class, 'speakable')] //p")
+        _subheadline_selector = XPath("//article[@role='article'] //h2[contains(@class, 'speakable')]")
```
addie9800, line 88: Is perhaps debatable: One could technically understand the interview questions in [this article](https://www.seznamzpravy.cz/clanek/domaci-zivot-v-cesku-clenity-teren-a-husta-mlha-velitel-popisuje-jak-hledali-pohresovaneho-chlapce-294237) as subheadlines. Question is, if we would like to do so as well.

addie9800: In [this article](https://www.seznamzpravy.cz/clanek/ekonomika-firmy-uhelne-elektrarny-ne-konci-takhle-to-probehne-293826), they are formatted as such

addie9800 (changes requested):
Thanks for fixing this publisher 👍

src/fundus/publishers/cz/seznamzpravy.py
MaxDall: Hmm, I get the argument, but there's hardly a reliable method to parse this. The only marker I could find is the class `e_Y` and the publisher uses it to high lite paragraphs as seen here [[1]](https://www.seznamzpravy.cz/clanek/domaci-zivot-v-cesku-osekane-rychlovlaky-nova-vlada-pocita-jen-s-paterni-siti-294026). As the publisher itself treats this interview questions inconsistently, I would stick to the selectors we have right now.

src/fundus/publishers/cz/seznamzpravy.py
MaxDall: Yes, this branch is actually based on the coverage branch.

addie9800 (dismissed):

PR #865, opened 2026-01-08
Catch parer error in encoding detection

MaxDall:


src/fundus/scraping/html.py
```diff
@@ -51,7 +51,10 @@ def _detect_charset_from_response(response: requests.Response) -> Optional[str]:
     # use response fallback to decode HTML in a first guess
     if not (guessed_text := response.content.decode(response.encoding or "utf-8", errors="replace")):
         return None
-    document = lxml.html.document_fromstring(guessed_text)
+    try:
+        document = lxml.html.document_fromstring(guessed_text)
```
addie9800, line 55: I think it might be beneficial to pass `parser=XMLParser(recover=True)` , as done [here](https://github.com/flairNLP/fundus/blob/bc42e21baca5aade6c47a54b94be81fb9cd9104a/src/fundus/scraping/url.py#L177) into the function instead of just catching the exception. The function in this case would not raise an exception anymore.

MaxDall: @addie9800 Hmm, sounds wird to pass an `XML` parser when what we're parsing is HTML. Am i missing something?

addie9800: My bad, I just copied what we used in the other case and was thrown off since we're catching a XMLSyntaxError. But the idea should, as far as I can see, also hold when passing `parser=HTMLParser(recover=True)`. Do you have the URL for which the issue arose? Then I could try around a bit.

MaxDall: Unfortunately not

MaxDall: After some research, I found that `document_fromstring` defaults to using `HTMLParser` when no parser is specified. Since `HTMLParser` has `recover=True` by default, explicitly providing a parser in this case wouldn't make any difference.

MaxDall: @addie9800 As #875 is already merged, i would suggest closing this one.

addie9800: Sounds good. Unfortunately my tests were also not able to reproduce the crash, so there is nothing left open to do

MaxDall: @addie9800 Thanks for trying anyways!

PR #869, opened 2026-01-13
Update error message

addie9800:
This PR updates the error message to also hint at a possible alternative issue causing the exception to be raised. This is particularly relevant to the process of adding/updating publishers. If a publisher unexpectedly used relative URLs, then the exception will be handled with this misleading error message leaving contributers unsure of why the image_extraction is not working.


MaxDall (approved):
@addie9800 Thanks for catching that! Totally overlooked that.

PR #887, opened 2026-02-07
Fix Nature

addie9800:
Unfortunately, since the publisher coverage did not fail and archive constantly redirecting, I can't accurately place the valid_until date. We know for sure it didn't work since the release of v0.5.5 due to qse errors.


src/fundus/publishers/uk/nature.py
```diff
@@ -85,3 +86,24 @@ def images(self) -> List[Image]:
                 author_selector=self._author_pattern,
                 lower_boundary_selector=self._lower_boundary_selector,
             )
+
+    class V1_1(V1):
+        _paragraph_selector = XPath(
+            "//div[@data-test='main-content' or @class='main-content']//p"
```
MaxDall: In this article [[1]](https://www.nature.com/articles/d41586-026-00662-1) the title of the related article is extracted as paragraph. In general related article titles are extracted.

This article [[2]](https://www.nature.com/articles/s41586-026-10314-z) couldn't be parsed.

MaxDall (changes requested):
@addie9800 Thanks a lot for updating the layout 👍

src/fundus/publishers/uk/nature.py
addie9800: Thanks for catching, should be fixed now 👍

MaxDall: @addie9800 Unfortunately the plaintext of 1 is now cut off.

addie9800: @MaxDall Unfortunately, I cannot reproduce that. The last extracted sentence in my test is: `It’s not crazy. It’s really exciting.”` as expected. Or what do you mean by `cut off`?

MaxDall: @addie9800 Hmm, that's weird. For me the last paragraph is the one before the `Reinventing MRI` subheadline, ending with `“It’s a tiny change on a tiny scale.”`

addie9800: @MaxDall This seems to be an authentication issue. My guess is that when you checked the Nature website you were probably connected in one way or another to the HU license. Does the issue persist, when accessing the articles from a private browser window?

MaxDall: @addie9800 Yes but worse. Im only getting the first Paragraph now, the rest is missing in the HTML which kinda renders the publisher unusable.

addie9800: @MaxDall For some articles that is definitely true, but we also mark them as `free_access=False`. But for other articles, such as: https://www.nature.com/articles/d41586-026-01290-5 , we can get a couple of sections. Granted, it's still not the entire article, but I would estimate around 75%, which is not too bad.

The main issue seems to be that we will not be able to access/parse older articles, because they don't even have this limited preview (the issue we now have with [1]), but are completely locked out. For the recent articles though, I would argue that this publisher is still useful.

MaxDall (approved):
@addie9800 Thanks for fixing this!

PR #888, opened 2026-02-19
Fix author parsing for `NDR`

MaxDall:
Fix author parsing for articles that contain nested lists as authors [[1]](https://www.ndr.de/geschichte/koepfe/lydia-und-josef-russnak-das-musikkonservatorium-oder-ich,russnak-100.html).


addie9800 (approved):
Thanks for fixing this. I used this opportunity to also upate the list of bloat keywords to include "Nachrichten" and "aktuell".

PR #892, opened 2026-03-16
Fix publisher alias lifetime in `EventDict`

MaxDall:
**Problem**
Two `KeyError` crashes caused by publisher aliases being cleaned up while other parts of the code still needed them — once in the main crawl loop, once in the exception handler.

**Root cause**
`EventDict` stored events under the thread ID as the canonical key and deleted everything when `context()` exited. Aliases needed to outlive their thread.

**Fix**
Events are now stored under the alias string as a persistent key. Thread IDs are only tracked while the thread is active. `context()` removes the thread-ID mapping on exit but keeps the alias-keyed events alive. Publisher context registration is also moved out of the scraping generator and into a non-generator thread wrapper so the alias covers the thread's full lifetime.

A `main_context()` method is added for the main crawl thread: it combines `context()` with a full `reset()` on exit and enforces single-instance semantics via a non-reentrant lock, ensuring clean state between consecutive `crawl()` calls.

### DIFF
--- a/src/fundus/utils/events.py
+++ b/src/fundus/utils/events.py
@@ -1,7 +1,7 @@
 import contextlib
 import json
 import threading
-from collections import defaultdict
+
 from typing import Callable, Dict, List, Optional, TypeVar, Union, overload
 
 from bidict import bidict
@@ -40,6 +40,9 @@ class ThreadEventDict(Dict[str, threading.Event]):
         Args:
             default_events: A list of event names for which Event objects
                 should be automatically created when accessed.
+            event_factory: An optional callable used to create new Event objects.
+                Receives the event name and returns a threading.Event. If None,
+                a plain threading.Event() is used.
         """
         super().__init__()
         self._default_events = default_events or []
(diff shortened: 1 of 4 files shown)


src/fundus/utils/events.py
```diff
@@ -224,15 +261,15 @@ def set_for_all(self, event: Optional[str] = None, future: bool = False):
         """
         with self._lock:
             if event is None:
-                for ident, events in self._events.items():
+                for key, events in self._events.items():
                     for name in events:
-                        self.set_event(name, ident)
+                        self.set_event(name, key)
```
addie9800, line 266: I think it would be better to only set events on active threads. With the current setup for example, when running the following code:
```python
for article in crawler.crawl(max_articles_per_publisher=10, only_complete=False, error_handling="suppress"):
   print(article)
```
The stop event is set once, when the publisher has fulfilled it's target and again, when the crawler is shutdown. From my understanding, the second stop event can be ommitted in that case. Prematurely cancelling the the crawler will lead to something like this (Kicker has not yet completed):
```bash
[...]
2026-04-09 20:17:34,516 - fundus.utils.events - DEBUG - Set event 'stop' for (Legal Tribune Online)
2026-04-09 20:17:34,516 - fundus.utils.events - DEBUG - Set event 'stop' for (Frankfurter Allgemeine Zeitung)
2026-04-09 20:17:34,516 - fundus.utils.events - DEBUG - Set event 'stop' for 21556  (Kicker)
2026-04-09 20:17:34,516 - fundus.utils.events - DEBUG - Set event 'stop' for (Norddeutscher Rundfunk (NDR))
2026-04-09 20:17:34,516 - fundus.utils.events - DEBUG - Set event 'stop' for (Sportschau)
2026-04-09 20:17:34,516 - fundus.utils.events - DEBUG - Set event 'stop' for (Hessenschau)
2026-04-09 20:17:34,516 - fundus.utils.events - DEBUG - Set event 'stop' for (WinFuture)
2026-04-09 20:17:34,516 - fundus.utils.events - DEBUG - Set event 'stop' for (Junge Welt)
2026-04-09 20:17:34,516 - fundus.utils.events - DEBUG - Set event 'stop' for (Krautreporter) 
[...]
```

addie9800 (changes requested):
Thanks for fixing this :)

src/fundus/utils/events.py
MaxDall: > Prematurely cancelling the the crawler will lead to something like this (Kicker has not yet completed)

Hmm, but would the crawler be canceled prematurely?

> the second stop event can be ommitted in that case.

Would be hard to differentiate that at the crawler, no?

MaxDall: @addie9800 While working on this, I noticed that `EventDict` still treated the thread ident as a potential event key. This introduces unnecessary complexity, particularly when trying to target only “active” threads, since ident-based entries don’t have a well-defined lifecycle.

Given that we need to address specific threads (not just broadcast), we require each thread to have a stable, canonical identifier nonetheless, therefore the thread ident–as–key logic is unnecessary overhead.

As a result, I removed it in #913 and implemented an `active_only` parameter for `set_for_all`

addie9800 (changes requested):
Looks really good. I just have one remark regarding a doc-string

addie9800 (approved):

PR #894, opened 2026-03-16
Replace `validators.url` with `urllib.parse` for URL validation

MaxDall:
Somehow the `validators` url verification was awfully slow when profiling the library — accounting for ~10% of total parse time across ~35k calls per run.

Replaced all `validators.url(...)` call sites with a small `is_valid_url()` helper backed by `urllib.parse.urlparse`, which is ~15x faster with identical results. The helper lives in `scraping/url.py` and is shared across `scraping/html.py`, `parser/utility.py`, and `parser/data.py`.

**Benchmark:** 27 US publishers × 50 iterations: **16.4s → 15.7s (~5% end-to-end speedup)**

**Also gets rid of the `validators` dependency**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,6 @@ dependencies = [
     "colorama>=0.4, <1",
     "typing-extensions>=4.6, <5",
     "langdetect>=1.0, <2",
-    "validators>=0.24, <1",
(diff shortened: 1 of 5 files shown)


addie9800 (approved):
Wow, that is quite a speed up. Looks good 👍

PR #896, opened 2026-03-17
Replace `urllib` parser with `robotspy`

MaxDall:
While fixing the `TheSun` parser, I discovered that Python's `urllib.robotparser.RobotFileParser`
fails to reliably parse robots.txt rules in certain real-world formats.

Further investigation revealed a well-known issue [[1](https://github.com/python/cpython/issues/138907)] in the official CPython implementation,
where the parser does not handle modern robots.txt conventions correctly — most notably,
blank lines between stacked `User-agent` declarations and their associated rules cause the
entire rule block to be silently dropped.

This PR replaces the official `RobotFileParser` with a `robotspy` drop in replacement.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,8 @@ dependencies = [
     "dict2xml>=1.7.6, <2",
     "xmltodict>=0.13.0, <1",
     "bidict>=0.23, <1",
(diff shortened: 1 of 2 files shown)


addie9800 (approved):
Thanks for fixing this 👍

PR #897, opened 2026-03-31
Remove `request_header` parameter for `WebSource`

MaxDall:
This PR removes the now somewhat forgotten and unnecessary `request_header` parameter from the `WebSource` constructor and uses the `Publisher.request_header` field instead. Additionally, based on these changes, some minor refactoring is applied.


addie9800 (approved):
Looks good to me 👍

PR #899, opened 2026-04-02
Remove unnecessary `join` for manager/pool in `_parallel_crawl`

### LINKED ISSUE #898: [Bug]: Fix UnboundLocalError in _parallel_crawl when Pool init fails
### Describe the bug

When Pool creation fails (or no work results in pool never being created), _parallel_crawl hits finally and calls pool.join() even though pool is unbound, raising UnboundLocalError. This PR initializes pool (and optionally manager) to None and guards cleanup.

### How to reproduce

```python
Fundus 0.5.5 (PyPI) on Python 3.12
Crash: UnboundLocalError: cannot access local variable 'pool' where it is not associated with a value from _parallel_crawl cleanup.
```

### Expected behavior.

If Pool cannot be initialized, the code-path should terminate gracefully. A pool.join() should not be enforced in the finally section.

### Logs and Stack traces

```stacktrace
02/04 08:57:45 AM root - ERROR - Fundus subprocess failed (rc=1):  
from self._parallel_crawl(warc_paths, article_task)
  File "/[...]/.venv/lib/python3.12/site-packages/fundus/scraping/crawler.py", line 728, in _parallel_crawl
    pool.join()
    ^^^^
UnboundLocalError: cannot access local variable 'pool' where it is not associated with a value
```

### Screenshots

_No response_

### Additional Context

Possibilities for fixing:
pool = None before try block in crawler.py _parallel_crawl add into line -> 708

### Environment

```markdown
OS Linux Ubuntu server
python 3.12
.venv
Fundus site-packages (0.5.5)
fundus crawler is called from an async
```

MaxDall:
As shown in #898, placing the pool/manager context inside the try–catch block can lead to issues: if either the manager or the pool raises an exception during construction, we may attempt to join an uninitialized variable.

Unlike `ThreadPool`, there is no real need to manually join or close the process pool or queue manager. Therefore, this PR removes the try–catch block.

Closes #898

### DIFF
--- a/src/fundus/scraping/crawler.py
+++ b/src/fundus/scraping/crawler.py
@@ -711,32 +711,24 @@ class CCNewsCrawler(CrawlerBase):
         # As one could think, because we're downloading a bunch of files, this task is IO-bound, but it is actually
         # process-bound. The reason is that we stream the data and process it on the fly rather than downloading all
         # files and processing them afterward. Therefore, we utilize multiprocessing here instead of multithreading.
-        try:
-            with Manager() as manager, Pool(
-                processes=min(self.processes, len(warc_paths)),
-                initializer=initializer,
-            ) as pool:
-                result_queue: Queue[Union[Article, Exception]] = manager.Queue(maxsize=1000)
+        with Manager() as manager, Pool(
+            processes=min(self.processes, len(warc_paths)),
+            initializer=initializer,
+        ) as pool:
+            result_queue: Queue[Union[Article, Exception]] = manager.Queue(maxsize=1000)
 
-                # Because multiprocessing.Pool does not support iterators as targets,
-                # we wrap the article_task to write the articles to a queue instead of returning them directly.
-                wrapped_article_task: Callable[[str], None] = queue_wrapper(result_queue, article_task)
+            # Because multiprocessing.Pool does not support iterators as targets,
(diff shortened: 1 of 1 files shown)


src/fundus/scraping/crawler.py
```diff
@@ -711,32 +711,24 @@ def _parallel_crawl(
...
 
-                # To avoid restricting the article_task to use only pickleable objects, we serialize it using dill.
-                serialized_article_task = dill_wrapper(spread_article_task)
+            # To avoid restricting the article_task to use only pickleable objects, we serialize it using dill.
+            serialized_article_task = dill_wrapper(spread_article_task)
 
-                # Finally, we build an iterator around the queue, exhausting the queue until the pool is finished.
-                yield from pool_queue_iter(pool.map_async(serialized_article_task, warc_paths), result_queue)
-        finally:
-            logger.debug(f"Shutting down {type(self).__name__!r} ...")
```
addie9800, line 734: Perhaps we could leave one of these logging messages, perhaps in the form of `f"{type(self).__name__!r} shut down"`

addie9800 (changes requested):
That makes sense, thanks for fixing. I only got a minor comment :)

addie9800 (approved):

PR #901, opened 2026-04-09
Close pwn-request vulnerability in documentation workflow

MaxDall:
The previous workflow used `pull_request_target` to trigger doc generation. This trigger runs with base-repo write permissions even for PRs from forks and unlike `pull_request`, it fires immediately without requiring maintainer approval. Because it also checked out and executed code from the fork (`scripts/generate_tables.py`), any contributor could run arbitrary code with a `contents: write` token against the main repository without any human gating the execution.

**Fix:** replace `pull_request_target` with `pull_request` and split the workflow into two jobs:

- `validate` — runs on PRs, no write access; regenerates the docs and fails if the output differs from what is committed, requiring contributors to run the script locally
- `update` — runs on merge to master and manual dispatch only; holds `contents: write` and auto-commits the regenerated docs

Write access is now only ever granted to already-merged, trusted code.

### DIFF
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -17,7 +17,7 @@ env:
   PYTHONPATH: .
 
 jobs:
-  update:
+  supported_publishers:
     runs-on: ubuntu-latest
 
     permissions:
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -39,7 +39,7 @@ jobs:
         run: python scripts/generate_tables.py
 
       - name: Commit changes
-        # [skip ci] prevents this commit from re-triggering the workflow.
+        # [skip ci] prevents this commit from triggering workflows.
         # https://github.com/stefanzweifel/git-auto-commit-action
         uses: stefanzweifel/git-auto-commit-action@v5
         with:
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -11,53 +11,13 @@ on:
       - src/fundus/publishers/**
       - scripts/generate_tables.py
 
-  pull_request:
-    paths:
-      - src/fundus/publishers/**
-      - scripts/generate_tables.py
-
   workflow_dispatch:
 
 env:
   PYTHONPATH: .
 
-# The workflow is split into two jobs to separate privileged and unprivileged work:
-#
-#   validate — runs on pull requests (no write access)
-#     Regenerates the docs and fails if the output differs from what is committed.
-#     Contributors are expected to run scripts/generate_tables.py locally before opening a PR.
-#
-#   update — runs after merge to master or on manual dispatch (write access)
-#     Regenerates the docs and auto-commits any changes. This is the only place
-#     where contents: write is granted, and it only ever runs on trusted, already-merged code.
-#     workflow_dispatch allows force-regenerating the docs from the Actions UI without
-#     a code change — useful if the docs ever get out of sync.
-
 jobs:
-  validate:
-    if: github.event_name == 'pull_request'
-    runs-on: ubuntu-latest
-
-    steps:
-      - name: Set up Git repository
-        uses: actions/checkout@v4
-
-      - name: Set up Python 3.9
-        uses: actions/setup-python@v5
-        with:
-          python-version: 3.9
-
-      - name: Install Fundus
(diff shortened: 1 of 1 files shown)


addie9800 (dismissed):
Thanks for fixing this 👍  Looks good to me.

.github/workflows/documentation.yaml
```diff
@@ -21,34 +20,52 @@ on:
...
-          # back to this branch.
-          ref: ${{ github.head_ref }}
+      - name: Install Fundus
+        run: pip install -e .[dev]
 
+      - name: Generate 'supported_publishers.md'
+        run: python scripts/generate_tables.py
+
+      - name: Fail if generated docs are stale
+        run: git diff --exit-code docs/
```
dobbersc: Instead of failing, wouldn't it be fine to notify (e.g. with a [warning](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-a-warning-message) in the `supported_publishers.md`) that the supported publisher tables need to be updated? With the current setup, the contributor must always run the generate tables themselves, making the update job redundant, as it will never trigger after a pull request merge. So, technically, we wouldn't need the validation or the update step, depending if we want to delegate the table generation to the user vs. the CI. Personally, I would suggest to update the documentation on pushes to the main branch.

dobbersc:
> Thanks for fixing this 👍 Looks good to me.

Sorry, did not realize you already reviewed this. I had not refreshed the page.

addie9800:
> Sorry, did not realize you already reviewed this. I had not refreshed the page.

No worries, I also only just saw that you said, you'd review it :)

.github/workflows/documentation.yaml
MaxDall: The update job is meant to be for manual maintainance, the intend design was to have control over the table changes through the PR, but maybe its better to fully automate the table creation since nobody ever checks the generated table for errors.

MaxDall: I removed the validation job

dobbersc: Yes, I'd imagine that for contributors, having to run the update script manually is just another hurdle in getting a PR merged.

.github/workflows/documentation.yaml
```diff
@@ -11,44 +11,21 @@ on:
...
-  CI_COMMIT_MESSAGE: >-
-    ${{ 
-      github.event_name == 'pull_request' 
-      && format('Update documentation from @ {0} (#{1})', github.event.pull_request.head.sha, github.event.number) 
-      || format('Update documentation from @ {0}', github.sha)
-    }}
 
 jobs:
-  supported_publishers:
+  update:
```
dobbersc: Maybe keep the name as `supported_publishers` again, since we removed the validation job.


```suggestion
  supported_publishers:
```

.github/workflows/documentation.yaml
```diff
@@ -62,7 +39,9 @@ jobs:
         run: python scripts/generate_tables.py
 
       - name: Commit changes
+        # [skip ci] prevents this commit from re-triggering the workflow.
```
dobbersc: Since we have set 

```yaml
push:
    branches:
      - master
    paths:
      # For the current generated documentation, these trigger paths are sufficient.
      # Long-term, job-level triggers are more appropriate.
      # https://how.wtf/run-workflow-step-or-job-based-on-file-changes-github-actions.html
      - src/fundus/publishers/**
      - scripts/generate_tables.py
```

to my understanding changes to the `docs/supported_publishers.md` file should not trigger the action again. Therefore, I'd say that the `[skip ci]` is not needed. Before, with the same logic, we did not require it as well.

MaxDall: Good catch! I would leave it in as this commit isn't supposed to trigger any other workflows. I think the source of confusion here is the comment. I changed it. @dobbersc What do you think?

dobbersc: Ah I see, you don't want to trigger any workflow for the documentation update. Then, wouldn’t to be more principled to add an exclusion for the docs folder for the relevant workflows? Then the skip is not only tied to some commit.

dobbersc: Otherwise this is good, so I'll approve to have the vulnerability fixed fast. If you would like to add workflow level exclusions for the docs folder, it could be done in a separate PR.

dobbersc (approved):

PR #903, opened 2026-04-15
Fix BR

addie9800:
Adds V1_2 for BR


MaxDall (dismissed):
@addie9800 Thank you so much for fixing all those publishers ❤️
Looks good!

PR #904, opened 2026-04-15
Fix TOnline

addie9800:
Adds V1_1 for TOnline


MaxDall (dismissed):
@addie9800 Thanks a lot for fixing this!

Nitpicky, but the headlines extracted from the JSON do not match the one used int the articles. [[1]](https://www.t-online.de/nachrichten/ausland/internationale-politik/id_101223306/iran-krieg-neue-waffenruhe-zeigt-trumps-strategisches-dilemma.html) [[2]](https://www.t-online.de/nachrichten/deutschland/militaer-verteidigung/id_101223526/strasse-von-hormus-wie-die-bundeswehr-bei-der-minenraeumung-helfen-koennte.html)

But im also fine with leaving it as is. Decide as you want.

PR #905, opened 2026-04-15
Fix ABC

addie9800:
This PR adds V1_1 for ABC


src/fundus/publishers/es/abc.py
```diff
@@ -50,12 +61,23 @@ def images(self) -> List[Image]:
...
-                    "./ancestor::div[contains(@class, 'voc-img-container')]//figcaption/span[contains(@class,'author')]"
-                ),
+                caption_selector=self._caption_selector,
+                author_selector=self._author_selector,
             )
+
+    class V1_1(V1):
+        _summary_selector = XPath("//div[@class='v-a-inf-c ']/h2[text()]")
+        _subheadline_selector = XPath("//div[@class='v-c-cmp v-n-mrg']//h3[@class='v-cmp-suh']")
+        _paragraph_selector = XPath("//p[@class='v-d-p' or @class='v-a-t']")
```
MaxDall: The recommended articles get parsed into the main text [[1]](https://www.abc.es/espana/castilla-leon/nuevo-centro-examinador-dgt-leon-operativo-finales-20260422122338-nt.html?ref=https%3A%2F%2Fwww.abc.es%2Fespana%2Fcastilla-leon%2Fnuevo-centro-examinador-dgt-leon-operativo-finales-20260422122338-nt.html) [[2]](https://www.abc.es/espana/comunidad-valenciana/hospital-general-valencia-impulsa-cirugia-corneal-precisa-20260422122056-nt.html)

MaxDall (changes requested):
@addie9800 Thanks so much for fixing this! There are only two minor issues.

The recommended articles bleed into the main text and it may be better to use a test case using sub headlines as well.

src/fundus/publishers/es/abc.py
addie9800: Thanks for catching this, it should be fixed now.

MaxDall (approved):
@addie9800 Thanks for fixing! Looks good.

PR #907, opened 2026-04-16
Fix TimesOfIndia

addie9800:
Adds V2 for TimesOfIndia


MaxDall (dismissed):
@addie9800 Thanks a lot for fixing this! Looks really good.

PR #910, opened 2026-04-20
Fix AnadoluAjansi

addie9800:
This PR adds V2 due to significant layout changes.


MaxDall (dismissed):
@addie9800 Thanks for fixing this! Looks good.

Unrelated to the changes: I would suggest removing all feeds unter the `/teyithatti/` subdomain, as they point to article layouts that differ from the main layout and which we're no longer to parse. Maybe we can also find some more URL sources.

addie9800:
> Maybe we can also find some more URL sources.

Unfortunately, I was unable to find any other URLSources that are on the main website. The Sitemap returns a 404 error and there is only one rss feed link.

MaxDall:
@addie9800 What about removing the `/teyithatti/` subdomian rss feeds?

addie9800:
> What about removing the /teyithatti/ subdomian rss feeds?

My bad, I forgot to select the `__init__.py` when creating the commit. Here it is :)

MaxDall (approved):

PR #911, opened 2026-04-20
Fix DailyNewsTZ

addie9800:
DailyNewsTZ now das two ld json objects containing identical titles, causing an issue as `scalar=True` expects a single entry. This PR just modifies the XPath to choose the first.


MaxDall (dismissed):
@addie9800 Oh wow, that's an easy fix! Thanks a lot!

PR #912, opened 2026-04-20
Fix DailyStar

addie9800:
Adds V1_1


MaxDall (dismissed):
@addie9800 Thanks so much for fixing! Looks really good.

PR #913, opened 2026-04-20
Redesign `EventDict` as a String-Alias-Only Registry

MaxDall:
### Motivation

`EventDict` previously supported a dual-key registry model, where `_events` could be indexed by either string aliases or integer thread IDs. This behavior was inherited from an earlier design, but in practice it introduced unnecessary complexity and led to inconsistent state management.

Several issues stemmed from this approach:

- `active_only` in `set_for_all` could not be implemented reliably, since integer-keyed entries had no associated lifecycle state, making the concept of “active” undefined.
- If a thread registered events before entering `context()`, it could end up with two separate `ThreadEventDict` instances—one keyed by thread ID and another by alias—leaving the original entry orphaned.
- Iteration over `_events` required repeated `isinstance` checks to distinguish between aliases and thread IDs.
- `_global_events` and `remove_alias()` were effectively unused, adding dead code to the implementation.

The core design issue was that targeted communication requires a stable identifier. String aliases already fulfill that role, while anonymous integer-keyed entries do not provide meaningful addressability. If a thread needs to be addressed, it should be registered under an explicit alias.

This change simplifies the registry model by making aliases the sole canonical key for `EventDict`, reducing ambiguity and eliminating the edge cases introduced by the legacy dual-key design.

### DIFF
--- a/src/fundus/utils/events.py
+++ b/src/fundus/utils/events.py
@@ -1,7 +1,7 @@
 import contextlib
 import json
 import threading
-from collections import defaultdict
+
 from typing import Callable, Dict, List, Optional, TypeVar, Union, overload
 
 from bidict import bidict
@@ -40,6 +40,9 @@ class ThreadEventDict(Dict[str, threading.Event]):
         Args:
             default_events: A list of event names for which Event objects
                 should be automatically created when accessed.
+            event_factory: An optional callable used to create new Event objects.
+                Receives the event name and returns a threading.Event. If None,
+                a plain threading.Event() is used.
         """
         super().__init__()
         self._default_events = default_events or []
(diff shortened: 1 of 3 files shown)


src/fundus/utils/events.py
```diff
@@ -107,180 +117,161 @@ def event_factory(name: str) -> threading.Event:
...
         Should only be called while holding the internal lock.
 
         Args:
-            key: Thread ID, alias, or None.
+            key: Alias string or ``None`` (current thread).
 
         Returns:
-            str: A formatted string such as ``"7740   (Sportschau)"`` or ``"(Sportschau)"``.
+            A formatted string such as ``"7740   (Sportschau)"`` or
+            ``"(Sportschau)"`` when the thread has already finished.
```
addie9800, line 162: For consistency, it might make sense to also mention the RuntimeError that is raised through _resolve

addie9800 (changes requested):
Looks good, I just have a minor remark regarding a doc-string

addie9800 (approved):

PR #914, opened 2026-04-22
Remove dispatch for deployment workflow and add version check

MaxDall:
This PR:
- fixes a bug regarding the workflow path for the linting workflow
- removes the workflow dispatch and permission check (new deployment ENVs a restricted on matching tags only now)
- add a package version check against the tag


addie9800 (approved):
Looks goodd 👍

PR #915, opened 2026-04-23
Add anonymous per thread entry for un aliased threads

MaxDall:
The previous changes made it impossible to use any object that queries/sets events (like `WebSource` etc.) outside of an alias context. This made testing and especially development very tedious in certain parts/scenarios. This PR adds an auto generated anonymous alias to threads that query/set events.

### DIFF
--- a/src/fundus/utils/events.py
+++ b/src/fundus/utils/events.py
@@ -130,6 +130,11 @@ class EventDict:
     def _resolve(self, key: Optional[str]) -> str:
         """Resolve ``None`` to the calling thread's alias, or return ``key`` unchanged.
 
+        When the calling thread has no registered alias, an anonymous per-thread
+        entry is created automatically so that operations like :meth:`is_event_set`
+        and :meth:`get` work safely outside of any :meth:`context` block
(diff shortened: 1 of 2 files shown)


addie9800 (commented):
Is there a way to clean-up the registered threads? I mean it probably isn't that much of an issue during normal usage, so if it's too much effort to implement, I wouldn't insist on it, but it might be nicer to remove the entry after completion.

MaxDall:
> Is there a way to clean-up the registered threads? I mean it probably isn't that much of an issue during normal usage, so if it's too much effort to implement, I wouldn't insist on it, but it might be nicer to remove the entry after completion.

The only thing i can think about managing resources like this is a context manager, but than we would be back at the same problem requiring a context manager to use basic objects.

There sure is a more sophisticated solution, im just not sure if's worth the afford as this is used mainly for debugging

addie9800:
> this is used mainly for debugging

That is a good point. I guess it's not as important, looking from that perspective.

addie9800 (approved):

PR #916, opened 2026-04-25
Fix TheBBC

addie9800:
Introduces V2


MaxDall (approved):
@addie9800 Looks amazing! Thank you so much ❤️

PR #920, opened 2026-05-13
Replace `requests` with `curl-cffi` and add browser impersonation support

MaxDall:
### Summary

- **Swap HTTP backend**: `requests` is replaced by `curl-cffi` across the entire scraping stack (`session.py`, `html.py`, `url.py`, `base_objects.py`). `curl-cffi` wraps libcurl with a TLS/JA3 fingerprint that matches real browsers, making it significantly harder for sites to detect and block Fundus.

- **Browser impersonation per publisher**: `Publisher` gains an optional `impersonate` parameter (e.g. `"chrome"`, `"safari"`). When set, every HTTP request (including robots.txt fetches, RSS/sitemap crawls, and article fetches) is made with the corresponding browser fingerprint. The value is validated against `curl-cffi`'s supported targets at construction time.

- **Reworked `SessionHandler`**: Replaced the old single shared session with a per-thread registry (`Dict[int, InterruptableSession]`). Each session is backed by a persistent worker thread, keeping the curl handle alive across requests for connection reuse. `context()` swaps the registry atomically, giving all threads a clean slate and restoring their original sessions on exit.

- **`URLSource.fetch()` API**: The abstract `__iter__` on `URLSource` is replaced by `fetch(session, headers)`, letting `WebSource` pass the publisher's session (with the right impersonate profile) down into RSS and sitemap sources. `__iter__` is kept as a convenience wrapper for standalone/testing use.

- **Encoding detection moved into the session layer**: The lxml-based charset detection in `html.py` is removed. `curl-cffi` receives a `default_encoding` callable (`_detect_encoding_from_bytes`) that checks the HTML `<meta charset>` tag first, then falls back to chardet, then to UTF-8, applied automatically before the response text is read.

- **Un-deprecate TLS-protected publishers**: Publishers that were previously deprecated due to TLS fingerprint blocking are re-enabled, now that impersonation makes them reachable again.

### Test plan

- [x] `pytest tests/test_session_handler.py` — 39 unit tests covering `InterruptableSession`, `SessionHandler`, redirect following, encoding detection, and context behaviour
- [x] Smoke-crawl a plain publisher (no impersonation) and verify articles are returned with correct encoding
- [x] Smoke-crawl an impersonation-enabled publisher (e.g. one of the re-enabled TLS-protected ones) and verify articles are returned
- [x] Verify `Publisher(impersonate="invalid")` raises `ValueError` at construction time

### DIFF
--- a/src/fundus/publishers/de/__init__.py
+++ b/src/fundus/publishers/de/__init__.py
@@ -514,6 +514,7 @@ class DE(metaclass=PublisherGroup):
             NewsMap("https://newsfeed.kicker.de/googlesitemapnews.xml"),
         ],
         url_filter=regex_filter("/slideshow|/video|heute-live|live-konferenz|/bilder|/ticker"),
+        impersonate="chrome",
     )
 
     Krautreporter = Publisher(
@@ -594,7 +595,6 @@ class DE(metaclass=PublisherGroup):
             RSSFeed("https://www.freiepresse.de/rss/rss_regional.php"),
             Sitemap("https://www.freiepresse.de/sitemaps/articles_last2years.xml", reverse=True),
         ],
-        impersonate="chrome",
     )
 
     RuhrNachrichten = Publisher(
--- a/src/fundus/publishers/de/__init__.py
+++ b/src/fundus/publishers/de/__init__.py
@@ -594,7 +594,7 @@ class DE(metaclass=PublisherGroup):
             RSSFeed("https://www.freiepresse.de/rss/rss_regional.php"),
             Sitemap("https://www.freiepresse.de/sitemaps/articles_last2years.xml", reverse=True),
         ],
-        deprecated=True,
+        impersonate="chrome",
     )
 
     RuhrNachrichten = Publisher(
--- a/src/fundus/publishers/fr/__init__.py
+++ b/src/fundus/publishers/fr/__init__.py
@@ -51,5 +51,4 @@ class FR(metaclass=PublisherGroup):
             NewsMap("https://www.lesechos.fr/sitemap_news.xml"),
         ],
         impersonate="chrome",
-        request_header={"User-agent": "Fundus"},
     )
--- a/src/fundus/publishers/fr/__init__.py
+++ b/src/fundus/publishers/fr/__init__.py
@@ -50,5 +50,6 @@ class FR(metaclass=PublisherGroup):
             Sitemap("https://sitemap.lesechos.fr/sitemap_index.xml", reverse=True),
             NewsMap("https://www.lesechos.fr/sitemap_news.xml"),
         ],
-        deprecated=True,
+        impersonate="chrome",
+        request_header={"User-agent": "Fundus"},
     )
--- a/src/fundus/publishers/jp/__init__.py
+++ b/src/fundus/publishers/jp/__init__.py
@@ -85,6 +85,7 @@ class JP(metaclass=PublisherGroup):
                 sitemap_filter=inverse(regex_filter(r"[a-z]*\.sitemap\.xml$")),
             )
         ],
+        deprecated=True,
         impersonate="chrome",
     )
 
@@ -109,5 +110,4 @@ class JP(metaclass=PublisherGroup):
                 sitemap_filter=inverse(regex_filter(r"type=articles")),
             )
         ],
-        impersonate="chrome",
     )
--- a/src/fundus/publishers/jp/__init__.py
+++ b/src/fundus/publishers/jp/__init__.py
@@ -85,7 +85,7 @@ class JP(metaclass=PublisherGroup):
                 sitemap_filter=inverse(regex_filter(r"[a-z]*\.sitemap\.xml$")),
             )
         ],
-        deprecated=True,
+        impersonate="chrome",
     )
 
     SankeiShimbun = Publisher(
@@ -102,12 +102,12 @@ class JP(metaclass=PublisherGroup):
         name="Nikkan Geadai",
         domain="https://www.nikkan-gendai.com/",
         parser=NikkanGeadaiParser,
-        deprecated=True,
         sources=[
             Sitemap(
                 "https://www.nikkan-gendai.com/sitemap.xml",
                 reverse=True,
-                sitemap_filter=inverse(regex_filter(r"type=news")),
+                sitemap_filter=inverse(regex_filter(r"type=articles")),
             )
         ],
+        impersonate="chrome",
     )
--- a/src/fundus/publishers/na/__init__.py
+++ b/src/fundus/publishers/na/__init__.py
@@ -21,5 +21,5 @@ class NA(metaclass=PublisherGroup):
                 languages={"en", "kj"},
             ),
         ],
-        deprecated=True,
+        impersonate="chrome",
     )
--- a/src/fundus/publishers/na/__init__.py
+++ b/src/fundus/publishers/na/__init__.py
@@ -21,5 +21,4 @@ class NA(metaclass=PublisherGroup):
                 languages={"en", "kj"},
             ),
         ],
-        impersonate="chrome",
     )
--- a/src/fundus/publishers/uk/__init__.py
+++ b/src/fundus/publishers/uk/__init__.py
@@ -63,7 +63,7 @@ class UK(metaclass=PublisherGroup):
             Sitemap("https://www.telegraph.co.uk/sitemap.xml"),
             NewsMap("https://www.telegraph.co.uk/custom/daily-news/sitemap.xml"),
         ],
-        deprecated=True,
+        impersonate="chrome",
     )
 
     iNews = Publisher(
--- a/src/fundus/publishers/uk/__init__.py
+++ b/src/fundus/publishers/uk/__init__.py
@@ -53,6 +53,7 @@ class UK(metaclass=PublisherGroup):
             Sitemap("https://www.mirror.co.uk/sitemaps/sitemap_index.xml", reverse=True),
             NewsMap("https://www.mirror.co.uk/map_news.xml"),
         ],
+        impersonate="chrome",
     )
 
     TheTelegraph = Publisher(
@@ -63,7 +64,7 @@ class UK(metaclass=PublisherGroup):
             Sitemap("https://www.telegraph.co.uk/sitemap.xml"),
             NewsMap("https://www.telegraph.co.uk/custom/daily-news/sitemap.xml"),
         ],
-        impersonate="chrome",
+        impersonate="chrome99_android",
     )
 
     iNews = Publisher(
@@ -121,6 +122,7 @@ class UK(metaclass=PublisherGroup):
             Sitemap(f"https://www.dailymail.co.uk/sitemap-articles-year~{year.year}.xml")
             for year in rrule(YEARLY, dtstart=datetime(2021, 1, 1), until=datetime.today())
         ],
+        impersonate="chrome",
     )
 
     EveningStandard = Publisher(
--- a/src/fundus/scraping/crawler.py
+++ b/src/fundus/scraping/crawler.py
@@ -600,9 +600,7 @@ class Crawler(CrawlerBase):
             queue_wrapper(result_queue, article_task, silenced_exceptions=(CrashThread,))
         )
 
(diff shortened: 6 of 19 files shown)


src/fundus/scraping/session.py
```diff
@@ -22,43 +28,119 @@ class CrashThread(BaseException):
...
+
+    Returns:
+        Detected encoding string, guaranteed non-empty.
+    """
+    if encoding := _detect_encoding_from_html(response):
+        logger.debug(f"Detected encoding from HTML: {encoding!r}")
+    elif encoding := chardet.detect(response)["encoding"]:
+        logger.debug(f"Detected encoding from chardet: {encoding!r}")
+    else:
+        logger.debug("Unable to detect encoding from response")
```
addie9800:
```suggestion
        logger.debug("Unable to detect encoding from response. Defaulting to <utf-8>")
```

src/fundus/scraping/session.py
```diff
@@ -22,43 +28,119 @@ class CrashThread(BaseException):
...
+        history: List[curl_cffi.requests.Response] = []
+        current = url
+        request_kwargs = {} if self.impersonate else kwargs
+
+        for _ in range(self.max_redirects):
+            response = self.get(current, **request_kwargs, allow_redirects=False)
+
+            if not (300 <= response.status_code <= 399):
+                object.__setattr__(response, "_history", history)
+                return response  # type: ignore[no-any-return]
```
addie9800: I think typing should be fine without the ignore statement. 
```suggestion
                return response
```

src/fundus/scraping/session.py
```diff
@@ -70,156 +152,97 @@ def _req():
...
-        socket.setdefaulttimeout(old_timeout)
+        prev_kwargs = self._session_kwargs
+        prev_sessions = self._sessions
+        self._session_kwargs = {**self.DEFAULT_SESSION_KWARGS, **kwargs}
+        self._sessions = {}
+
+        try:
+            yield self
+        finally:
+            self._context_lock.release()
```
addie9800: Without too much background in threading, I would assume that the block release should occur last, after the original parameters have been reset and the sessions have been successfully closed.

src/fundus/scraping/url.py
```diff
@@ -112,21 +112,30 @@ class URLSource(Iterable[str], ABC):
...
     @abstractmethod
-    def __iter__(self) -> Iterator[str]:
+    def fetch(self, session: InterruptableSession, headers: Dict[str, str]) -> Iterator[str]:
+        """Fetch URLs using the provided session and headers.
+
+        Args:
+            session: The HTTP session to use for requests.
+            headers: Request headers to include. Note that when the session was created
+                with an impersonate profile, headers may be dropped in favour of the
+                browser fingerprint (see InterruptableSession.get_with_interrupt).
```
addie9800, line 127: You refer to `InterruptableSession.get_with_interrupt()` here, but from what I can see, the described behavior is not really apparent in the code, nor in the relevant documentation. If this comment is present, I would expect either additional details in the docstring or an explicit code snippet in the function implementing this behavior.

tests/test_session_handler.py
```diff
@@ -75,3 +100,416 @@ def set_context_fail():
...
+            nonlocal pre_context_session, inside_context_session
+            pre_context_session = session_handler.get_session()
+            ready.set()
+            entered.wait()
+            inside_context_session = session_handler.get_session()
+            sampled.set()
+
+        t = Thread(target=worker)
+        t.start()
+        ready.wait()
```
addie9800: It might make sense to add a timeout here (and handle the timeout case), in case something goes wrong when spawning the thread, causing the tests to hang.

tests/test_session_handler.py
```diff
@@ -75,3 +100,416 @@ def set_context_fail():
...
+
+        with session_handler.context(timeout=9999):
+            entered.set()
+            sampled.wait()
```
addie9800: Similarly here.

tests/test_session_handler.py
```diff
@@ -75,3 +100,416 @@ def set_context_fail():
...
+        def worker():
+            nonlocal pre_context_session, post_context_session
+            pre_context_session = session_handler.get_session()
+            entered.set()
+            exited.wait()
+            post_context_session = session_handler.get_session()
+
+        t = Thread(target=worker)
+        t.start()
+        entered.wait()
```
addie9800: Similarly here

tests/test_session_handler.py
```diff
@@ -75,3 +100,416 @@ def set_context_fail():
...
+        assert not session._worker_thread.is_alive()
+
+    def test_close_is_nonblocking(self):
+        """close() returns immediately even when a request is in flight."""
+        session = InterruptableSession()
+        inside_request = threading.Event()
+
+        def slow_follow_redirects(url, **kwargs):
+            inside_request.set()
+            time.sleep(30)
```
addie9800: I think it might be a bit cleaner to use another event such as `allow_return` for example and replace `time.sleep(30)` with `allow_return.wait(timeout=15)`, allowing a clean finish as soon as the session is closed, when setting `allow_return.set()` after `session.close()`.

tests/test_session_handler.py
```diff
@@ -75,3 +100,416 @@ def set_context_fail():
...
+        session.close()
+
+    def test_crash_thread_on_stop_event(self):
+        """CrashThread is raised in the caller when the stop event fires while polling."""
+        session = InterruptableSession()
+        inside_request = threading.Event()
+
+        def slow_follow_redirects(url, **kwargs):
+            inside_request.set()
+            time.sleep(30)
```
addie9800: Similar problem as in `test_close_is_nonblocking`

tests/test_session_handler.py
```diff
@@ -75,3 +100,416 @@ def set_context_fail():
...
+        assert first is not second
+        assert second.impersonate == "chrome"
+
+    def test_context_restores_session_on_exception(self):
+        session_handler = SessionHandler()
+        prev = session_handler.get_session()
+
+        with pytest.raises(RuntimeError):
+            with session_handler.context(timeout=5):
+                raise RuntimeError("boom")
```
addie9800, line 479: 💥

src/fundus/publishers/de/__init__.py
addie9800, line 579: `Kicker`, `TheMirror` and `DailyMail` also need `impersonate="chrome"` to function (otherwise access to robots.txt is forbidden resulting in disallow all fallback)

addie9800 (changes requested):
Thanks a lot for adding this, it already looks really good. There is still one thing I would like to at least mention and perhaps open a discussion about. As we are now introducing browser impersonation, which, depending on the use case, might stand in an ethical gray area, I would discuss the necessity of the following extra measures:
- introduce documentation (for contributors and also for users)
- allow users to reject impersonation

src/fundus/scraping/session.py
MaxDall: At least for me it isn't :( Maybe you're on a different mypy version than the repo? `mypy==1.9.0`

src/fundus/scraping/url.py
MaxDall: It is, but apparently at the wrong place. It is located in `_follow_redirects`. I'll relocate it and add some docstrings. Good catch!

tests/test_session_handler.py
MaxDall: Good catch!

src/fundus/publishers/jp/__init__.py
```diff
@@ -85,7 +85,7 @@ class JP(metaclass=PublisherGroup):
                 sitemap_filter=inverse(regex_filter(r"[a-z]*\.sitemap\.xml$")),
             )
         ],
-        deprecated=True,
+        impersonate="chrome",
```
addie9800, line 89: The sitemap does not contain any articles anymore, so I would keep this deprecated.

src/fundus/scraping/crawler.py
```diff
@@ -476,6 +476,7 @@ def __init__(
         threading: bool = True,
         ignore_robots: bool = False,
         ignore_crawl_delay: bool = False,
+        impersonate: bool = False,
```
addie9800, line 479: When running a Crawler with this setup:
```python
publisher = PublisherCollection.jp.NikkanGeadai
crawler = Crawler(publisher, impersonate=False)
```
I can still crawl `NikkanGeadai` without issues, compared to on `master` where this request would be blocked. This also doesn't apply to all publishers, e.g. for `LesEchos` it works as expected. To be honest, I don't understand `curl_cffi` well enough to know if this is an issue or not, but I thought I'd mention it.

src/fundus/publishers/uk/__init__.py
```diff
@@ -63,7 +63,7 @@ class UK(metaclass=PublisherGroup):
             Sitemap("https://www.telegraph.co.uk/sitemap.xml"),
             NewsMap("https://www.telegraph.co.uk/custom/daily-news/sitemap.xml"),
         ],
-        deprecated=True,
+        impersonate="chrome",
```
addie9800: This publisher also doesn't work with impersonation.

src/fundus/publishers/de/__init__.py
```diff
@@ -594,7 +594,7 @@ class DE(metaclass=PublisherGroup):
             RSSFeed("https://www.freiepresse.de/rss/rss_regional.php"),
             Sitemap("https://www.freiepresse.de/sitemaps/articles_last2years.xml", reverse=True),
         ],
-        deprecated=True,
```
addie9800, line 597: Just with the implementation of curl-cffi as it is now, it doesn't even need the impersonation attribute (refer to my other comment in `JP`)

src/fundus/publishers/fr/__init__.py
```diff
@@ -50,5 +50,6 @@ class FR(metaclass=PublisherGroup):
             Sitemap("https://sitemap.lesechos.fr/sitemap_index.xml", reverse=True),
             NewsMap("https://www.lesechos.fr/sitemap_news.xml"),
         ],
-        deprecated=True,
+        impersonate="chrome",
+        request_header={"User-agent": "Fundus"},
```
addie9800: Is this custom request_header necessary? It works for me even without it.

src/fundus/publishers/na/__init__.py
```diff
@@ -21,5 +21,5 @@ class NA(metaclass=PublisherGroup):
                 languages={"en", "kj"},
             ),
         ],
-        deprecated=True,
+        impersonate="chrome",
```
addie9800: Impersonation is not necessary with curl-cffi. (Refer to previous comments)

addie9800 (changes requested):

src/fundus/publishers/jp/__init__.py
MaxDall: Good catch! unfortunately yes.

src/fundus/scraping/crawler.py
MaxDall: Even without directly impersonating the TSL handshake of a given browser profile, the `curl-cffi` package changes fundus TSL fingerprint, as it uses `curl` as a backend. So any bot protection tuned to filter out, let's say, python`s `request` module, will fail to detect a request through `curl-cffi` as a bot.

Further I think I prematurely tested `curl-cffi`s browser profiles against the `requests` module and not the `culr-cffi` base profile to see if a browser profile would be actually necessary. Thanks a lot for catching this! I will remove the unnecessary  profiles.

src/fundus/publishers/uk/__init__.py
MaxDall: It seems to work with `chrome99_android`

src/fundus/publishers/fr/__init__.py
MaxDall: No, looks like a leftover to me. Thanks for catching. I'm gonna remove that.

src/fundus/publishers/na/__init__.py
MaxDall: I'll remove it. Thanks for catching!

addie9800 (approved):
Looks great, thank you so much 🚀

PR #922, opened 2026-05-18
Fix BerlinerZeitung

addie9800:


MaxDall (changes requested):
@addie9800 Thanks a lot for updating this one. In line with the [docs](https://github.com/flairNLP/fundus/blob/master/docs/how_to_add_a_publisher.md#8-maintaining-publishers) and in terms of readability i would see the update as a new major version.

MaxDall (approved):

PR #923, opened 2026-05-18
Fix JungeWelt

addie9800:


MaxDall (approved):
@addie9800 Thanks so much for fixing this!

PR #924, opened 2026-05-18
Fix TagesAnzeiger

addie9800:
Also renames files and parser to comply with fundus default.


MaxDall (approved):
@addie9800 Thanks a lot for fixing this!

PR #927, opened 2026-05-24
Add afp faktencheck

addie9800:


MaxDall (approved):
@addie9800 Thanks so much for adding! Looks really good 🚀

PR #928, opened 2026-05-24
Add `GlobalNews`

addie9800:


src/fundus/publishers/ca/global_news.py
```diff
@@ -0,0 +1,55 @@
...
+    generic_author_parsing,
+    generic_date_parsing,
+    generic_topic_parsing,
+    image_extraction,
+)
+
+
+class GlobalNewsParser(ParserProxy):
+    class V1(BaseParser):
+        _paragraph_selector = XPath("//article/p[not(text()='—') and text()]")
```
MaxDall: Articles from under the `/the-curator/` subdomain aren't properly extracted. We should either filter out the subdomain or adjust the selectors. [[1]](https://globalnews.ca/the-curator/11873266/best-summer-toys/)

MaxDall: There is a "meta" line selected at the end of this article [[1]](https://globalnews.ca/news/11895605/quebec-water-reserves-at-risk/). Not sure though if it is worth filtering these out, as they are structural indistinguishable from the actual paragraphs, and would require a lot of regex selection.

MaxDall: The table here isn't extracted as well [[1]](https://globalnews.ca/news/11895241/metro-vancouver-stage-3-water-restriction/). Maybe it's worth now aiming for a general solution regarding tables, as this is a recurring topic.

MaxDall (changes requested):
@addie9800 Thanks so much for adding this. Looks really good! Just some minor adjustments.

src/fundus/publishers/ca/global_news.py
addie9800: Honestly, as that subdomain seems to be primarily advertisements, I would just filter it out.

addie9800: Well, we can try. Maybe this case covers it :)

addie9800: Do you mean the list? I updated it for now, but yes, a more general solution could be quite beneficial.

MaxDall (approved):
@addie9800 Thanks a lot! Looks really good 🚀

PR #929, opened 2026-05-25
Add `OttawaCitizen`

addie9800:


src/fundus/publishers/ca/ottawa_citizen.py
```diff
@@ -0,0 +1,15 @@
+from lxml.etree import XPath
+
+from fundus.parser import ParserProxy
+from fundus.publishers.ca.national_post import NationalPostParser
+
+
+class OttawaCitizenParser(ParserProxy):
+    class V1(NationalPostParser.V1_1):
+        _paragraph_selector = XPath("//div[@class='story-v2-content-element-inline']/p[span[text()] or text()]")
```
MaxDall: **Lists dropped from body.** This `<p>`-only selector skips `<ul>`/`<ol>`, so list content disappears. On this article the eligibility list renders as a dangling "…fell into either of two groups:" followed straight by the next paragraph — the `<ul>` is gone. [[1]](https://ottawacitizen.com/public-service/4000-public-servants-approved-early-retirement)

Suggest appending `| //div[@class='story-v2-content-element-inline']/ul/li | //div[@class='story-v2-content-element-inline']/ol/li` (likely worth fixing in the `NationalPostParser` base too).

src/fundus/publishers/ca/ottawa_citizen.py
```diff
@@ -0,0 +1,15 @@
...
+        _subheadline_selector = XPath(
+            "//div[@class='story-v2-content-element-inline']/h3 |"
+            "//div[@class='story-v2-content-element-inline']/p/*[self::strong or self::b]"
```
MaxDall: **Inline bold treated as a subheadline.** `p/*[self::strong or self::b]` matches *any* bold inside a paragraph, not just standalone subheadlines, so inline emphasis gets fragmented/duplicated into the body. E.g. "…reporters **Natasha Baldin** and **Don Brennan** will answer…" splits off stray "Natasha Baldin and" / "Don Brennan will answer…" lines. [[1]](https://ottawacitizen.com/sports/pwhl-expansion-ottawa-charge-live-q-a)

Restrict to bold that is the sole child of its `<p>`, e.g. `//div[@class='story-v2-content-element-inline']/p[not(text()[normalize-space()]) and count(*)=1]/*[self::strong or self::b]`.

src/fundus/publishers/ca/national_post.py
```diff
@@ -54,15 +68,11 @@ def title(self) -> Optional[str]:
 
         @attribute
         def topics(self) -> List[str]:
-            preliminary_topics = self.precomputed.ld.bf_search("keywords")
-            filter_list = ["Curated", "News", "Newsroom daily", "story", "Canada", "World", "nationalpost.com"]
-            topic_filter = regex_filter(
-                r"([0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}|NLP Entity Tokens|NLP Category|NP Comment|Category:)"
+            return generic_topic_parsing(
```
MaxDall, line 71: **Behavior change, not a pure refactor:** `regex_filter` dropped the whole topic on match; `substitution_pattern` only strips the matched substring. So `"Category: X"` now yields `"X"` instead of being dropped. UUID / `NLP …` tokens still reduce to `""` and drop out, so this is likely an improvement — just flagging it.

src/fundus/parser/utility.py
```diff
@@ -414,7 +414,12 @@ def apply_result_filter(input_list: List[str], result_filter: Optional[Union[Pat
     if isinstance(result_filter, Pattern):
         return [topic for topic in dict.fromkeys(input_list) if not re.search(result_filter, topic)]
     else:
-        return [topic for topic in dict.fromkeys(input_list) if result_filter is None or topic not in result_filter]
+        normalized_result_filter = {val.lower() for val in result_filter} if result_filter else set()
```
MaxDall, line 417: **Cross-cutting:** this makes `Set` `result_filter` case-insensitive for *every* publisher using one (`za/times_live`, `vn/vnexpress`, `uk/nature`, `ap_news`, …), not just this PR. It's needed so the lowercase bloat entries match capitalized topics, and the full parser suite stays green — no regressions. 👍

MaxDall (changes requested):
Thanks for the addition! I crawled a handful of live articles and read the bodies. The parser is close, but two issues drop/garble body content and should be fixed before merge (both inherited from the `NationalPostParser` selectors, so the fix probably belongs in the base):

- **Lists dropped** — `<p>`-only paragraph selector skips `<ul>`/`<ol>`, leaving dangling "…:" sentences.
- **Inline bold treated as subheadline** — fragments/duplicates paragraphs that bold names or words inline.

Two non-blocking notes inline as well (topics behavior change + the cross-cutting `apply_result_filter` change, which is fine — full parser suite stays green). Details on the lines below.

MaxDall:
@addie9800 I experimented with automating the review process with Claude Code 🚀 Please let me know what you think.

addie9800:
@MaxDall Conceptually, this looks really good and Claude did a good job of reviewing the PR. There are a couple of things I noticed though:
- There is some bloat, that passed through undetected `Read the questions and answers in the comments section below.`, it could make sense to try to make Claude aware of this issue
- Also, Claude commented on some additional changes I made, but did not review them. I am not sure though if that is because there were no issues or if it was out of scope. (But maybe in that case it would also make sense to maybe not include this in this PR type)
- An additional point to include could be trying to keep the selectors minimal. Claudes suggestion was pretty much a chain of different selectors for each case.

src/fundus/publishers/ca/ottawa_citizen.py
```diff
@@ -0,0 +1,19 @@
...
+
+from fundus.parser import ParserProxy
+from fundus.publishers.ca.national_post import NationalPostParser
+
+
+class OttawaCitizenParser(ParserProxy):
+    class V1(NationalPostParser.V1_1):
+        _paragraph_selector = XPath(
+            "//div[@class='story-v2-content-element-inline']/p[span[text()] or text()] | "
+            "//div[@class='story-v2-content-element-inline']/ul/li"
```
MaxDall: **Blocker — ordered lists dropped.** Body dangles on "…Here's the process:" with the `<ol>` gone: "Choose Your Sportsbook… Create Your Account… Verify… Deposit… Place Your Bet". Selector matches `ul/li` but not `ol`. Fix: `…/*[self::ul or self::ol]/li` (direct-child, so the nested `more-topic` `<ol>` chrome stays out). [[1]](https://ottawacitizen.com/sponsored/life-sponsored/best-soccer-betting-sites-in-canada)

src/fundus/publishers/ca/ottawa_citizen.py
```diff
@@ -0,0 +1,19 @@
...
+        )
+        _subheadline_selector = XPath(
+            "//div[@class='story-v2-content-element-inline']/h3[not(re:test(string(), '(?i)^read the questions and answers'))] |"
+            "//div[@class='story-v2-content-element-inline']/p/*[(self::strong or self::b) and not(text())]",
```
MaxDall, line 15: **Nit — dead branch.** Real `<p><strong>` subheads have direct text, so `not(text())` excludes them; inline `<strong>` is already covered by the paragraph's `text()`. Matches almost nothing — drop it to keep the selector minimal.

MaxDall (changes requested):
**Re-review `OttawaCitizen`** — read 4 live articles / 2 layouts. One blocker; the inline-bold fix works.

- **Blocker — `<ol>` still dropped** (the `<ul>`-only half of the earlier fix) → inline.

Non-blocking:
- `_subheadline_selector`'s `not(text())` branch is effectively dead → inline.
- News articles are `isAccessibleForFree:false`, but `free_access` is unimplemented — same as parent `NationalPost` and all of `ca/`, so flagging not blocking.
- Minor topic noise: `Realtime`, `Life-Sponsored`, `BestOntarioCasinoSites.ca`.

src/fundus/publishers/ca/ottawa_citizen.py
addie9800: I am not too sure about this. I mean it is only also a nitpick, but I think this should be what addressed Claudes previous comment. (@MaxDall does Claude take replies to it's previous comments into account?)

MaxDall: @addie9800 Yes, the skill uses the entire conversation for context. To settle the question, I would just leave the selector as is.

MaxDall (approved):
@addie9800 Thank's a lot for adding this one ❤️ And also taking part in this somewhat unusual experiment 🚀

PR #935, opened 2026-06-08
Enable impersonation in coverage script

MaxDall:
Apparently we forgot to enable impersonation in the coverage script with #920


addie9800 (approved):
Ah yes, good catch.

PR #939, opened 2026-06-09
Catch base curl_cffi Timeout instead of unused ReadTimeout subclass

MaxDall:
curl_cffi maps every libcurl timeout (`CURLE_OPERATION_TIMEDOUT`) to the base `Timeout` and never raises the vendored `ReadTimeout`/`ConnectTimeout` shims (both tagged `# not used`). The `except (..., ReadTimeout)` clauses carried over from the `requests` port therefore caught nothing, so a publisher timeout escaped the scraping task, was wrapped into a `RemoteException`, and crashed `publisher_coverage.py` despite `error_handling=suppress`.

Swaps `ReadTimeout` for the base `Timeout` in `html.py`, `url.py`, and `base_objects.py`. Since `ReadTimeout`/`ConnectTimeout` both subclass `Timeout`, this is strictly broader and future-proof.

### DIFF
--- a/src/fundus/publishers/base_objects.py
+++ b/src/fundus/publishers/base_objects.py
@@ -5,7 +5,7 @@ from warnings import warn
 
 import more_itertools
 from curl_cffi.requests import BrowserType, BrowserTypeLiteral
-from curl_cffi.requests.exceptions import ConnectionError, HTTPError, ReadTimeout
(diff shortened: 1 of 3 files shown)


addie9800 (approved):
Ah wow, good thing those classes exist then...

PR #940, opened 2026-06-16
Add agent-skills infrastructure and installer

MaxDall:
### What
Introduces a home for **agent skills** — playbooks/scripts meant to be executed by AI coding agents — under `skills/`, plus an installer that copies them into an agent's config directory.

The repo stays **agent-neutral**: we track skill *sources* here, but never commit any agent's config dir (`.claude/` is now gitignored). Each developer installs the skills they want into their own agent.

### Contents
- **`skills/install.py`** — thin launcher for the installer dashboard.
- **`skills/installer/core.py`** — stdlib-only install/uninstall + skill discovery (any folder with a `SKILL.md` is a skill). Install is **atomic**: if a skill's `requirements.txt` step fails, the copied folder is rolled back, so `✓` always means the skill *and* its deps landed.
- **`skills/installer/tui.py`** — Textual status dashboard: a live *skills × scopes* matrix (project `./.claude/skills/` vs user `~/.claude/skills/`); toggle to (un)install in place.
- **`skills/README.md`** — layout conventions, install flow, how to add a skill.
- **`pyproject.toml`** — adds `textual` to the `dev` extra (the dashboard's only dependency).
- **`.gitignore`** — ignores the installed-skills dir (`.claude/`).

### Notes
- Only **Claude Code** is wired up today; other agents are a documented future mapping in `available_agents()`.
- The first skill built on this — `review-publisher` — ships as a companion PR. This PR is the framework; on its own the installer matrix is empty until a skill folder exists.

### Testing
`python skills/install.py` opens the dashboard; toggling project/user installs and uninstalls the discovered skill folders.

### DIFF
--- a/skills/README.md
+++ b/skills/README.md
@@ -1,0 +1,89 @@
+# Agent skills
+
+Playbooks and skills meant to be **executed by AI coding agents**, not human contributors.
+
+The repo stays **agent-neutral**: we track the skill *sources* here, but we do **not** commit any
+agent's config directory (`.claude/` is gitignored). Each developer installs the skills they want into
+their own agent with the installer below.
+
+## Skills
+
+| Skill | What it does |
+|-------|--------------|
+| [`review-publisher`](review-publisher/SKILL.md) | Reviews a publisher PR (new publisher or parser-version change): crawls live articles to verify the extracted `ArticleBody` mirrors the real article, checks `VALID_UNTIL` / version bumps / `validate=False` / `free_access`, and drafts a single GitHub review. |
+
+`install.py` (the installer launcher) and the `installer/` package behind it, plus this
+`README.md`, round out the folder.
+
+## Layout
+
+Each skill is a **self-contained folder** — a `SKILL.md`, the `PLAYBOOK.md` it points at, and any helper
+scripts under `scripts/` — so the whole thing can be copied anywhere and still work. The installer copies
+the **entire folder**, so anything you drop in ships automatically; if the folder has a `requirements.txt`
+at its root, the installer also pip-installs it into your current interpreter. Install is **atomic**: if
+that pip step fails, the copied folder is rolled back (the cell flashes `!`, then shows `·`) so an install
+never half-lands — `✓` always means the skill *and* its deps are in.
+
+Two path conventions keep a copied folder working:
+
+- `SKILL.md` links to `PLAYBOOK.md` as a **sibling**, so the link survives being copied into `.claude/`.
+- **`${CLAUDE_SKILL_DIR}` is substituted by Claude Code in `SKILL.md` content only** — it is *not* an
+  environment variable, and other bundled files are read verbatim. So `SKILL.md` states the skill's own
+  directory once (via the placeholder) and tells the agent to substitute it into the `<skill>` paths
+  the `PLAYBOOK.md` commands use. A bare `scripts/…` path would wrongly assume the working dir is the
+  skill dir; a `${CLAUDE_SKILL_DIR}` outside SKILL.md would expand to nothing in a shell.
+
+References from `PLAYBOOK.md` to *other* repo docs use **repo-root-relative** links
+(e.g. `/docs/attribute_guidelines.md`): they resolve on GitHub and from the repo root, which is where
+the review skill runs anyway.
+
+## Install
+
+The installer is a full-screen **status dashboard**. Run it with no arguments:
+
+```bash
+python skills/install.py
+```
+
+It shows a live matrix of *skills* × *scopes* for the selected agent. Each row is a skill; the
+**Project** and **User** columns show whether it's installed (`✓`) or not (`·`). Move the cursor with
+`↑`/`↓`, then toggle a scope to act on the highlighted skill — `p` for project, `u` for user. Toggling
+installs the skill if it's absent and uninstalls it if it's present; the matrix updates in place, so you
+never have to re-run. `r` refreshes from disk and `q` quits.
+
+The two scopes:
+
+- **project** installs to `./.claude/skills/`. The skill is scoped to this repo, where it's relevant,
+  and never pushed (`.claude/` is in the committed `.gitignore`).
+- **user** installs to `~/.claude/skills/`, available in every project. Nothing touches the repo. (The
+  review skill references repo files, so it only makes sense while your working directory is the Fundus
+  repo regardless.)
+
+The installer needs the `textual` package, which ships in the project's `dev` extra:
+
+```bash
+pip install -e .[dev]
+```
+
+Re-install (toggle off, then on) after editing a skill source to refresh the installed copy. Restart
+your agent so it picks up a newly installed skill.
+
+## Supported agents
+
+| Agent | Status | Target |
+|-------|--------|--------|
+| **Claude Code** (`claude`) | supported | `.claude/skills/<name>/` |
+| Codex, Cursor, … | not yet | no 1:1 `SKILL.md` concept; would be a best-effort mapping (e.g. a Codex prompt or a `.cursor/rules` file). Add to `available_agents()` in `installer/core.py` when wanted. |
+
+## Adding a skill
+
+Create `skills/<name>/` with a `SKILL.md` (frontmatter `name`/`description`) and a `PLAYBOOK.md`
+it links to as a sibling; put helper scripts under `scripts/`. The installer **discovers skills from
+the folder layout** (any directory here containing a `SKILL.md`), so there is nothing to register —
+run `python skills/install.py` and the new skill shows up as a row. If the skill needs packages
+beyond Fundus, drop a `requirements.txt` at its root — the installer pip-installs it on install (and
+leaves it in place on uninstall).
+
+Keep the path conventions above: state the skill's own directory in `SKILL.md` via
+`${CLAUDE_SKILL_DIR}` (it is substituted nowhere else), and point cross-doc links at repo-root paths
+(`/docs/...`).
--- a/skills/README.md
+++ b/skills/README.md
@@ -45,19 +45,16 @@ The installer is a full-screen **status dashboard**. Run it with no arguments:
 python skills/install.py
 ```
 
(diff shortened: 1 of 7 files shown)


skills/install.py
```diff
@@ -0,0 +1,49 @@
...
+def main() -> int:
+    # Allow `python skills/install.py` to find the sibling `installer` package regardless
+    # of the current working directory.
+    sys.path.insert(0, str(Path(__file__).resolve().parent))
+    try:
+        from installer.tui import run
+    except ModuleNotFoundError as exc:
+        if exc.name in {"textual", "rich"}:
+            print(
+                "The skill installer dashboard needs the 'textual' package.\n"
```
addie9800: nit: It might be more helpful to not selectively test for textual and one of it's dependencies here, but rather either:
- always suggest running the dev extra install or
- check for all necessary dependencies (i.e. `platformdirs` is also essential) and 
    - explicitely name the missing package or 
    - rephrase the error message to say needs the 'textual' package and all dependencies.

skills/installer/tui.py
```diff
@@ -0,0 +1,204 @@
...
+        """Repopulate every row from disk for the current agent."""
+        agent = self._agent()
+        self.table.clear()
+        for name in sorted(self.skills):
+            skill = self.skills[name]
+            self.table.add_row(
+                Text(name, style="bold"),
+                _status(agent.is_installed("project", skill)),
+                _status(agent.is_installed("user", skill)),
+                Text(textwrap.shorten(skill.description, width=70, placeholder="…")),
```
addie9800: One could consider making the width dynamic here, when supported by the screen and if it doesn't complicate the code too much. For me for example, there is a lot of unnecessary dead space:

Image

skills/installer/core.py
```diff
@@ -0,0 +1,219 @@
...
+            return InstallResult(False, [*req_log, f"  rolled back '{skill.name}' ({scope}) — nothing left installed"])
+        head = f"installed '{skill.name}' for {self.name} ({scope}) -> {dest / 'SKILL.md'}"
+        return InstallResult(True, [head, *req_log])
+
+    def uninstall(self, scope: str, skill: Skill) -> List[str]:
+        dest = self.destination(scope, skill)
+        if dest.exists():
+            shutil.rmtree(dest)
+            return [f"removed '{skill.name}' for {self.name} ({scope}) -> {dest}"]
+        return [f"nothing installed at {dest}"]
```
addie9800, line 122: I would also expect all the packages, that were installed and are now obsolete to be uninstalled. When typing this out, I realized that one would have to ensure that no requirement is removed that is still required by other skills. On a similar note: not really a big issue for now, but with more skills, one might eventually encounter issues with contradicting (version) requirements for skills. As of now, only the most recently installed skill will be available, right?

skills/installer/core.py
```diff
@@ -0,0 +1,219 @@
...
+    1:1 ``SKILL.md`` concept; add them here when wanted.
+    """
+    return {"claude": Agent("claude", _claude_skills_root)}
+
+
+# --- helpers ---
+
+
+def _install_requirements(skill: Skill) -> Tuple[bool, List[str]]:
+    """If the skill bundles a ``requirements.txt``, pip-install it into the current interpreter.
```
addie9800, line 143: Could you perhaps provide a couple more details for the use-case of the user scope? Usually the skills we install will probably be quite project specific and unlikely to be required on the user-level. In case required, the execution can still be problematic from what I can see, because the requirements are always installed from the project-level interpreter. Even if the user-level is specified, if a venv/conda env are used when running this installer, the deps will not be installed globally, if I am not mistaken.

skills/README.md
```diff
@@ -0,0 +1,89 @@
...
+## Supported agents
+
+| Agent | Status | Target |
+|-------|--------|--------|
+| **Claude Code** (`claude`) | supported | `.claude/skills/<name>/` |
+| Codex, Cursor, … | not yet | no 1:1 `SKILL.md` concept; would be a best-effort mapping (e.g. a Codex prompt or a `.cursor/rules` file). Add to `available_agents()` in `installer/core.py` when wanted. |
+
+## Adding a skill
+
+Create `skills/<name>/` with a `SKILL.md` (frontmatter `name`/`description`) and a `PLAYBOOK.md`
```
addie9800: Having not too much experience with Markdown, I didn't know what exactly you meant with frontmatter until looking into the review-skill PR. Maybe a template or short example might make adding the first skill simpler.

skills/installer/core.py
```diff
@@ -0,0 +1,219 @@
...
+
+def _frontmatter_description(skill_md: Path) -> str:
+    """Pull the one-line ``description`` from a SKILL.md's YAML frontmatter.
+
+    Handles both inline (``description: foo``) and block-scalar (``description: >-``) forms;
+    returns ``""`` when the file is missing or has no frontmatter description.
+    """
+    try:
+        text = skill_md.read_text(encoding="utf-8")
+    except OSError:
```
addie9800: This is a bit too broad. `FileNotFoundError` might be a better choice to allow other exceptions to be raised.

skills/installer/core.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+        dest = self.destination(scope, skill)
+        if dest.exists():
+            shutil.rmtree(dest)  # refresh in place
+        dest.parent.mkdir(parents=True, exist_ok=True)
+        shutil.copytree(skill.source, dest)
+        req_ok, req_log = _install_requirements(skill)
+        if not req_ok:
+            shutil.rmtree(dest, ignore_errors=True)  # undo the copy — keep the matrix honest
+            return InstallResult(False, [*req_log, f"  rolled back '{skill.name}' ({scope}) — nothing left installed"])
```
addie9800: Also here, when rolled back, I would also assume the successful installations to be rolled back as well.

skills/installer/core.py
```diff
@@ -0,0 +1,219 @@
...
+
+    pip prints its "A new release of pip is available / To update, run …" upgrade notice
+    *last*, so a naive last-N-lines tail shows that noise instead of the real failure. We
+    drop that notice and, when pip emitted its own ``ERROR:`` lines, surface those; otherwise
+    we fall back to the last few non-empty lines.
+    """
+    noise = ("A new release of pip", "To update, run")
+    lines = [ln.rstrip() for ln in f"{stdout}\n{stderr}".splitlines() if ln.strip()]
+    lines = [ln for ln in lines if not any(n in ln for n in noise)]
+    errors = [ln for ln in lines if ln.lstrip().startswith("ERROR")]
```
addie9800: I would include WARNING level as well - might make debugging easier for the user:

```
WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x7fa8d8545450>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution')': /simple/numpy/
ERROR: Could not find a version that satisfies the requirement numpy (from versions: none)
ERROR: No matching distribution found for numpy
```

skills/installer/tui.py
```diff
@@ -0,0 +1,204 @@
...
+    TITLE = "Fundus Skill Installer"
+
+    # The matrix is the point of the screen: it gets the flexible space *and* a minimum,
+    # so a short terminal shrinks the log, never the skills. The agent panel only exists
+    # when there is an actual choice to make (see compose).
+    CSS = """
+    DataTable { height: 1fr; min-height: 6; margin: 1 1 0 1; border: round $primary; }
+    #agent { height: auto; border: round $primary; padding: 0 1; margin: 1 1 0 1; }
+    #agent > RadioButton { width: auto; }
+    RichLog { height: 5; border: round $panel; padding: 0 1; margin: 1; }
```
addie9800: I think it would be good to increase the size of the log. For longer error messages (e.g. version mismatch in numpy) it is very difficult to read and get an overview of the output.

addie9800 (changes requested):
This is a very promising in new step towards making fundus more skalable. I am looking forward to seeing the progress this allows us to make 🚀

skills/install.py
MaxDall: Done — any missing dev-extra module (incl. `platformdirs`) now triggers the message, and it names the actual missing package via `exc.name`.

skills/installer/tui.py
MaxDall: Textual's `DataTable` has no fill-column, so instead of computing a width I dropped the Description column: the **full** description now shows in a detail panel below the table, updated as you move the cursor. No truncation, no dead space.

skills/installer/core.py
MaxDall: Agreed this is thorny — shared deps and version conflicts make safe removal really an env-manager's job. We keep deps additive to your project env and leave them on uninstall; documented that trade-off.

skills/installer/core.py
MaxDall: Good push — there's no strong use case for user scope with repo-bound Fundus skills, and the venv-vs-global mismatch you spotted made it worse. Removed the user scope entirely; install is now project-only.

skills/README.md
MaxDall: Added a folder-layout tree and a minimal `SKILL.md` frontmatter example to the "Adding a skill" section.

skills/installer/core.py
MaxDall: Narrowed to `except FileNotFoundError`.

skills/installer/core.py
MaxDall: Same reasoning as the uninstall thread: we roll back the copied files but leave pip packages (removing shared deps is unsafe). Also fixed the log line so it no longer implies nothing was installed.

skills/installer/core.py
MaxDall: Now surfaces `WARNING` lines too (they often explain the error), and bumped the tail 3 → 5 so multi-line failures aren't clipped.

skills/installer/tui.py
MaxDall: Increased the log height 5 → 10.

addie9800 (approved):

PR #941, opened 2026-06-16
Add review-publisher skill

MaxDall:
### What
Adds the **`review-publisher`** agent skill: it reviews a Fundus publisher PR (a new publisher, or an added/changed parser version) by crawling **live articles** and verifying the extracted `ArticleBody` mirrors the real article — no dropped paragraphs, no leaked boilerplate — then drafts a single GitHub review.

Built on the skills framework from its companion PR #940, but **independent of it**: the scripts run standalone (`python skills/review-publisher/scripts/review.py …`) and don't import the installer.

### How it works
A crawl-once-then-sweep flow is a hard gate on every publisher under review:
- **`scripts/review.py`** — the driver (`crawl`/`sweep`/`show`/`adjudicate`/`status`/`payload`). Every candidate the sweep surfaces must be explicitly adjudicated, and `status` must report `READY` before any verdict.
- **`scripts/_sweep.py`** — diffs the extracted body against the live document to catch missing/leaked content.
- **`scripts/_store.py`** — persists crawl/adjudication state across invocations.
- **`scripts/sampler/`** — layout sampling, so the crawl picks structurally diverse articles instead of near-duplicates.
- **`SKILL.md` / `PLAYBOOK.md`** — the agent-facing entry point and step-by-step procedure.

### Library change
Adds **`BaseParser.body_selectors()`** (`src/fundus/parser/base_parser.py`) — a public accessor returning a version's `summary` / `subheadline` / `paragraph` selectors. This gives external tooling (the skill) a supported way to read them instead of reaching into the private `_*_selector` attributes.

### Testing
`tests/test_review_skill.py` pins the sweep gate's failure modes — above all, the ways it could report a silent false "clean".

### DIFF
--- a/skills/review-publisher/PLAYBOOK.md
+++ b/skills/review-publisher/PLAYBOOK.md
@@ -1,0 +1,307 @@
+# How to Review a Publisher PR
+
+> **Audience: AI coding agents.** This is an operational playbook meant to be executed by an agent,
+> not a human contributor guide. It's also the source doc behind the `review-publisher` skill.
+
+A review playbook for PRs that **add a new publisher** or **add/change a parser version**. The matching
+authoring process is [`how_to_add_a_publisher.md`](/docs/how_to_add_a_publisher.md) — you're reviewing the
+fruits of it, so the checks below map onto the promises that process makes.
+
+All mechanics run through one driver. `<skill>` below is the skill directory SKILL.md told you
+(`${CLAUDE_SKILL_DIR}` is substituted only there); the cross-doc links above and below are
+repo-root-relative — read them from the repo root, where you are anyway.
+
+    python "<skill>/scripts/review.py" {crawl,sweep,show,adjudicate,status,payload} <cc>.<Class>
+
+The driver enforces the bookkeeping (what was crawled, what was swept, what is still un-adjudicated);
+this playbook is about the part that stays yours: **judgment**. The split is deliberate — the driver
+may refuse, warn, and count, but it never decides; you never count by hand.
+
+**The unit of review is the PR's diff, not the whole publisher.** Review what *this PR* changes. On a
+parser-version bump, scope your reading to the new version and how it differs from the old — don't
+re-litigate code the PR doesn't touch. A new publisher is all-new, so there the diff *is* the whole file.
+
+**A multi-publisher PR is N independent reviews, not one.** Run §2 once per publisher — each gets its
+own crawl, sweep, and `status` READY. A blocker on one publisher does **not** discharge the body checks
+on the others: the loud finding is a distraction from the clean-reading publishers, which get the *same*
+full treatment, never a lighter pass.
+
+## What you're protecting
+
+> **The extracted `ArticleBody` must mirror the real article — no missing content, no extra content.**
+
+This isn't aesthetic. Fundus maps extracted text back to the source HTML for annotation, so a dropped
+paragraph or a leaked photo caption corrupts that mapping. Over-capture is a blocker, not a nit.
+
+## Rules
+
+- **Don't run `pytest`, `mypy`, `ruff`, or any other checks locally — the GitHub CI covers all of
+  that.** Your value-add is confirming the parser is correct on **live** articles; the unit test only
+  checks one frozen HTML snapshot, so a green test proves nothing about today's site or other layouts.
+- **Cite everything.** Every finding needs quotable evidence: the verbatim dropped/leaked text, the
+  article URL, and the offending selector. Adjudication notes are part of that trail — they land in
+  `findings.json` verbatim.
+- **Keep scratch out of the repo.** `review.json` and any throwaway dump go in the cache dir the
+  driver prints (`<tempdir>/fundus-review/…`) or any OS temp dir — never the working directory.
+- **Never truncate driver output.** Don't pipe `crawl` through `[:N]` or cap the printed text — a
+  slice that cuts off partway through the draw reads clean on articles you never saw. The state file
+  records what was actually cached, and `status` reconciles the counts for you.
+
+## Inputs
+
+- **PR number** and whether it's **your own** PR (changes the verdict event — see §4). **If the user
+  didn't name a PR, resolve it before doing anything else:** `gh pr view --json number,title,author`
+  reads the current branch's open PR; show it and confirm. If the branch has no open PR, ask — don't
+  guess. (The `author` also settles the own-PR question.)
+- The publisher(s) under review as `PublisherCollection.<cc>.<Class>`, and the parser version(s) touched.
+
+## 1. Static read
+
+```bash
+gh pr checkout <PR_NUMBER>
+gh pr diff <PR_NUMBER>
+```
+
+Skim the diff and sanity-check the parser:
+
+- `@attribute` return types match [`attribute_guidelines.md`](/docs/attribute_guidelines.md).
+- **`@attribute(validate=False)` bypasses CI.** Type conformance is enforced by the unit tests for
+  *validated* attributes only. Any attribute marked `validate=False` is checked by **nobody but you** —
+  read its logic and return value by hand.
+- **`free_access` / paywall.** If the publisher runs a subscription model, `free_access` must be
+  implemented (off `isAccessibleForFree` in the `ld+json`). A missing or wrong `free_access` on a
+  premium publisher is a blocker. If the publisher is fully free, there's nothing to check here.
+- **`VALID_UNTIL` is *not* inherited** — every `BaseParser` subclass defaults to `date.max`. When a PR
+  adds a new version for a layout change, the *previous* version must get an explicit `VALID_UNTIL`
+  (the day before the change), and the newest must leave it unset. (Subclassing another publisher's
+  version and leaving `VALID_UNTIL` unset is fine — that's the intended `date.max`.)
+- **Version bump fits the change.** Selector-only fix → minor bump (`V1_1(V1)`); new or substantially
+  changed attributes → major bump (`V2`). Flag a `V2` that's really just selectors, or a minor bump
+  that quietly changes attribute behavior.
+- **Shared-utility changes** (`parser/utility.py`: `generic_topic_parsing`, `apply_result_filter`,
+  `image_extraction`, …) affect **every** publisher, not just this one. Check the call sites and rely
+  on the full (CI-run) suite staying green.
+
+## 2. Crawl live articles and verify the body
+
+### Crawl once
+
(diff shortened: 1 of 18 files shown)


skills/review-publisher/scripts/sampler/_features.py
```diff
@@ -0,0 +1,132 @@
...
+import numpy as np
+from lxml.html import HtmlElement
+from sklearn.feature_extraction.text import TfidfVectorizer
+from sklearn.metrics.pairwise import cosine_distances
+
+# Tags carrying no layout meaning (or only noise) — skipped when building tag paths.
+_SKIP_TAGS = frozenset({"script", "style", "noscript", "template", "svg", "path", "br", "wbr"})
+
+DEFAULT_TEXT_SHARE = 0.7  # tightest container holding >= this share of body text (validated)
+_MAX_LINK_DENSITY = 0.5  # a block whose text is >this fraction inside <a> is boilerplate
```
addie9800:
```suggestion
_MAX_LINK_DENSITY = 0.5  # a block whose text is > this fraction inside <a> is boilerplate
```

skills/review-publisher/scripts/sampler/sampler.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class Sampler:
+    """Reduce a crawled pool of articles to a layout-diverse sample.
+
+    Parameters
+    ----------
+    text_share     : body-isolation tightness for the density finder (0.7 validated; <0.9).
+    k_max          : cap on the number of layouts the clustering may find (mode 2).
+    max_path_len   : optionally coarsen tag paths to the last k tags (None = full path).
```
addie9800: nit: avoid confusion with parameter clustering parameter `k`
```suggestion
    max_path_len   : optionally coarsen tag paths to the last m tags (None = full path).
```

skills/review-publisher/scripts/_store.py
```diff
@@ -0,0 +1,196 @@
...
+    """Wipe and recreate `cache_dir` for a fresh crawl — refusing anything that isn't a review cache.
+
+    The guard is what makes `--cache-dir` safe: a non-empty directory without a
+    `state.json` (someone's working tree, a typo'd path) is never deleted.
+    """
+    if cache_dir.exists():
+        if any(cache_dir.iterdir()) and not (cache_dir / STATE_FILE).exists():
+            raise SystemExit(
+                f"refusing to wipe {cache_dir}: non-empty and no {STATE_FILE}, so it doesn't look like "
+                f"a review cache. Use an empty or not-yet-existing --cache-dir."
```
addie9800:
```suggestion
                f"a review cache. Use a valid or not-yet-existing --cache-dir."
```

skills/review-publisher/scripts/_store.py
```diff
@@ -0,0 +1,196 @@
...
+    newline-translate on Windows (\\r\\n -> \\r\\r\\n on disk).
+    """
+    (cache_dir / html_filename(index)).write_bytes(article.html.content.encode("utf-8"))
+    body = article.body
+    return {
+        "index": index,
+        "url": article.html.requested_url,
+        "crawl_date": article.html.crawl_date.isoformat(),
+        "title": article.title,
+        "authors": list(article.authors),
```
addie9800: Calling these list constructors should be redundant.

skills/review-publisher/scripts/_store.py
```diff
@@ -0,0 +1,196 @@
...
+        gaps.append("no articles in the cache — 0 crawled is itself a blocker-level finding (see PLAYBOOK §2)")
+    sweep = state.get("sweep")
+    if not sweep:
+        gaps.append("no sweep recorded — run `sweep`")
+    else:
+        if crawl.get("finished") and sweep.get("swept_at", 0) < crawl["finished"]:
+            gaps.append("the sweep predates the last crawl — re-run `sweep`")
+        pending = pending_candidates(state)
+        if pending:
+            ids = ", ".join(c["id"] for c in pending[:15])
```
addie9800: If more than 15, it might be nice to indicate that with `, ...` or something similar.

src/fundus/parser/base_parser.py
```diff
@@ -233,6 +234,22 @@ def __init_subclass__(cls, **kwargs):
...
+        """The version's body selectors, keyed 'summary' / 'subheadline' / 'paragraph'.
+
+        These are the selectors a version feeds to ``extract_article_body_with_selector``;
+        a value is None when the version does not declare the corresponding selector
+        (or builds its body another way entirely). Public so external tooling (e.g. the
+        review skill under skills/) does not have to reach into the private
+        ``_*_selector`` attributes.
+        """
+        return {
+            "summary": getattr(cls, "_summary_selector", None),
```
addie9800, line 253: I think up until now, we do not enforce this naming convention for the content selectors, this might be unreliable. Perhaps we should add a test for it, if we want this to work reliably?

tests/test_review_skill.py
```diff
@@ -0,0 +1,201 @@
...
+
+        cache = tmp_path / "cache"
+        cache.mkdir()
+        (cache / _store.STATE_FILE).write_text("{}", encoding="utf-8")
+        (cache / "01.html").write_text("x", encoding="utf-8")
+        _store.prepare_cache_dir(cache)  # a real cache is wiped and recreated
+        assert cache.exists() and not any(cache.iterdir())
+
+        _store.prepare_cache_dir(tmp_path / "fresh")  # nonexistent is simply created
+        assert (tmp_path / "fresh").is_dir()
```
addie9800: Even if it is in the temp directory, we could clean up the files after execution.

skills/review-publisher/PLAYBOOK.md
```diff
@@ -0,0 +1,307 @@
...
+- **Over-capture**: boilerplate or captions leaking *into* the body.
+
+For images, compare each `caption`/`authors`/`is_cover` against the live `<figure>` to confirm they're
+paired with the right image and that prefixes like `"Photo by "` are stripped — these map to the
+parser's `caption_selector`, `author_selector` (whose `credits` named group is stripped from the
+caption), and the `upper_`/`lower_boundary_selector` cover boundaries. Name the offending selector when
+you report.
+
+To see articles the extraction filter *dropped* entirely (not just mis-parsed), re-crawl with
+`crawl(max_articles=..., only_complete=False)` in a scratch snippet — incomplete articles then show up
```
addie9800: I was wondering, whether the default crawl should already be run with `only_complete=False`.

skills/review-publisher/scripts/review.py
```diff
@@ -0,0 +1,472 @@
...
+    state = new_state(args.publisher, args.pool)
+    write_state(cache_dir, state)
+
+    pool: List[Article] = []
+    completed = False
+    try:
+        # The only networked step: draw a candidate pool, then reduce it to a layout-diverse subset.
+        # Sampling needs the whole pool up front, so unlike the per-article cache this buffers in
+        # memory — an interrupted crawl caches nothing and is simply re-run (it's the cheap step).
+        pool = list(Crawler(publisher).crawl(max_articles=args.pool))
```
addie9800: As mentioned in another comment, I was wondering, whether or not it might make sense to add `only_complete=False` here.

skills/review-publisher/PLAYBOOK.md
```diff
@@ -0,0 +1,307 @@
...
+
+```bash
+python "<skill>/scripts/review.py" sweep <cc>.<Class>                 # same cache, no network
+python "<skill>/scripts/review.py" sweep <cc>.<Class> --version V1_1  # pin a legacy version; re-runs free
+```
+
+The sweep applies the version's *real* body selectors to each cached article and emits two kinds of
+candidate, each with an id:
+
+- **DROP** — a structural block (`table`/`ul`/`ol`/`dl`/`blockquote`/`pre`) or `<p>`/`<h*>` in the
```
addie9800: If I understand this correctly, if these tags are missing, then it will result in a block. As fundus does not officially support tables, I guess this should be removed as well. Also blockquote objects are sometimes used to highlight sections of the text and it has to be decided on a case-to-case basis, whether this is a drop or intended.

skills/review-publisher/scripts/_store.py
MaxDall: Right — both are already `List[str]`. Dropped.

skills/review-publisher/scripts/_store.py
MaxDall: Added `, ...` when the list is truncated; the total count was already in the message.

src/fundus/parser/base_parser.py
MaxDall: Agreed, and it had already drifted: five parsers passed their subheadline selector under a non-canonical name (`zdf`, `asahi_shimbun`, `anadolu_ajansi`, `haberturk`, `the_sun`), so `body_selectors()` silently read `None` for them. Renamed all five and added `TestParser::test_body_selector_naming`, which AST-checks every version's call.

Longer term the convention shouldn't be a convention at all — #958 proposes declaring the selectors on `BaseParser`, which retires the test.

tests/test_review_skill.py
MaxDall: Added a `cache_root` fixture that wraps `tmp_path` and removes it afterwards.

skills/review-publisher/scripts/review.py
MaxDall: Done. One knock-on worth flagging: a body-less article would make the sweep report every block on the page as a drop candidate, so the sweep now returns N/A for it — same lane as "no `_paragraph_selector`" — and `find_leaks` skips articles with no units so they don't inflate the repetition threshold.

skills/review-publisher/PLAYBOOK.md
MaxDall: Done — see the reply on `review.py`. This paragraph is gone: the default draw already includes those articles, so there is nothing to re-crawl for.

skills/review-publisher/PLAYBOOK.md
MaxDall: Fixed the guidance rather than the detection. The bullet now spells out the adjudication per tag: `table` is normally `ok` since `ArticleBody` has no representation for tabular content, and `blockquote` is case-by-case.

I kept both reported because a `table` only reaches the drop scan when *none* of its text was captured (any ancestor of a captured node counts as captured), so the case it catches is prose inside a cell that the selector missed. Also replaced the `blocker` example, which used a dropped table.

addie9800 (dismissed):

PR #942, opened 2026-06-19
Add impersonation for several protected publishers

MaxDall:


addie9800 (approved):
Thanks for updating these as well 👍

PR #943, opened 2026-07-04
Nine news update

addie9800:
Note that this PR also introduces support for domain migration for publishers. This is mainly necessary to ensure that backward crawling can be run and aggregated together with the same publisher instance as a forward crawl.

Pytest is failing due to a Wikipedia layout change fixed in #944, this PR should be merged after.


src/fundus/publishers/au/nine_news.py
```diff
@@ -53,3 +55,44 @@ def images(self) -> List[Image]:
                 caption_selector=XPath("./ancestor::figure//figcaption/text()[1]"),
                 author_selector=XPath("./ancestor::figure//figcaption/text()[last()]"),
             )
+
+    class V2(BaseParser):
+        _paragraph_selector = XPath("//div[@class='sc-iKFaHN gHSkdP']/p")
```
MaxDall: **Nit — volatile selectors.** `sc-iKFaHN gHSkdP` is a styled-components hash that changes on any redeploy (body then silently empties). `//div[@data-ui-testid='body-content-all']/div` yields identical output (verified) — and you already use `data-ui-testid` for `topics`/`images`.

src/fundus/publishers/base_objects.py
```diff
@@ -142,6 +142,7 @@ def __init__(
         disallows_training: bool = False,
         suppress_robots: bool = False,
         impersonate: Optional[BrowserTypeLiteral] = None,
+        deprecated_domains: List[str] = [],
```
MaxDall: **Nit — mutable default arg.** `deprecated_domains: List[str] = []` shares one list across all `Publisher`s. Read-only today, but prefer `Optional[List[str]] = None` with `or []`.

MaxDall (commented):
`au.NineNews` V2 + domain migration — bodies verified coherent on 12 live articles. Two nits inline.

MaxDall (approved):

PR #947, opened 2026-07-11
fix(publisher): update upper boundary selector for WDR parser

addie9800:
Adjust XPath to include additional class for enhanced parsing accuracy.

- Extend `upper_boundary_selector` to handle 'sectionArticle' class
- Maintain compatibility with existing `segment` class selection


MaxDall (approved):
@addie9800 Thank for fixing ❤️

PR #951, opened 2026-07-12
feat(HankookIlbo): introduce parser version 3

addie9800:


src/fundus/publishers/kr/hankook_ilbo.py
```diff
@@ -144,3 +147,46 @@ def images(self) -> List[Image]:
                 caption_selector=XPath("./ancestor::div[@class='editor-img-box']//div[@class='caption']"),
                 author_selector=re.compile(r"(?!.+\.)(.+=|\.)?(?P<credits>.+?)( 제공| 기자)?$"),
             )
+
+    class V3(BaseParser):
+        _paragraph_selector = XPath("//div[@class='article-view']/p[text()]")
```
MaxDall: **Blocker — `p[text()]` drops paragraphs whose text is entirely inside inline tags.** The predicate needs a *direct* text-node child, so `<p class='editor-p'><span>…</span><span>…</span></p>` never matches. On [[1]](https://www.hankookilbo.com/news/article/A2026073010100004523) 212 chars vanish: "…정보를 전쟁 중인 국가에 판매하는 것도 가능하다. WSJ는 중국 스타트업 기업 '미자르비전'의…"; on [[2]](https://www.hankookilbo.com/news/article/A2026073015560000620) the lede quote `<p class='editor-p'><strong>"저는 서울에서 태어났지만…"</strong></p>` goes, leaving the body to open "…밝힌 첫 일성이다" with nothing to refer to.

Fix: `//div[@class='article-view']/p[normalize-space()]` — still skips the `<p><br></p>` spacers, and across all 12 crawled articles it recovers exactly these two paragraphs and loses none.

src/fundus/publishers/kr/hankook_ilbo.py
```diff
@@ -144,3 +147,46 @@ def images(self) -> List[Image]:
...
+        _subheadline_selector = XPath("//div[@class='article-view']/h3")
+        _summary_selector = XPath("//div[@class='article-view']/h2")
```
MaxDall, line 156: **Blocker — `<br>`-separated summary lines are fused, inventing word boundaries.** `<h2 class='sub-tit-ll'>` holds several bullets split by `<br>`; taking its text content concatenates them: "…대피 일부 귀가 **허용스페인도** 국가 비상사태…" (all 12 crawled articles; the new fixture bakes it in as one string, where `V2`'s has three).

`V2` handled exactly this with `transform_breaks_to_tag(node, tag="h2", replace=True)` — already imported. Fix: re-add it as a `@function(priority=1)` over the `h2`, and over `p.editor-p` too: on [[1]](https://www.hankookilbo.com/news/article/A2026073017410000445) the entire article is one `<br>`-separated `<p>` and comes out as a single run-on paragraph.

src/fundus/publishers/kr/hankook_ilbo.py
```diff
@@ -144,3 +147,46 @@ def images(self) -> List[Image]:
...
+
+        _topics_selector = XPath("//div[@class='flex flex-wrap gap-8']//a")
+
+        @attribute
+        def body(self) -> Optional[ArticleBody]:
```
MaxDall, line 165: **Blocker — quoted source material in `div.module-*` is covered by no selector.** These modules sit inside `div.article-view` but hold their text in `<blockquote>`/nested `<p>`, so nothing catches them. On [[1]](https://www.hankookilbo.com/news/article/A2026073016010000916) the quoted letter goes and the body opens "…건넨 편지의 일부 내용이다" — *is part of the letter* — with no letter; on [[2]](https://www.hankookilbo.com/news/article/A2026073014550003971) it is 2,253 chars, the scholars' statement reproduced in full under `[전문]`.

Fix: extend `_paragraph_selector` to reach `div.module-quote`/`div.module-stance` content (and `_subheadline_selector` to their `<h3>`), keeping the sibling `div.module-series` teaser list out.

src/fundus/publishers/kr/hankook_ilbo.py
```diff
@@ -144,3 +147,46 @@ def images(self) -> List[Image]:
                 caption_selector=XPath("./ancestor::div[@class='editor-img-box']//div[@class='caption']"),
                 author_selector=re.compile(r"(?!.+\.)(.+=|\.)?(?P<credits>.+?)( 제공| 기자)?$"),
             )
+
+    class V3(BaseParser):
+        _paragraph_selector = XPath("//div[@class='article-view']/p[text()]")
+        _subheadline_selector = XPath("//div[@class='article-view']/h3")
+        _summary_selector = XPath("//div[@class='article-view']/h2")
+
+        _topics_selector = XPath("//div[@class='flex flex-wrap gap-8']//a")
```
MaxDall: **Nit — `_topics_selector` is dead code.** `topics` reads the `article:tag` meta tag instead, and nothing else references this XPath. Drop it, or switch `topics` over to it.

MaxDall (changes requested):
**REQUEST_CHANGES** — three body-correctness blockers in the new `V3`.

**`kr.HankookIlbo`** — read 12 of 120 scanned (31 flagged by the offline scan, 11 of them in the draw); layouts covered: straight news, politics/analysis with subheads, sports, image-heavy, plus three non-prose pages (horoscope ×2, editorial cartoon) that correctly yield no body. One-off over-capture: none — nothing outside the article container leaked into a body, and image `caption`/`authors`/`is_cover` check out against the live `<figure>`s (credits like `AP 뉴시스`, `공항사진기자단` are stripped into `Image.authors`). `V2`'s `VALID_UNTIL` and the missing `VALID_UNTIL` on `V3` are correct in shape; the `data.py` empty-tag guard is a sound generalisation.

Blockers (evidence in the inline threads):

- `_paragraph_selector` uses `p[text()]`, which drops every `<p class='editor-p'>` whose text is entirely inside `<span>`/`<strong>` — real prose, and in one case the lede quote the next paragraph refers back to. `hankook_ilbo.py:152`.
- `<br>`-separated summary lines are concatenated without a separator, fusing words across the break on **every** article. `V2` solved this with `transform_breaks_to_tag`; `V3` dropped that step. `hankook_ilbo.py:154`.
- Quoted source material in `div.module-quote` / `div.module-stance` is covered by no `V3` selector — up to 2.2k chars of a quoted statement lost, and a body left opening on a dangling reference. `hankook_ilbo.py:159`.

Nits: dead `_topics_selector` (`:156`); `@class='article-view'` is an exact match, so the `article-view luck-new` variant yields an empty body (`:152`); `<dl class='annotation-list'>` footnotes are dropped while the body keeps the marker (`효과1`); `authors` off `article:author` returns the publisher name (`['한국일보']`) on wire/cartoon pages and unsplit strings (`박준석 기자 외 1명`, in the new fixture) where `V2`'s `writer` selector returned people. Also worth a unit test on the `data.py` empty-tag path.

One question: what dates `V2`'s `VALID_UNTIL = 2026-06-01`? The two fixtures bracket it at 2026-01-29 / 2026-07-12, so the boundary looks estimated rather than observed.

src/fundus/publishers/kr/hankook_ilbo.py
addie9800: The second case seems to be an info box that we usually ignored. I have ignored it for now as well. The blockquote case has been updated.

MaxDall (approved):

MaxDall (approved):
@addie9800 Thanks for updating the parser! Looks really good 🚀

PR #953, opened 2026-07-18
Update aftonbladet

addie9800:


src/fundus/publishers/se/aftonbladet.py
```diff
@@ -56,7 +65,17 @@ def topics(self) -> List[str]:
...
                 paragraph_selector=self._paragraph_selector,
-                caption_selector=XPath("./ancestor::figure//figcaption/span[@class='image-caption']"),
-                author_selector=XPath("./ancestor::figure//figcaption/span[contains(@class,'image-byline')]"),
+                caption_selector=self._caption_selector,
+                author_selector=self._image_author_selector,
             )
+
+    class V1_1(V1):
+        _summary_selector = XPath("(//header)[2]/p")
+        _paragraph_selector = XPath("(//section[@class='article-body'])[1]/p")
```
MaxDall: **Blocker — `<ul>` lists inside the body are dropped.** The p-only selector leaves the body reading "Nyckelfunktioner:" followed by nothing; the list that belongs there — "Självrengörande dockningsstation som tömmer dammbehållaren, tvättar mopparna och fyller på vatten automatiskt…" — is missing (9 such `<ul class='_list_16k26_3 … layout-component'>`, all direct children of `section.article-body`, in this one article). Fix: extend `_paragraph_selector` with `| (//section[@class='article-body'])[1]/ul/li`. [[1]](https://www.aftonbladet.se/harligthemma/a/8pXj0W/robotdammsugare-tips-2026-dreame-roborock)

src/fundus/publishers/se/aftonbladet.py
```diff
@@ -56,7 +65,17 @@ def topics(self) -> List[str]:
...
+        _subheadline_selector = XPath("(//section[@class='article-body'])[1]/h2")
+
+        _caption_selector = XPath("./ancestor::figure//figcaption/span")
```
MaxDall: **Blocker — gallery figures extract neither caption nor credits.** Their caption sits directly in the element, with no `<span>` child for `/span` to match: `<figcaption class='_srOnly_…'>Över 49 000 har tagit sig över gränsen från Marocko senaste dygnen. <cite>Foto: Antonio Sempere /AP/TT</cite></figcaption>`. 37 of the 42 images across the draw come back `caption=None, authors=[]` (all 13 in [[1]](https://www.aftonbladet.se/mode/a/ExBmAA/brunt-mode-host-2026), all 11 in the robot-vacuum guide). Fix: point `_caption_selector` at the `figcaption` element itself. [[2]](https://www.aftonbladet.se/nyheter/a/wr9JXd/akesson-spanien-maste-stanga-sin-grans)

src/fundus/publishers/se/aftonbladet.py
```diff
@@ -56,7 +65,17 @@ def topics(self) -> List[str]:
...
+        _image_author_selector = re.compile(r"(?i)foto:\s*(?P<credits>.*)\s*$")
```
MaxDall, line 76: **Blocker — the "Visa mer" toggle leaks into image credits.** `figcaption` ends with `<span class='_showMore_…' hidden>Visa mer</span>`, which the caption selector picks up and `(?P<credits>.*)\s*$` then swallows: every non-empty `authors` in the draw ends that way, e.g. `['Emma Da Silva/AP/TT Visa mer']` (also 4 more), and the fixture added here records it for all six images (`"Robin Lorentz Allard Visa mer"`). Fix: exclude `span[contains(@class,'showMore')]` in `_caption_selector`. [[1]](https://www.aftonbladet.se/nyheter/a/oE1xaB/analys-klimatforandringar-gjorde-branderna-varre)

src/fundus/publishers/se/aftonbladet.py
```diff
@@ -15,12 +16,20 @@
...
         )
         _subheadline_selector = XPath("//h2[@data-test-tag='paragraph-header']")
 
+        _caption_selector = XPath("./ancestor::figure//figcaption/span[@class='image-caption']")
+        _image_author_selector: Union[XPath, Pattern[str]] = XPath(
+            "./ancestor::figure//figcaption/span[contains(@class,'image-byline')]"
+        )
+        _image_selector = XPath("//figure//img")
+
         _paywall_selector = XPath("//main/vev")
```
MaxDall: **Nit — `_paywall_selector` is inherited unchanged and I could not confirm it still marks the new layout.** `//main/vev` matches nothing in any of the 10 crawled pages, but all 10 are free (`"isAccessibleForFree": true`), so `free_access=True` is correct for them and the selector stays untested against a Plus article. Worth verifying on one — the new layout carries `isAccessibleForFree` in its `ld+json`.

src/fundus/publishers/se/aftonbladet.py
```diff
@@ -56,7 +65,17 @@ def topics(self) -> List[str]:
...
+                image_selector=self._image_selector,
                 paragraph_selector=self._paragraph_selector,
-                caption_selector=XPath("./ancestor::figure//figcaption/span[@class='image-caption']"),
-                author_selector=XPath("./ancestor::figure//figcaption/span[contains(@class,'image-byline')]"),
+                caption_selector=self._caption_selector,
+                author_selector=self._image_author_selector,
             )
+
+    class V1_1(V1):
+        _summary_selector = XPath("(//header)[2]/p")
```
MaxDall, line 69: **Nit — positional selectors where a class is available.** `(//header)[2]` and `(//section[@class='article-body'])[1]` select by position, yet the elements they hit carry `class='article-intro'` and `class='article-body'`. If a page ever grows another `<header>`, the summary silently comes from the wrong element rather than extracting nothing. `//header[@class='article-intro']/p` is equivalent today and fails loudly.

MaxDall (changes requested):
**REQUEST_CHANGES** — V1_1 drops in-body `<ul>` lists, and image caption/credits are wrong on most images.

`se.Aftonbladet`: read 10 of 100 scanned (18 flagged, 8 in draw); layout coverage news / opinion / listicle / image-heavy, over-capture, image attributes, `VALID_UNTIL` and the V1→V1_1 bump all checked. 3 blockers + 2 nits inline.
Not inline:
- Live-chat and liveblog pages (`/a/Rr77qd/aftonbladet-direkt` plus two `chatta-…` pages, 3 of 100 scanned) extract an empty body. They aren't articles and the default `only_complete=True` crawl drops them — no action needed, noted so the scan counts aren't read as a parser fault.
- The 10 flagged articles outside the draw fall into three benign classes: 7× the site masthead `<dl class='_editorial_…'>` on those body-less pages, 2× the "Fakta" side box (a widget appended after the running text, excluded by V1 as well), 1× an image-gallery `<ul>` whose text is caption + credits rather than body prose.

MaxDall (approved):
@addie9800 Thanks a lot for addressing these layout changes ❤️

PR #954, opened 2026-07-19
fix(publisher): update Daily News TZ sitemap URL and domain

addie9800:
Correct domain and sitemap URL for Daily News (Tanzania).

- Update domain to `https://dailynews.co.tz/`
- Replace sitemap URL with `https://dailynews.co.tz/sitemap_index.xml`
- Add reverse filtering for sitemap processing


MaxDall (approved):
@addie9800 Thanks a lot for updating this publisher ❤️

PR #955, opened 2026-07-19
fix(publisher): update EveningStandard selectors for improved parsing

addie9800:
Align parsing logic with website changes.

- Update paragraph, summary, and subheadline selectors in V1_1
- Adjust data structure in test resources for summaries

Note: breaking with convention, this PR updates the selectors of the current version without creating a new version. The reason for this decision is that the previous selectors were based on something like hashes for class names, which are apparently not consistent across system restarts. So the layout itself did not actually change, but rather the host machine got restarted (probably). This is underlined by the fact, that I accidentally fixed a bug in the current selector where the summary of the test case was not selected with the current setup.


MaxDall (approved):
@addie9800 Thanks a lot for fixing 🙏 And yes, 100% agree with the use of hashes in selectors. Maybe it would be good to put that in the documentation as well?

PR #961, opened 2026-08-02
docs: clarify usage of dynamically generated class names in selector guidelines

addie9800:
Update docs as proposed by @MaxDall in #955 .


MaxDall (approved):
@addie9800 Thanks a lot for addressing this!

PR #965, opened 2026-08-18
feat(publisher): Fix `KleineZeitung` - introduce `V1_1`

addie9800:


MaxDall:
@addie9800 Seems almost like we have to deprecate this one. There are hardly any none pay-walled articles.

addie9800:
> @addie9800 Seems almost like we have to deprecate this one. There are hardly any none pay-walled articles.

Granted, there really aren't a lot of non-paywalled articles, but also not none. And the ones that are not paywalled are also not uninteresting. So I think there might not really be a benefit to actually blocking it.

MaxDall:
@addie9800 Aight, then im gonna give this one a proper review.

src/fundus/publishers/at/kleine_zeitung.py
```diff
@@ -56,5 +60,12 @@ def images(self) -> List[Image]:
                     "./ancestor::figure//*[self::figcaption or contains(@class, 'md:hidden')]|"
                     "./ancestor::div[contains(@class, 'not-prose') or @class=' mb-0']//small"
                 ),
-                author_selector=re.compile(r"©(?P<credits>.*?)$"),
+                author_selector=self._image_author_selector,
             )
+
+    class V1_1(V1):
+        _paragraph_selector = XPath("//div[contains(@class,'prose')]/p")
```
MaxDall, line 67: **Blocker — article prose nested inside a story element is dropped.** The `Reaktionen` factbox sits at `div.prose > div.klz-story-element-factbox > … > div.klz-story-elements-nested`, so its `p.klz-story-element-paragraph` never reaches this direct-child selector — 991 characters of reporting lost: “Die NEOS sehen in den Plänen für die Kapitalerhöhung grundsätzlich eine Chance: „Die Steiermark braucht massive Investitionen …“, so Klubobmann Niko Swatek”. Fix: select the story element by class on a descendant axis — `//div[contains(@class,'klz-story-elements')]//p[contains(@class,'klz-story-element-paragraph')]`; that also closes `contains(@class,'prose')` matching `not-prose` (the „Mehr zum Thema“ teaser box). [[1]](https://www.kleinezeitung.at/artikel/41260165/suche-gestartet-energie-steiermark-soll-investor-an-bord-holen)

src/fundus/publishers/at/kleine_zeitung.py
```diff
@@ -56,5 +60,12 @@ def images(self) -> List[Image]:
...
+        _subheadline_selector = XPath("//div[contains(@class,'prose')]/h2")
```
MaxDall, line 68: **Blocker — same nesting gap on the subheadline side, plus an `h3` variant this selector cannot see.** The factbox heading `<h2 class='klz-story-element-title'>Reaktionen</h2>` is inside `div.klz-story-elements-nested`, so the dropped paragraph above also loses its headline. The pool scan found the same class rendered as `<h3 class='klz-story-element-title …'>` on a second article[[2]](https://www.kleinezeitung.at/artikel/41246240/gerlitzen-baustelle-truebt-erfreuliche-tourismusbilanz). Fix: `//div[contains(@class,'klz-story-elements')]//*[contains(@class,'klz-story-element-sub-head') or contains(@class,'klz-story-element-title')]` — matching by class rather than by tag and depth. Both this and the paragraph selector above are exact no-ops on `KleineZeitung_2026_08_18.html.gz`, so the fixture stays green. [[1]](https://www.kleinezeitung.at/artikel/41260165/suche-gestartet-energie-steiermark-soll-investor-an-bord-holen)

src/fundus/publishers/at/kleine_zeitung.py
```diff
@@ -56,5 +60,12 @@ def images(self) -> List[Image]:
...
+        _summary_selector = XPath("//p[@class='klz-text-article-lead']")
+
+        _image_author_selector = XPath("./ancestor::figure//small")
```
MaxDall, line 71: **Nit — escaped markup lands verbatim in image `authors`.** Swapping the regex for an XPath also drops the regex path's cleanup in `image_extraction`, so the `<small>` node's text is taken as-is. Kleine Zeitung double-escapes some credits (`© &lt;em&gt;Land Steiermark&lt;/em&gt;`), yielding `authors = ['<em>Land Steiermark</em>']`. Captions themselves stay clean — the `<small>` sits outside `<figcaption>` in the new layout, so the XPath swap is otherwise right. Fix: strip tag-like fragments from the selected credit text before `image_author_parsing`. [[1]](https://www.kleinezeitung.at/artikel/41260165/suche-gestartet-energie-steiermark-soll-investor-an-bord-holen)

MaxDall (changes requested):
**REQUEST_CHANGES** — `V1_1`'s direct-child body selectors drop article prose nested inside Kleine Zeitung's story elements.

`at.KleineZeitung`: read 14 of 99 scanned (92 flagged, 14 in draw); layouts, over-capture, image attributes checked. 2 blockers + 1 nit inline.

Not inline:
- **Pool composition** — 91 of 99 articles yield no body, which is the paywall rather than a selector fault: of the 14 read, 12 carry `klz-paywalled-content` on the story container and ship zero paragraphs, and the inherited `free_access` marks all 14 in agreement with that class, so these are distinguishable downstream. A small remainder are *free* yet still ship an empty `klz-story-elements` container — e.g. [/artikel/41259727](https://www.kleinezeitung.at/artikel/41259727/lipizzanerheimat-musikwandertag-war-wieder-voller-erfolg) — body text absent from the HTML entirely, so no selector can reach it. Neither is held against this PR.
- **Layout coverage is thin, unavoidably** — only 8 of the 99 pooled articles have a body at all, so the straight-news / opinion / listicle / image-heavy spread could not be exercised. The offline scan did cover all 99: 2 of those 8 lose content, both to the nesting defect below (the second is [/artikel/41246240](https://www.kleinezeitung.at/artikel/41246240/gerlitzen-baustelle-truebt-erfreuliche-tourismusbilanz), dropping an `<h3 class='klz-story-element-title …'>`). The other 6 swept clean.

Open question: how was `VALID_UNTIL = 2026-07-24` on `V1` established? It asserts the relaunch landed 2026-07-25; I could not verify that independently, and the two fixtures (2025-06-30, 2026-08-18) leave a 13-month window.

src/fundus/publishers/at/kleine_zeitung.py
addie9800: Unfortunately the article is paywalled, so I can't confirm properly, but if it is a factbox, it is out-of-scope for fundus.

src/fundus/publishers/at/kleine_zeitung.py
addie9800: As above, if it's just a factbox, then it is intentionally dropped.

src/fundus/publishers/at/kleine_zeitung.py
addie9800: This seems to be an edge case, which is unfortunately not that simple to solve, as far as I can see. I would accept this parsing error in this case. What do you think?

PR #967, opened 2026-08-18
fix(crawler): publisher skipping bug fix

addie9800:
This PR fixes a bug where a publisher is skipped if access to the robots.txt is not granted, even though `ignore_robots` is set to `True`


MaxDall (approved):
@addie9800 Oh wow, thanks a lot for fixing this bug ❤️

PR #970, opened 2026-08-18
feat(publisher): Introduce `V1_1` for `MexicoNewsDaily`

addie9800:


src/fundus/publishers/mx/mexico_news_daily.py
```diff
@@ -55,3 +57,6 @@ def images(self) -> List[Image]:
                 upper_boundary_selector=XPath("//h1"),
                 author_selector=re.compile(r"\((?P<credits>.*?)\)\s*$"),
             )
+
+    class V1_1(V1):
+        _paragraph_selector = XPath("//div[@class='tts_content_wrapper_1']/p[text()]")
```
MaxDall: **Blocker — `p[text()]` drops any `<p>` whose text sits inside `<em>`/`<a>`, so real prose is lost.** “…UPDATE June 20, 4:16pm CDT: FIFA announced a fine of 10,000 Swiss francs (US $10,000) against the Mexican Football Federation…” never reaches the body [[1]](https://mexiconewsdaily.com/news/player-asks-fans-to-resist-puto-chant-as-fifa-investigates/); a `CORRECTION:` paragraph goes the same way [[2]](https://mexiconewsdaily.com/news/tariffs-incomprehensible-says-economy-secretary/), and all 100 pooled articles show an uncaptured `<p>` (mostly the `Source: …` trailer). Fix: `/p[normalize-space()]`, plus a bloat filter for the `Source:`/subscribe trailers, which would otherwise leak in.

MaxDall (changes requested):
**REQUEST_CHANGES** — the new `V1_1` carries V1's `p[text()]` predicate over, and it silently drops `<em>`-wrapped article prose.

`mx.MexicoNewsDaily`: read 10 of 100 scanned (6 flagged, 6 in draw); static read (`VALID_UNTIL`, version scope, image caption/author/cover pairing), Tier-1 coherence and a structural sweep of all 10 checked. 1 blocker inline.
Not inline:
- Nit — the whole 100-article draw came out of the 2018 end of `post-sitemap` (all ten cached articles publish May–Jun 2018), so neither this review nor the new `MexicoNewsDaily_2026_08_18` fixture — itself a 2018 article — exercises a current, possibly paywalled page. `Sitemap(..., reverse=True)` in `mx/__init__.py` would make crawls start from the newest posts and keep future reviews on the live layout.

MaxDall (approved):

PR #973, opened 2026-08-25
feat(publisher): add Wikinews

MaxDall:
## Summary

Adds Wikinews as a publisher covering 36 language editions through a single MediaWiki API source, and reworks how a publisher's sources are held and restricted so that one publisher can serve many languages usefully.

Wikinews is the motivating case: because all 36 editions sit under one publisher, `crawl(language_filter=...)` is the only way to reach a single edition. That exposed several problems in the existing filter path, which the rest of this PR addresses.

## What changes

**`SourceHandler`** (`url.py`) now owns a publisher's URL sources — ordering them by type, grouping them into batches, and filtering by type and language. `Publisher.source_mapping` stays as a deprecated view. Sources marked `interleave` share a batch.

**`InterleavedSource`** (`html.py`) round-robins a batch of sources instead of exhausting one before starting the next, so a publisher with 36 editions yields from all of them rather than draining the first. It satisfies the `HTMLSource` protocol, so a group is interchangeable with a single source. A publisher's sources now share **one clock**, so the crawl-delay paces the publisher rather than each source — previously, round-robining *n* sources would have fired *n* requests before any of them waited.

**`Publisher.restrict()`** replaces `FilteredPublisher`. It returns a narrowed view of a publisher, and restrictions accumulate: `search(languages=["de"])` followed by `crawl(language_filter=["de", "en"])` yields German only, never both. Publishers are narrowed as soon as the information arrives — source types in `Crawler.__init__`, languages in `crawl()` — so `WebScraper` receives a publisher that is already exactly what it should crawl and no longer carries filter policy. The collection's own publishers are never mutated.

**Publisher admission** in `crawl()` reduces to narrowing everything up front and keeping whatever survives, with a single per-crawler hook for which parser versions an extraction can reach.

**`ParserProxy.versions_covering(start, end)`** selects parser versions by date range. `CCNewsCrawler` now checks attribute support against the versions valid in the crawled window instead of all versions ever.

## Fixes

- **Multi-language sources were crawled and then fully discarded.** `search(languages=["en"])` followed by `crawl(language_filter=["uk"])` kept Ukrainska Pravda's two sitemaps, since both declare `uk`, then dropped every article. Seven sources in the collection declare more than one language. Such a publisher is now rejected up front.
- **`search()` language restrictions could widen.** `FilteredPublisher.from_publisher` overwrote the stored filter instead of intersecting; `extend_filter` intersected but was never called.
- **Per-publisher restrictions were flattened into one global list.** Publishers restricted to German and French respectively produced a meaningless global `{de, fr}` applied to both.
- **`crawl()` and `search()` disagreed on attribute support** — `crawl()` used `attribute_mapping` (all versions, deprecated included), `search()` used the current version only. `crawl()` was the more permissive of the two, producing silent empty crawls.
- **`CCNewsCrawler` admitted publishers for attributes outside its date range**, e.g. a 2016-only crawl accepting a publisher whose 2025 parser version added the attribute.
- **`from_publisher` claimed to deepcopy but shallow-copied** via `__dict__` surgery that bypassed `__init__`.

## Behaviour changes

- **`crawl()` raises `ValueError`** when no publisher matches, instead of logging an error and yielding nothing. This matches `Crawler.__init__`'s existing behaviour for all-deprecated publishers.
- **Skipped publishers are reported as a count**, not per publisher with reasons.
- **`PublisherGroup.search()` returns `List[Publisher]`**; `FilteredPublisher` is gone from the public API.
- **`BaseScraper` takes `publisher_mapping`** in place of `parser_mapping`, and `scrape()` no longer takes `language_filter`.

## Verification

`ruff`, `ruff format`, `mypy` and the full suite pass at **every commit**, not just the tip. Measured against the collection: Wikinews yields 1/36 sources for `["de"]` and 2/36 for `["de", "fr"]`; the whole collection under `["de"]` admits 71/181 publishers and 1064/2326 sources. `versions_covering` was checked against `ParserProxy.__call__` on both sides of a version boundary.

`tests/test_publisher_collection.py` now validates language codes against the ISO 639-3 registry export rather than scraping a Wikipedia table — Wikinews introduced codes like `shn` and `guw` that have no two-letter form.


src/fundus/publishers/base_objects.py
```diff
@@ -202,35 +205,32 @@ def __init__(
...
     @property
     def disallows_training(self) -> bool:
         return self._disallows_training or self.robots.disallows_training()
 
+    @property
+    def sources(self) -> SourceHandler:
+        return self._sources
+
     @property
     def source_mapping(self) -> Dict[Type[URLSource], List[URLSource]]:
```
addie9800, line 220: As far as I can see there is only one use of this within our own project files (and it can easily be replaced with the new version). Hence, I would opt for removing this entirely regarding our upcoming release of `1.0`

src/fundus/publishers/base_objects.py
```diff
@@ -254,69 +254,57 @@ def __eq__(self, other) -> bool:
...
-        # we iterate source by source instead of checking self.languages and self.source_types,
-        # because we need to know if there is a source supporting the given combination of
-        # <source_types> and <languages>
-        filtered_sources = [
-            source
-            for source in more_itertools.flatten(self.source_mapping.values())
-            if (type(source) in source_types if source_types else True)
-            and (source.languages & set(languages) if languages else True)
-        ]
+        """Whether this publisher has a source left to contribute under the given restrictions.
```
addie9800, line 257:
```suggestion
        """Returns True, if this publisher has a source left to contribute under the given restrictions.
```

src/fundus/scraping/crawler.py
```diff
@@ -266,20 +268,46 @@ def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _T:
     return wrapper
 
 
+def supports_attributes(extraction_filter: Optional[ExtractionFilter], versions: Iterable[Type[BaseParser]]) -> bool:
+    """Whether <versions> together cover what <extraction_filter> requires.
```
addie9800, line 272: I think I am not too much of a fan of `whether ...`, because it it's not immediately clear. I think I prefer a blunt. `Returns True, if ...`. I guess this may be personal preference, but if you don't have any strong opinions I would opt for changing it :)

src/fundus/scraping/crawler.py
```diff
@@ -266,20 +268,46 @@ def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _T:
...
+            attributes a parser version has to support.
+        versions: The parser versions the extraction can end up using.
+
+    Returns:
+        bool: True if the required attributes are within reach.
+    """
+    if not isinstance(extraction_filter, Requires):
+        return True
+
+    supported = set(more_itertools.flatten(version.attributes().names for version in versions))
```
addie9800, line 285: I think this logic can lead to similar issues as previously. If we have the requirements of `authors` and `topics` and one version `V1` only supports `authors`, whereas `V2` only supports `topics` We would return True, even though we wouldn't get a single article satisfying the constraints.

src/fundus/scraping/html.py
```diff
@@ -106,6 +118,46 @@ def reset(self):
...
 
+def _interruptable_sleep(seconds: float) -> None:
+    """Sleeps <seconds>, waking early once the crawl is stopped."""
+    __EVENTS__.get("stop").wait(seconds)
+
+
+def build_clock(
+    publisher: Publisher,
+    delay: Optional[Delay] = None,
+    ignore_robots: bool = False,
```
addie9800, line 129: Hm, I think I am not a fan of having two args that are functionally equivalent. I think I would prefer to have one arg and then or the inputs when calling the function.

But honestly, maybe this is just exposes a bigger issue where `ignore_robots` implies `ignore_crawl_delay`. I feel like `ignore_robots` wants to be `ignore_robots_access_rules`, in that case we don't have that ambiguity anymore. Also because, if I see this correctly, we have no way of respecting the delay, but not the access rules. And maybe `ignore_crawl_delay` should also be renamed to `ignore_robots_crawl_delay`, otherwise it might not be intuitive to be able to specify a delay as an arg and at the same time set an arg to ignore it.

src/fundus/scraping/url.py
```diff
@@ -249,3 +253,121 @@ def _extract_predicate(text: str, pattern: Pattern[str]) -> str:
...
+                    f"Allowed are {', '.join(repr(cls.__name__) for cls in iterate_all_subclasses(URLSource))}"
+                )
+
+        self._sources: Tuple[URLSource, ...] = tuple(sorted(collected, key=self._rank))
+
+    @classmethod
+    def _rank(cls, source: URLSource) -> int:
+        """The position of <source>'s type in the crawl order.
+
+        Types not listed in <__SOURCE_ORDER__> are proceeded last rather than raising, so that
```
addie9800, line 285:
```suggestion
        Types not listed in <__SOURCE_ORDER__> are processed last rather than raising, so that
```

src/fundus/scraping/url.py
```diff
@@ -249,3 +253,121 @@ def _extract_predicate(text: str, pattern: Pattern[str]) -> str:
...
+            languages (Optional[Iterable[str]]): Only keep sources covering at least one of
+                these languages.
+
+        Returns:
+            SourceHandler: A new handler over the surviving sources.
+        """
+        allowed_types = set(source_types) if source_types else set()
+        allowed_languages = set(languages) if languages else set()
+
+        return type(self)(
```
addie9800, line 358: Is there a reason you use `type(self)` here, instead of `SourceHandler` directly? It might make it more readable to use the latter.

tests/test_wikinews.py
```diff
@@ -0,0 +1,163 @@
...
+
+import pytest
+
+from fundus import PublisherCollection
+from fundus.publishers.international.wikinews import DEFAULT_EDITION, EDITIONS, WikinewsParser
+
+WIKINEWS = PublisherCollection.international.Wikinews
+
+
+def parser_for_html(html: str) -> WikinewsParser.V1:
```
addie9800, line 20: Why is this hardcoded to the version? Would a potential version update the require also updating all the tests?

addie9800 (changes requested):
Thanks a lot for reworking all of this. Looks looks really good :)

src/fundus/publishers/base_objects.py
MaxDall: Good catch! I will remove the property and replace the usage with `if notpublisher.sources`

src/fundus/scraping/crawler.py
MaxDall: I agree!

src/fundus/scraping/crawler.py
MaxDall: Oh, thanks a lot for catching that! You're absolutely right. I addressed the bug.

src/fundus/scraping/html.py
MaxDall: I get your point and you're totally right about `build_clock`. I removed `ignore_robots` as parameter for `build_clock` and or'ed instead. However, i still think the two parameters earn their place in the library as a hierarchy. `Ignore_robots` completely ignores the robots (access + crawl delay), `ignore_crawl_delay` just ignores the crawl delay while still respecting the robots.

tests/test_wikinews.py
MaxDall: I'll dropped the tests as they didn't fight the overall fundus test schema. Another option would have been to add a new test schema that would have allowed us to test publisher specific or to add functionalities that would allow us to add multiple html fixtures per version. As I do not care as much about WikiNews, I decided neither of them would be worth the effort right now, also the later one will come at one point.

addie9800 (dismissed):
Looks good 👍

PR #975, opened 2026-08-27
Fix review cache reuse, enable impersonation, track skill drift

MaxDall:
## 1. Review caches are scoped to the commit under review

Caches were keyed on the publisher alone (`<tmp>/fundus-review/<cc>.<Class>`) and never deleted. Only `crawl` wipes, so a `sweep`/`status`/`payload` run before it replayed the *previous* review's articles and adjudications — silently, and `payload_gaps` couldn't see it. 47 such directories were on my machine, one a finished review from a month earlier that `payload` would still have emitted findings from.

They are now keyed on the commit:

```
<tmp>/fundus-review/<short-sha>/<cc>.<Class>/
```

The sha comes from `git rev-parse --short HEAD` against the repo `fundus` is imported from — derived, not passed, so no flag to get wrong and the same value under any agent. Two properties fall out for free: a different commit can't inherit these articles, and a mid-review push retires the cache rather than letting adjudications be replayed against superseded parser code.

- **`done`** (new) tears the commit's caches down; PLAYBOOK §6 runs it after the POST.
- **`crawl`** prunes caches older than 7 days.
- **`--cache-dir` is gone.** It bypassed the scoping on every path that mattered.
- **`--pr`** stays on `crawl` as metadata: recorded in `state.json` so `status` and `payload` can name the PR and fill the `gh` commands. Not part of any path.
- PLAYBOOK §1 now says to `gh pr checkout` **before** crawling. Wrong order used to silently review `master`; now the checkout moves `HEAD` and the driver forces a re-crawl.

## 2. The review crawl uses browser impersonation

`cmd_crawl` now passes `Crawler(publisher, impersonate=True)`.

A declared `impersonate="chrome"` only applies when the *caller* opts in this way (`scraping/html.py:125`); publishers declaring none are unaffected. Without it, every publisher behind a TLS-fingerprint anti-bot layer drew zero articles and the review reported "sources or parser likely broken" for a publisher that works. Verified against `mx.MexicoNewsDaily` (`chrome131`): articles with the flag, nothing without.

The profile is recorded in the state and printed by `crawl` and `status`, so a review's body claims are attributable to a specific fingerprint. §1 gains a bullet on both directions a declaration can be wrong — present but unjustified, absent but needed — and §2 a by-hand check for the empty draw, including the `robots.txt` latch (a 403 sticks, so it needs a fresh process).

## 3. Installed skills are tracked for drift

`skills/` is the source and `.claude/skills/` an installed copy, but nothing compared them — the installed `review-publisher` had a fix in `_sweep.py` that existed nowhere else and had drifted unnoticed for weeks.

Installs now write an `.install-manifest.json` with the skill's `version` and a content fingerprint. `Agent.status()` re-fingerprints both sides against it, separating the two kinds of drift — they have opposite fixes:

| | Meaning | What to do |
|---|---|---|
| `·` | not installed | install |
| `✓` | in sync | nothing |
| `↑` | the repo moved on | re-install |
| `≠` | the **installed copy** was edited in place | port it back to `skills/` first — a re-install destroys it |
| `?` | no usable manifest, and the two differ | compare by hand |

A version alone can't catch `≠` — nobody editing an installed copy bumps a label; the fingerprint does that work. `version:` stays optional: unversioned skills show `—` and still get full drift detection.

The manifest carries a `format`, so changing the fingerprint algorithm degrades old manifests to `?` rather than accusing an untouched copy of local edits. Not hypothetical — it fired on this repo's own install mid-branch.

`review-publisher` is tagged `version: 2.0.0` — major: `--cache-dir` removed, cache location moved, `state.json` schema changed.

## Review notes

Two agent review passes ran over this branch. Everything raised is fixed except these, kept knowingly:

- `source` and `installed_at` in the manifest have no reader — human breadcrumbs, labelled as such.
- `fingerprint` covers names and bytes, not mode bits or empty directories: a `chmod -x` on an installed script still reads `✓`. Documented in the docstring.
- Two agents reviewing the same commit concurrently share a cache dir, and the second `crawl` would wipe the first's adjudications. Rare enough to accept; the alternative re-introduces per-session keys and loses cross-session resume.

### DIFF
--- a/skills/README.md
+++ b/skills/README.md
@@ -10,7 +10,7 @@ their own agent with the installer below.
 
 | Skill | What it does |
 |-------|--------------|
-| [`review-publisher`](review-publisher/SKILL.md) | Reviews a publisher PR (new publisher or parser-version change): crawls live articles to verify the extracted `ArticleBody` mirrors the real article, checks `VALID_UNTIL` / version bumps / `validate=False` / `free_access`, and drafts a single GitHub review. |
+| [`review-publisher`](review-publisher/SKILL.md) | Reviews a publisher PR (new publisher or parser-version change): crawls live articles to verify the extracted `ArticleBody` mirrors the real article, checks `VALID_UNTIL` / version bumps / `validate=False` / `free_access` / `impersonate`, and drafts a single GitHub review. |
 
 `install.py` (the installer launcher) and the `installer/` package behind it, plus this
 `README.md`, round out the folder.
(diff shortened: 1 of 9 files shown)


MaxDall:
@addie9800 Regarding the review: i think the only files that may be worth your time looking at are the individual `.md` files (Skill, Playbook, Readme). In short this just scopes the review skills cache to a specific commit hash, so that stale caches do not interfere with the review process, as well as adding versioning to the skill system and enabling `impersonation` for the draft pool.

addie9800 (commented):
This looks really good, I only have one question though: as I see it, this new convention of referencing the cache directory using the git hash would not only guarantee that the adjudications are not stale, but also require a recrawl for every commit that is reviewed, right? Wouldn't it be better to track the latter by perhaps the PR id, because even though the adjudications need to be refreshed, the articles should be fine.

MaxDall:
> This looks really good, I only have one question though: as I see it, this new convention of referencing the cache directory using the git hash would not only guarantee that the adjudications are not stale, but also require a recrawl for every commit that is reviewed, right? Wouldn't it be better to track the latter by perhaps the PR id, because even though the adjudications need to be refreshed, the articles should be fine.

I totally get your point. I originally opted for the PR id, but later thought that this would also include the review findings etc. and rather wanted to be save than having stale files/findings corrupt the review. Further, i guess my thought was that re-crawling for a new assessment doesn't really hurt.

addie9800 (approved):
Nice 👍

PR #978, opened 2026-09-07
feat(publisher): Add `V1_1` for `RuhrNachrichten`

addie9800:


src/fundus/publishers/de/rn.py
```diff
@@ -54,3 +56,12 @@ def images(self) -> List[Image]:
                 caption_selector=XPath("./ancestor::figure//figcaption/text()"),
                 author_selector=XPath("./ancestor::figure//figcaption/span"),
             )
+
+    class V1_1(V1):
```
MaxDall, line 60: **Blocker — `RuhrNachrichten` is a premium publisher with no `free_access`.** 64 of 100 crawled articles are RN+ pages that serve no paragraph text at all. The signal is right there in the `ld+json`: `"isAccessibleForFree": "False"` on all 7 paywalled articles in the draw, absent on the free ones.

Fix: `free_access` off `self.precomputed.ld.bf_search("isAccessibleForFree")` — note it is the string `"False"`, not a bool. Pre-existing; `V1` lacks it too. [[1]](https://www.ruhrnachrichten.de/dortmund-sport/kreisliga-c-tili-ethirmanasingam-aus-eving-selimiye-spor-tamilstars-interview-w1243193-2002217991/)

tests/resources/parser/test_data/de/RuhrNachrichten.json
```diff
@@ -259,5 +259,99 @@
...
+            "Neben ihrem Mann Gerd hinterlässt Gesine Homrighausen auch ihre Kinder und Enkelkinder. Rebecca Gutzat will nicht glauben, dass ihre Freundin nicht woanders für Lachen sorgen wird: „Vielleicht ziehst du jetzt irgendwo deinen bunten Anzug an, nimmst deine Luftballons und deinen Teddy und machst dort weiter, wo du aufgehört hast – Menschen zum Lachen zu bringen.“"
+          ]
+        }
+      ]
+    },
+    "images": [
+      {
+        "versions": [
+          {
+            "url": "https://www.ruhrnachrichten.de/wp-content/uploads/2026/09/07/12/630_0900_6808292_-320x214.jpg,",
```
MaxDall, line 305: **Blocker — half of every extracted image URL ends in a stray comma and 404s.** RN's srcset is `…-320x214.jpg, …-640x428.jpg 2x`; `_srcset_pattern` in `src/fundus/parser/utility.py:662` matches the url as `\S+`, so a descriptor-less candidate swallows the separator (also line 314 in this block, and on every image of both live articles that had one).

Fix: `(?P<url>[^\s,]+)`. Base-parser bug affecting all publishers — but this fixture freezes the broken URL as expected output. [[1]](https://www.ruhrnachrichten.de/werne/bombendrohung-freilichtbuehne-werne-premiere-abbruch-sicherheit-zukuenftige-veranstaltungen-w1245065-2002218791/)

src/fundus/publishers/de/rn.py
```diff
@@ -54,3 +56,12 @@ def images(self) -> List[Image]:
...
+        _bloat_regex = r"(?i)^verfasst von"
```
MaxDall, line 61: **Nit — the bloat filter is anchored on un-normalized text, and never fired.** `re:test(string(), '(?i)^verfasst von')` anchors at position 0 of the concatenated descendant text, so a leading newline in the markup defeats it — and no "Verfasst von" paragraph appears in any of the 10 crawled articles or in the new fixture, so the branch is untested. Which layout does it target?

Fix: `normalize-space()` rather than `string()`; and consider `ol[@class='wp-block-list']` alongside `ul` while you are in there.

MaxDall (changes requested):
**REQUEST_CHANGES** — the new `V1_1` selector extracts free articles faithfully; both blockers are pre-existing and sit outside it.

`de.RuhrNachrichten`: read 10 of 100 scanned (82 flagged, 10 in draw); `V1_1` body selectors, `VALID_UNTIL`/version split, layout coverage, over-capture and image attributes checked. 2 blockers + 1 nit inline.
Not inline:
- 64/100 articles carry no body: RN+ pages ship only `<h2>` subheads inside `div.article__content > div.PremiumContent` and none of the paragraph text, so `bool(body)` is `False` and the default `only_complete=True` crawl drops them. Not a selector defect — but see the `free_access` comment.
- 8/100 flag only on a dropped `<p class="product-card__description">` (newsletter-widget pitch); correctly kept out of the body.
- Layout coverage: the draw spanned straight news, a sports report, an interview and a live-ticker, plus one bullet-list piece that exercises the new `ul.wp-block-list` branch. No opinion/column or image-heavy gallery surfaced in 100 articles.

Neither blocker is a regression from this diff — happy to see them split off if you would rather land the version bump as-is.

tests/resources/parser/test_data/de/RuhrNachrichten.json
MaxDall: Fixed at the root in #982 (`parse_srcset` was matching the url as `\S+`). The correction I suggested above — `[^\s,]+` — turned out to be wrong: it truncates urls carrying commas of their own, like the CDN transformations Nine News and NZZ serve. #982 strips only commas that *terminate* a url instead.

Once that lands, this fixture needs `python -m scripts.generate_parser_test_files -oj -p RuhrNachrichten` — it clears both trailing commas here and the `"type": null` on the same two versions, which was a knock-on of the same bug.

src/fundus/publishers/de/rn.py
addie9800: `free_access` is implemented in exactly that manner in `BaseParser`. It seems to also work reliably.

tests/resources/parser/test_data/de/RuhrNachrichten.json
addie9800: Done, thanks for fixing.

src/fundus/publishers/de/rn.py
MaxDall: @addie9800 Sorry you had to deal with that, this one slipped though my "review" of the agents review.

MaxDall (approved):
@addie9800 Thanks a lot for updating this one!

PR #980, opened 2026-09-08
fix(publisher): replace `TaipeiTimes` sources with RSS feed

addie9800:
The Newsmap and SiteMap objects are no longer accessible


MaxDall (changes requested):
@addie9800 Thanks for addressing this. Unfortunately it seems like there's nothing we can do. Maybe it would be good to leave the sitemap references in a comment stating to re add them once they're accessible again? What do you think?

addie9800:
> @addie9800 Thanks for addressing this. Unfortunately it seems like there's nothing we can do. Maybe it would be good to leave the sitemap references in a comment stating to re add them once they're accessible again? What do you think?

Yes, I also thought about it when removing them and ultimately decided against, but also wasn't too sure about it. I change it :)

MaxDall (approved):

PR #982, opened 2026-09-08
fix(parser): strip only url-terminating commas when parsing srcset

MaxDall:
`parse_srcset` matched a candidate's url as `\S+`, so a candidate written without a descriptor swallowed the comma that separates it from the next one:

```
srcset="a-320x214.jpg, a-640x428.jpg 2x"   ->   1x: "a-320x214.jpg,"
```

The url then 404s, and `ImageVersion._parse_type` reads the extension as `jpg,` and yields `type: null`.

The naive fix (`[^\s,]+`) would truncate urls carrying commas of their own — CDN transformations like `.../w_300,h_200/img.jpg`, which this repo already crawls (Nine News, NZZ). The pattern now strips only commas that *terminate* a url, matching the HTML srcset grammar. Two extras fall out: repeated separators (`a.jpg,, b.jpg 2x`) and whitespace ahead of the comma (`a.jpg 1x , b.jpg 2x`, previously yielding a bogus url of just `","`).

210 urls across 8 fixtures were storing the broken value; regenerated with `python -m scripts.generate_parser_test_files -oj`, every change a pure trailing-comma strip. Found while reviewing #978, whose new `RuhrNachrichten` fixture hits the same bug — it needs the same `-oj` regeneration once this lands.

### DIFF
--- a/src/fundus/parser/utility.py
+++ b/src/fundus/parser/utility.py
@@ -659,7 +659,10 @@ def image_author_parsing(authors: Union[str, List[str]]) -> List[str]:
 
 
 # https://regex101.com/r/MplUXL/2
-_srcset_pattern = re.compile(r"(?P<url>\S+)\s*(?P<descriptor>[0-9.]+[wx])?(,?\s*)")
+# candidates are separated by a comma that may sit flush against the preceding url, e.g.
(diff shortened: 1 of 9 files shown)


addie9800 (approved):
Nice fix 👍

